PR #1, opened 2018-08-22
Add initial project structure based on create-react-app

vojtechszocs:
- initial structure based on [create-react-app](https://github.com/facebook/create-react-app) template
- simple [Storybook](https://github.com/storybooks/storybook) config that favors co-location of components and their stories
- added dependency on [patternfly-react](https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-react) package

We can start adding new components and their stories into `src/components` directory.

First-time setup: make sure Node.js v8 and Yarn is installed, then run `yarn install` to pull project's dependencies.

Run `yarn storybook` and open http://localhost:9001/ to see and interact with components covered by their stories. Run `yarn test` to execute unit tests through Jest, using `jsdom` env.

There's also `yarn build` and `yarn start` to build the whole application and host it through webpack dev-server, but we're focusing on individual components for now.

As the next step (besides adding new components), we should simplify the project structure and align it with other projects like `patternfly-react` and `ovirt-web-ui`.


package.json
```diff
@@ -0,0 +1,99 @@
...
+    "css-loader": "0.28.7",
+    "dotenv": "4.0.0",
+    "dotenv-expand": "4.2.0",
+    "eslint": "4.10.0",
+    "eslint-config-react-app": "^2.1.0",
+    "eslint-loader": "1.9.0",
+    "eslint-plugin-flowtype": "2.39.1",
+    "eslint-plugin-import": "2.8.0",
+    "eslint-plugin-jsx-a11y": "5.1.1",
+    "eslint-plugin-react": "7.4.0",
```
ohadlevy: how about using the patternfly eslint rules?

package.json
```diff
@@ -0,0 +1,99 @@
...
+    "promise": "8.0.1",
+    "raf": "3.4.0",
+    "react": "^16.4.2",
+    "react-dev-utils": "^5.0.1",
+    "react-dom": "^16.4.2",
+    "resolve": "1.6.0",
+    "style-loader": "0.19.0",
+    "sw-precache-webpack-plugin": "0.11.4",
+    "url-loader": "0.6.2",
+    "webpack": "3.8.1",
```
ohadlevy: imho many of these needs to be dev-dep, not to mention there are really a lot here that I'm not sure we need, plus its an old version of webpack, which I'm not sure we should start with?

README.md
```diff
@@ -1,2 +1,2444 @@
-# web-ui
-User interface project
+This project was bootstrapped with [Create React App](https://github.com/facebookincubator/create-react-app).
```
mareklibra: I would either remove this file or rename it to sort of README_react-create-app.md.
We should provide our own README file with related content.

But this can be done in a follow-up.

package.json
mareklibra: While contributing to patternfly-react project, I found them over-complicated. The resulting syntax is confusing to me. But this is just my opinion, I can get used to whatever syntax rules anyone with stronger opinion prefers.

package.json
mareklibra: Full agreement with the dev-deps.

I think this is what was generated by most recent react-create-app, right?
Considering project context, sounds reasonable to me to stay aligned with openshift/console on that.
IIRC, patternfly-react has no requirement on particular webpack version.

mareklibra: Lesson learned from other projects: we should keep updating packages and corresponding sources as often as possible. In small steps. Otherwise it becomes nightmare very quickly. 
We should agree on process for that - once a week? By human is enough, IMO.

public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "React App",
+  "name": "Create React App Sample",
```
mareklibra: Our name?

package.json
jelkosz: once a week is a bit too often for a human I'd say. I'd suggest to do it at the beginning of each sprint (e.g. once in 3 weeks). This way we don't risk to break it just before the iteration ends :)

public/manifest.json
jelkosz: We could start with "kubevirt web ui" and the long would be like "A UI for Managing Kubevirt".
What you think?

mareklibra: Kubevirt? :-D 
If so, I would rename whole project to "kui" or better "kua" (if anyone asks then Kubevirt UserInterface in Action). The second one sounds best to me.

If Kubevirt is not an option in the name, then what about "Virtualization User Interface for Kubernetes"

package.json
ohadlevy: the main value is that we should work across multiple repos (pf-react, console and this) - it would make life a lot easier to have similar linting configurations....

package.json
mareklibra: Right, makes sense

ohadlevy: there are tools to automate it(e.g. https://greenkeeper.io/), while not sure i love  that option, i think a good way is to keep version matrix relaxed on main branch, and lock it for stable releases, this way you fix issues quickly (assuming its easy fix) along the way, but keep stable branches stable over time.

README.md
ohadlevy: :+1: to remove.

public/manifest.json
ohadlevy: Naming things... I would not care about this for now, I suggest open a specific PR (after this one) to discuss the name if you like, but not block getting the foundations in.

package.json
vojtechszocs: @ohadlevy This is just the initial setup based on create-react-app template to get us started. There will be follow-up changes to simplify project's dependencies and align them with other projects you've mentioned. This also applies to webpack & Babel config etc.

We can indeed try using [eslint-plugin-patternfly-react](https://github.com/patternfly/patternfly-react/tree/master/packages/eslint-plugin-patternfly-react) later on, I've no experience with it myself though. If we end up not using it, we'll still want to get as much overlap as possible to ensure the code looks similar to other projects.

ohadlevy: @vojtechszocs I understand and support this initiative, some things are easier to remove than others, overall, don't see my comments as blocking merging this.

I do think it does make sense to use the pf-react eslint plugin to avoid large refactoring later on.

package.json
vojtechszocs: Guys, all of this is currently taken from create-react-app template :smile: 

> imho many of these needs to be dev-dep

Agreed, it makes sense to separate developer/tool deps from build deps.

> not to mention there are really a lot here that I'm not sure we need

Yes, we will simplify/remove & align deps with other projects. With this PR, @rawagner can start experimenting with "Create/Import VM" wizard component in Storybook, while we can update/simplify/improve the project structure.

> plus its an old version of webpack, which I'm not sure we should start with?

webpack 4.x released on Feb 2018 but there might still be plugins/loaders which don't work 100% with the new version. For a brand new project, I agree that it makes sense to use the latest stable version, so we'll upgrade webpack version as a follow-up. (Fun fact, [ovirt-web-ui uses webpack 1.x](https://github.com/oVirt/ovirt-web-ui/blob/master/package.json#L63) and it might not yield very optimized JS output, but it still works.)

> I think this is what was generated by most recent react-create-app, right?

Exactly! :grin: 

> Lesson learned from other projects: we should keep updating packages and corresponding sources as often as possible. In small steps. Otherwise it becomes nightmare very quickly.
> I'd suggest to do it at the beginning of each sprint (e.g. once in 3 weeks). This way we don't risk to break it just before the iteration ends :)

Agreed. As @ohadlevy mentioned, there are tools like GitHub [Greenkeeper](https://github.com/integration/greenkeeper) to monitor & update-test your project's dependencies, we can use that.

README.md
vojtechszocs: OK, it will be renamed.

public/manifest.json
vojtechszocs: > manifest.json provides metadata used when your web app is added to the homescreen on Android

So I guess we can remove this file :smile: 

As for the naming, it's currently `kubevirt-web-ui` in `package.json` file, and I'm all for a better name!

vojtechszocs:
Apologies to everyone who got confused by the overall complexity of this PR (dependencies, lots of files that might not be necessary at this point) - it's the `create-react-app` template.

We will simplify things & align with other projects, of course - in parallel to other people who might start experimenting with new components and testing them via Storybook.

vojtechszocs:
If everyone's OK with this PR, we can merge it and improve from there.

I've opened #2 to track those improvements.

PR #4, opened 2018-08-28
[WIP] Implement Create VM Wizard component

rawagner:
storybook is published here https://rawgit.com/rawagner/web-ui/create_VM_wizard_storybook/index.html


ohadlevy:
thanks! do you mind publishing the storybook?

rawagner:
@ohadlevy where should I publish it ? github pages of my forked repo ?

mareklibra:
> github pages of my forked repo

Seems to be best option. It will stay close to each PR.
And we can maintain storybook for master.

rawagner:
@mareklibra OK, I will go with the github pages. Just be aware that github pages is using gh-pages or master branch. Which means you can have only one github page for repository. It will work for now, as I have only one PR but if I will have more open PRs I wont be able to host storybook for each of them.

ohadlevy:
yeah, best is to use rawgit (so you can push multiple branches), have a
look at pf-react for the npm task (i think its storybook-deployer)

ohadlevy:
do we have a group of UX developers so we can ping them to look at the
stroybook?

rawagner:
@lizsurette

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+
+export class CreateVMWizard extends React.Component {
+
+    constructor(props){
+      super(props);
+      this.state = defaultState;
+    }
+  
+    onFormChange = (newValue, target) => {
+      let valid = null;
```
vojtechszocs: To avoid confusion with `wizardValid` being a boolean, can we call this `validMsg` or similar?

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+      }
+      const basicSettings = {
+        ...this.state.basicSettings,
+        [target]: {
+          value: newValue,
+          valid: valid
+        }
+      };
+      this.setState({
+        basicSettings: basicSettings
```
vojtechszocs: You can use the shorthand syntax here.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+      {
+        title: 'Basic Settings',
+        render: () => <Form horizontal>
+          {Object.keys(this.basicSettings).map(key => {
+            if (this.basicSettings[key].isVisible && this.basicSettings[key].isVisible(this.state.basicSettings) === false){
+              return null;
+            }
+            let child = null;
+            const valid = _.get(this.state.basicSettings[key], 'valid', null);
+            if (this.basicSettings[key].type === 'text') {
```
suomiy: please consider using switch

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+        ...this.state.basicSettings,
+        [target]: {
+          value: newValue,
+          valid: valid
+        }
+      };
+      this.setState({
+        basicSettings: basicSettings
+      });
+      this.validateWizard(basicSettings);
```
vojtechszocs: At this point, the component has already re-rendered, since `this.setState` triggers a new render.

`this.validateWizard` calls `this.setState` again, triggering another re-render.

We can improve this by doing validation as part of component's render.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+  
+    }
+  
+    wizardStepsNewVM = [
+      {
+        title: 'Basic Settings',
+        render: () => <Form horizontal>
+          {Object.keys(this.basicSettings).map(key => {
+            if (this.basicSettings[key].isVisible && this.basicSettings[key].isVisible(this.state.basicSettings) === false){
+              return null;
```
suomiy: can be undefined,
`let child;` also

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+      //check if all fields are valid
+      for (let key in values) {
+        if (values[key].valid && (this.basicSettings[key].isVisible ? this.basicSettings[key].isVisible(values) : true)) {
+          wizardValid = false;
+          break;
+        }
+      }
+
+      this.setState({
+        wizardValid: wizardValid
```
vojtechszocs: As far as I can see, this `validateWizard` method could simply return a boolean value and get called from component's render, instead of doing it via state update.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+                  bsStyle="default"
+                  className="wizard-dropdown"
+                  title={_.has(this.state.basicSettings[key], 'value') ? this.state.basicSettings[key].value : this.basicSettings[key].default}
+                  onSelect={value => this.onFormChange(value, key)}>
+                  {this.basicSettings[key].values.map(value => <MenuItem key={value.name} eventKey={value.value ? value.value : value.name}>{value.name}</MenuItem>)}
+                </DropdownButton>
+              </ButtonGroup>;
+            }
+            if (this.basicSettings[key].type === 'checkbox') {
+              child = <Checkbox checked={_.get(this.state.basicSettings[key], 'value', false)} onChange={event => this.onFormChange(event.target.checked, key)}>{this.basicSettings[key].title}</Checkbox>;
```
suomiy: please put props on new lines for better readability

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+            }
+            return <FormGroup key={key} validationState={valid ? 'error' : null}>
```
suomiy: please use brackets () and put FormGroup on the newline. IMO it is more readable because of the same spacing

suomiy: + applies also elsewhere

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+
+      //check if all required fields are defined
+      const requiredKeys = Object.keys(this.basicSettings).filter(key => this.isFieldRequired(key, values));
+      const requiredKeysInValues = Object.keys(values).filter(key => this.isFieldRequired(key, values));
+      if(requiredKeys.length !== requiredKeysInValues.length){
+        wizardValid = false;
+      }
+
+      //check if all fields are valid
+      for (let key in values) {
```
vojtechszocs: Code that enumerates over all (including inherited) properties of an object is more dangerous than code that iterates over own (excluding inherited) properties of an object.

I'd suggest using `Object.keys` instead of `for..in`.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+        if (values[key].valid && (this.basicSettings[key].isVisible ? this.basicSettings[key].isVisible(values) : true)) {
```
vojtechszocs: The second part of the `if` condition can be written as
```
(this.basicSettings[key].isVisible && this.basicSettings[key].isVisible(values)) || true
```
instead of using the ternary operator. Even better, it could be extracted into a `const` variable so that the condition itself is easier to reason about.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+      {
+        title: 'Result',
+        render: () => {
+          this.props.k8sCreate(this.createVM(this.state.basicSettings, this.state.network, this.state.storage))
+          return <p>object visible in console</p>
+        }
+      }
+    ]
+
+    createVM = (basicSettings, network, storage) => {
```
suomiy: I would rather see this in another file and give it a data in stable format. Maybe some file like `requests.js` ? Then it would create final data structure vm.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+      const requiredKeysInValues = Object.keys(values).filter(key => this.isFieldRequired(key, values));
+      if(requiredKeys.length !== requiredKeysInValues.length){
+        wizardValid = false;
+      }
+
+      //check if all fields are valid
+      for (let key in values) {
+        if (values[key].valid && (this.basicSettings[key].isVisible ? this.basicSettings[key].isVisible(values) : true)) {
+          wizardValid = false;
+          break;
```
vojtechszocs: I guess this `break` is the reason of using `for..in` loop.

A better alternative IMO is to use `Object.keys` + `Array.prototype.find` to find the first non-valid element and compute `wizardValid` value based on that.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+      this.setState({
+        wizardValid: wizardValid
+      });
+    }
+
+    isFieldRequired = (key, basicSettings) => {
+      return this.basicSettings[key].required && (this.basicSettings[key].isVisible ? this.basicSettings[key].isVisible(basicSettings) : true);
+    }
+  
+    basicSettings = {
```
vojtechszocs: What's the relation between component instance `this.basicSettings` and `this.state.basicSettings` ? (Do we need the second one?)

vojtechszocs: Also, this looks like a declaration of form fields to be *interpreted* by the wizard component.

This is a good idea, as it separates "what" from "how" in terms of form fields.

In future, I'd like us to extract this concept into smaller, independent and reusable bits, including the common validation logic.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+      authKeys:{
+        title: 'Authenticated SSH Keys',
+        type: 'textarea',
+        isVisible: (basicSettings) => _.get(basicSettings, 'cloudInit.value', false),
+        required: true
+      }
+  
+    }
+  
+    wizardStepsNewVM = [
```
vojtechszocs: And this looks like a concept of wizard step definition, built on top of above form field declarations.

I like this separation, since we can extract individual form field declarations into separate module and reuse them in multiple components if necessary.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+      {
+        title: 'Basic Settings',
+        render: () => <Form horizontal>
```
vojtechszocs: Please do
```
render: () => (
  <Form horizontal>...</Form>
)
```
for better structuring and readability.

This applies to code below too.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+          {Object.keys(this.basicSettings).map(key => {
+            if (this.basicSettings[key].isVisible && this.basicSettings[key].isVisible(this.state.basicSettings) === false){
```
vojtechszocs: A lot of this logic looks generic enough, we can extract it into separate module in the future, to keep this particular component as small and simple as possible.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+      }
+      if(basicSettings.startVM){
+        vm.spec.running=basicSettings.startVM.value;
+      }
+      return vm;
+    }
+  
+    closeWizard = () => {
+      this.setState(defaultState);
+      this.props.closeWizard();
```
vojtechszocs: If `closeWizard` is meant to be a callback prop, please call it `onClose` instead.

vojtechszocs: Also, patternfly-react `Wizard` already provides `onExited` callback prop that executes when the modal finishes its fade-out transition, I'd suggest using that instead.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+    onStepChanged = (index) => {
+      this.setState({ activeStepIndex: index })
+    }
+  
+    render() {
+      return <Wizard.Pattern
+        show={this.props.show}
+        onHide={this.closeWizard}
+        onClose={this.closeWizard}
+        steps={this.wizardStepsNewVM}
```
vojtechszocs: `steps` is where the vast majority of wizard UI & logic lives.

I think that in future, we should find a better representation, I recall talking with patternfly-react folks about allowing `steps` to be represented as React components, instead of plain JS array.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    onStepChanged = (index) => {
+      this.setState({ activeStepIndex: index })
+    }
+  
+    render() {
+      return <Wizard.Pattern
+        show={this.props.show}
+        onHide={this.closeWizard}
```
vojtechszocs: Wizard's `onClose` is triggered when you click the "X" button in top right corner. WizardPattern implements it for us, but doesn't expose it as a prop, so `onClose` should be removed here.

As for `onHide`, this is a function that WizardPattern calls like so:
```
onHide(onFinalStep) // true or false, whether we are on the final step
```

TL;DR I'd just keep the `onHide` and remove `onClose`.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+        onClose={this.closeWizard}
+        steps={this.wizardStepsNewVM}
+        activeStepIndex={this.state.activeStepIndex}
+        onStepChanged={index => this.onStepChanged(index)}
+        nextStepDisabled={!this.state.wizardValid}
+        nextText={this.state.activeStepIndex === 2 ? 'Create Virtual Machine':'Next'} />;
```
vojtechszocs: As for JSX expressions, please use one of these forms (in-line with patternfly-react):
```
<Example foo={bar} />

// lots of props
<Example
  foo={bar}
/>

<Example foo={bar}>
  <AnotherExample />
</Example>

// lots of props
<Example
  foo={bar}
>
  <AnotherExample />
</Example>
```

src/components/wizards/create-vm/StorageTable.js
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import { TablePfProvider, selectionHeaderCellFormatter, selectionCellFormatter, Table } from 'patternfly-react';
+import { Header, Body} from 'reactabular-table';
+
+export class StorageTable extends React.Component {
+
+    state = {
```
vojtechszocs: You can use this form in `CreateVmWizard` component too.

Let's be consistent with how we initialize component's state, using ES6 class fields sounds good to me.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,358 @@
+import React from 'react';
```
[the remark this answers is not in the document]
vojtechszocs: I also prefer just `.js` for both React components and other JS code.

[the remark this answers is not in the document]
vojtechszocs: That's simply because the presence of JSX expressions doesn't make the particular file special, JSX is transformed into JS anyway :smiley:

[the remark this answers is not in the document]
rawagner: +1 for just .js

src/components/wizards/new-vm/NewVMWizard.stories.js
```diff
@@ -0,0 +1,96 @@
...
+        {
+            name: 'template1'
+        },
+        {
+            name: 'template2'
+        }
+    ]
+
+    openWizard = () => {
+        this.setState({show: true});
```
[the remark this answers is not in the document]
vojtechszocs: Please take some inspiration from
* https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-react/src/components/Wizard/Stories/WizardPatternExample.js
* https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-react/src/components/Wizard/Stories/WizardPatternExampleStory.js

src/components/wizards/create-vm/CreateVMWizard.js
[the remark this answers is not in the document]
vojtechszocs: I think the intent is to simplify the render by isolating wizard's steps into a separate variable (or technically an instance property).

We should strive to separate "what" from "how" so that the actual component is as simple as possible.

src/components/wizards/new-vm/NewVMWizard.stories.js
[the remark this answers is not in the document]
suomiy: I guess what bothers me the most that there are 3 possible wizzards in the component and the prop callbacks are tangled to the component. Is it possible to make it somehow simpler? 

I guess we could make the component lifecycle dependent only on the props or make it standalone.

src/components/wizards/create-vm/CreateVMWizard.js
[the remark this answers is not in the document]
suomiy: makes sense, + extracting the generic stuff outside of this component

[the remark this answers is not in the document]
ohadlevy: thats why i wanted to start with redux, so you have to split it...

src/components/wizards/create-vm/CreateVMWizard.css
```diff
@@ -0,0 +1,7 @@
+.wizard-dropdown {
+    text-align: left;
+}
+
+.wizard-dropdown .caret {
+    float: right;
+}
```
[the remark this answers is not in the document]
rawagner: added newline to all files

src/components/wizards/new-vm/NewVMWizard.stories.js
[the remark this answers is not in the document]
rawagner: I changed it to what @suomiy suggested. Just render or not render at all. Is is the preferred way ?

src/components/wizards/create-vm/CreateVMWizard.js
rawagner: done

src/components/wizards/create-vm/CreateVMWizard.js
rawagner: done

src/components/wizards/create-vm/CreateVMWizard.js
rawagner: done

src/components/wizards/create-vm/CreateVMWizard.js
rawagner: done, hopefully everywhere

src/components/wizards/create-vm/CreateVMWizard.js
rawagner: OK, I moved it to k8s/request.js

src/components/wizards/create-vm/CreateVMWizard.css
```diff
@@ -0,0 +1,7 @@
+.wizard-dropdown {
```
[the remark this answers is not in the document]
rawagner: done

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+import { Form, FormControl, Col, ButtonGroup, DropdownButton, MenuItem, Checkbox, FormGroup, ControlLabel, HelpBlock, Wizard } from 'patternfly-react';
```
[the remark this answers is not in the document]
rawagner: done

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+    onFormChange = (newValue, target) => {
+      let valid = null;
+      if (this.basicSettings[target].required && newValue.trim().length === 0){
+        valid= `${this.basicSettings[target].title} is required`;
+      }
+      const basicSettings = {
+        ...this.state.basicSettings,
+        [target]: {
+          value: newValue,
+          valid: valid
```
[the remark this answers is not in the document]
rawagner: done

src/components/wizards/create-vm/CreateVMWizard.js
rawagner: renamed to validMsg

src/components/wizards/create-vm/CreateVMWizard.js
rawagner: onClose removed

src/components/wizards/create-vm/StorageTable.js
rawagner: Ok, class fields it is.

src/components/wizards/create-vm/CreateVMWizard.js
```diff
@@ -0,0 +1,366 @@
...
+export class CreateVMWizard extends React.Component {
+
+    constructor(props){
+      super(props);
+      this.state = defaultState;
+    }
+  
+    onFormChange = (newValue, target) => {
+      let valid = null;
+      if (this.basicSettings[target].required && newValue.trim().length === 0){
```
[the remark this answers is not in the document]
rawagner: `this.basicSettings` represents fields of form (type, label, validation). `this.state.basicSettings` represents actual values that are changing based on user input (value, validation message). Maybe I should rename one of them to avoid confusion

src/components/wizards/new-vm/OperatingSystem.js
```diff
@@ -0,0 +1,425 @@
+export const os = [
+    {
+        name: 'Alpine Linux 3.5',
+        value: 'alpinelinux3.5'
```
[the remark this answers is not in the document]
rawagner: renamed to id.

src/components/wizards/new-vm/NewVmWizard.stories.js
```diff
@@ -0,0 +1,93 @@
...
+  closeWizard = () => {
+    this.setState({show: false});
+  }
+
+  render() {
+    return (
+      <React.Fragment>
+        <Button onClick={this.openWizard}>Open New VM Wizard</Button>
+          {this.state.show ? <NewVmWizard
+            closeWizard={this.closeWizard}
```
suomiy: I think onClose would be more fitting

src/components/wizards/new-vm/NewVmWizard.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+      return (
+        <React.Fragment>
+          <Wizard show onHide={this.props.closeWizard}>
+            <Wizard.Header title="Create Virtual Machine" onClose={this.props.closeWizard} />
+            <Wizard.Body>
+              <Wizard.Row>
+                <Wizard.Main>
+                  <Wizard.Contents stepIndex={0} activeStepIndex={0} className="wizard-content">
+                    <button className="btn btn-link wizard-button " onClick={this.openCreateVmWizard}>
```
suomiy: You could make a component out of this

src/components/wizards/create-vm/CreateVmWizard.js
```diff
@@ -0,0 +1,228 @@
...
+        required: true
+      },
+      flavor:{
+        title: 'Flavor',
+        type: 'dropdown',
+        default: '--- Select Flavor ---',
+        values: this.props.flavors.concat([{name:'Custom'}]),
+        required: true
+      },
+      memory:{
```
suomiy: please better validation for memory/cpu/URL

src/utils/forms/FormFactory.js
```diff
@@ -0,0 +1,68 @@
...
+            {props.fields[key].type === 'checkbox' ? null:
+              <ControlLabel className={props.fields[key].required ? 'required-pf': null}>{props.fields[key].title}</ControlLabel>}
+          </Col>
+          <Col sm={_.get(props.fields[key], 'labelSize') ? 8-props.fields[key].labelSize : 5}>
+            {child}
+            <HelpBlock>{validMsg ? validMsg: null}</HelpBlock>
+          </Col>
+        </FormGroup>);
+  })
+}
```
suomiy: newline - better to setup an IDE to do it for you ..

src/utils/forms/FormFactory.js
```diff
@@ -0,0 +1,68 @@
...
+    ButtonGroup,
+    DropdownButton,
+    MenuItem,
+    Checkbox
+  } from 'patternfly-react';
+
+export const FormFactory = (props) => {
+  return Object.keys(props.fields).map(key => {
+    if (props.fields[key].isVisible && props.fields[key].isVisible(props.fieldsValues) === false){
+      return undefined;
```
suomiy: just `return;` is enough

src/k8s/request.js
```diff
@@ -0,0 +1,115 @@
...
+        template: {}
+      }
+  }
+  if(basicSettings.namespace){
+    vm.metadata.namespace = basicSettings.namespace.value;
+  }
+  if(basicSettings.description){
+    vm.metadata.labels.description = basicSettings.description;
+  }
+  addFlavor(vm, basicSettings);
```
suomiy: I think few newlines here and there add to the readability, but that is just my opinion.

src/constants/index.js
```diff
@@ -0,0 +1,4 @@
+export const API_VERSION='kubevirt.io/v1alpha2'
+export const VM_KIND='VirtualMachine'
+export const OS_LABEL='kubevirt.io/os'
+export const FLAVOR_LABEL='kubevirt.io/flavor'
```
suomiy: \n

src/components/wizards/new-vm/NewVmWizard.js
```diff
@@ -0,0 +1,61 @@
...
+            </Wizard.Body>
+            <Wizard.Footer>
+              <Button bsStyle="default" className="btn-cancel" onClick={this.props.closeWizard}>Cancel</Button>
+            </Wizard.Footer>
+          </Wizard>
+          {wizard}
+        </React.Fragment>
+      );
+    }
+}
```
suomiy: Are we using React Proptypes in this project or not ?

src/components/wizards/new-vm/NewVmWizard.js
ohadlevy: there is a pf-react Button Component?

src/components/wizards/new-vm/NewVmWizard.js
```diff
@@ -0,0 +1,61 @@
...
+      return (
+        <React.Fragment>
+          <Wizard show onHide={this.props.closeWizard}>
+            <Wizard.Header title="Create Virtual Machine" onClose={this.props.closeWizard} />
+            <Wizard.Body>
+              <Wizard.Row>
+                <Wizard.Main>
+                  <Wizard.Contents stepIndex={0} activeStepIndex={0} className="wizard-content">
+                    <button className="btn btn-link wizard-button " onClick={this.openCreateVmWizard}>
+                      <div className="col-md-12">
```
ohadlevy: use Col component ?

src/components/wizards/new-vm/NewVmWizard.js
```diff
@@ -0,0 +1,61 @@
...
+                        <span className="pficon-virtual-machine fa-5x" />
```
ohadlevy: use Icon component?

src/components/wizards/new-vm/NewVmWizard.js
ohadlevy: we should, :+1:

src/components/wizards/new-vm/NewVmWizard.stories.js
```diff
@@ -0,0 +1,93 @@
...
+import { NewVmWizard } from './NewVmWizard';
+import { Button } from 'patternfly-react';
+import { os } from './OperatingSystem';
+
+class NewVmWizardExample extends React.Component {
+  state = {
+    show: false
+  }
+
+  storageClasses = [
```
ohadlevy: maybe move to a fixture files? could be reused in tests?

lizsurette:
@rawagner 

Agreed with others, this is looking great! A few overall comments:

- When the user selects a provision source, what do you think about removing the indentation for any fields that are added? I think it will read better if they are inline. (Similar to how the fields show when the Enable cloud-init box is checked.

- Would it be possible to split the OS field into Operating System and Version? I got some feedback that this list could get super long if we leave it as one. Of course this would mean that depending on which OS is selected, the versions would need to be updated.

- Although it might be annoying to need two different styles, could we tighten up the padding between the check box options at the bottom of the Basic Settings steps? Maybe just allow them to live within one form-group if possible?

src/components/wizards/create-vm/CreateVMWizard.js
vojtechszocs: Yes, this is also the syntax form preferred by patternfly-react project.

ohadlevy:
note its used below, so  I guess it should be used consistently.

src/components/wizards/create-vm/CreateVMWizard.js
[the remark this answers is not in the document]
vojtechszocs: I see, I'd rename them to better reflect their purpose, i.e. `this.basicStepFields` and `this.state.basicStepValues` or similar.

I'd use the word "step" since it's the grouping concept that puts multiple form fields together, from visual/user perspective.

src/components/wizards/new-vm/NewVmWizard.js
rawagner: I will use Button from pf-react and also extract the whole 'button with icon and label' to separate component

src/components/wizards/new-vm/NewVmWizard.js
suomiy: @vojtechszocs @mareklibra do you think it is better to use components or simple classes ?

src/components/wizards/new-vm/NewVmWizard.stories.js
vojtechszocs: I agree, in React world, `onSomething` is a standard convention for callback props.

src/components/wizards/new-vm/NewVmWizard.js
vojtechszocs: Yes, let's reuse patternfly-react's `Button` if possible.

src/components/wizards/new-vm/NewVmWizard.js
vojtechszocs: Of course we should use them :smiley: maybe not right now, since we're prototyping our first complex wizard component, but we should use them at some point.

src/components/wizards/create-vm/CreateVmWizard.js
```diff
@@ -0,0 +1,228 @@
+import React from 'react';
+import { 
+  Form,
+  Wizard
+} from 'patternfly-react';
+import './CreateVmWizard.css';
+import * as _ from 'lodash';
```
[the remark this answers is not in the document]
vojtechszocs: IIRC there are webpack plugins to address these kinds of optimization issues, I will look into that later on.

ohadlevy:
there is, but i still think its a better practice to get used to instead of
importing all of it.

src/components/wizards/create-vm/CreateVmWizard.js
```diff
@@ -0,0 +1,228 @@
...
+    state = {
+      activeStepIndex: 0,
+      basicVmSettings: {},
+      wizardValid: false
+    }
+  
+    onFormChange = (newValue, target) => {
+      let validMsg = null;
+      if (this.formFields[target].required && newValue.trim().length === 0){
+        validMsg = `${this.formFields[target].title} is required`;
```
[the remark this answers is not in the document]
vojtechszocs: > this is not i18n friendly, @vojtechszocs add to the list of project infra?

Added that to the list of infra tasks.

We should use the standard ICU message format. In `ovirt-engine-ui-extensions` project, we're using [intl-messageformat](https://github.com/yahoo/intl-messageformat) - every localized message is a function that accepts zero or more parameters, returning a localized string.

> overall, i feel this should probably live outside of the component in some i18n or forms related code?

Yes, i18n is another concern and should be put into its dedicated module.

src/components/wizards/create-vm/CreateVmWizard.js
```diff
@@ -0,0 +1,228 @@
...
+    }
+  
+    validateWizard = values => {
+      let wizardValid = true;
+
+      //check if all required fields are defined
+      const requiredKeys = Object.keys(this.formFields).filter(key => this.isFieldRequired(key, values));
+      const requiredKeysInValues = Object.keys(values).filter(key => this.isFieldRequired(key, values));
+      if(requiredKeys.length !== requiredKeysInValues.length){
+        wizardValid = false;
```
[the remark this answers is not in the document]
vojtechszocs: Thanks @ohadlevy for the links!

`finaly-form` is React agnostic, implements observer pattern to subscribe to form/field changes and allows their updates.

`formik` is React specific, and from quickly looking at their examples, is geared towards HTML `<form>` handling, i.e. the case when you use DOM form/field elements directly.

suomiy:
> Would it be possible to split the OS field into Operating System and Version? I got some feedback that this list could get super long if we leave it as one. Of course this would mean that depending on which OS is selected, the versions would need to be updated.

It makes sense, but I would leave it until we know that we get this kind of information from the API - which will be preset os labels.

src/components/wizards/new-vm/NewVmWizard.js
vojtechszocs: Yup, we should use patternfly-react's `Grid`, `Row` and `Col` components instead of plain HTML/CSS markup.

Also, at some point, we'll transition to patternfly-react v4 which will implement its own grid system.

ohadlevy:
not before tectonic migration please..

src/components/wizards/create-vm/CreateVmWizard.js
```diff
@@ -0,0 +1,228 @@
...
+        default: '--- Select Workload Profile ---',
+        values: this.props.workloadProfiles,
+        required: true
+      },
+      startVM:{
+        title: 'Start virtual machine on creation',
+        type: 'checkbox'
+      },
+      createTemplate:{
+        title: 'Create new teplate from configuration',
```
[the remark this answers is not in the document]
rawagner: fixed

src/components/wizards/new-vm/NewVmWizard.stories.js
rawagner: renamed to onHide

src/components/wizards/new-vm/NewVmWizard.js
rawagner: Moved to separate component `ButtonWithIcon` which is using `Button`, `Icon` and `Col` pf-react components

src/utils/forms/FormFactory.js
rawagner: fixed

src/k8s/request.js
rawagner: added a few new lines here and there :)

src/utils/forms/FormFactory.js
rawagner: this results in eslint warning. Anyways I avoided it by filtering the fields which are not visible and than map only the visible fields.

src/components/wizards/create-vm/CreateVmWizard.js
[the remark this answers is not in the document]
rawagner: changed all `import * as _ from 'lodash` to more precise ones (get/has etc)

src/components/wizards/create-vm/CreateVmWizard.js
```diff
@@ -0,0 +1,228 @@
...
+} from 'patternfly-react';
+import './CreateVmWizard.css';
+import * as _ from 'lodash';
+import { StorageTable } from './StorageTable';
+import { createVM } from '../../../k8s/request';
+import { FormFactory } from '../../../utils/forms/FormFactory';
+
+export class CreateVmWizard extends React.Component {
+
+    state = {
```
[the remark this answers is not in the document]
rawagner: no, it should not. It is a class property which is not 'prefixed' with const

src/components/wizards/create-vm/CreateVmWizard.js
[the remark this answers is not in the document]
rawagner: I will wait for infra updates then

src/components/wizards/new-vm/NewVmWizard.js
rawagner: pf-react `Button`, `Col` and `Icon` used instead

rawagner:
@lizsurette 

> When the user selects a provision source, what do you think about removing the indentation for any fields that are added? I think it will read better if they are inline. (Similar to how the fields show when the Enable cloud-init box is checked.

No problem, I saw it in earlier version of Create VM design. It is now removed, less work for me :)


> Would it be possible to split the OS field into Operating System and Version? I got some feedback that this list could get super long if we leave it as one. Of course this would mean that depending on which OS is selected, the versions would need to be updated.

as @suomiy said, lets wait for API updates, we will change it then. I agree that if we should list all possible values into one dropdrown, it would be a huge one..

> Although it might be annoying to need two different styles, could we tighten up the padding between the check box options at the bottom of the Basic Settings steps? Maybe just allow them to live within one form-group if possible?

I will, TODO

src/components/wizards/new-vm/NewVmWizard.stories.js
rawagner: sounds good, TODO

rawagner:
@lizsurette 

> Although it might be annoying to need two different styles, could we tighten up the padding between the check box options at the bottom of the Basic Settings steps? Maybe just allow them to live within one form-group if possible?

checkboxes are tighter now. You can take a look if you want, storybook is updated.

lizsurette:
> No problem, I saw it in earlier version of Create VM design. It is now removed, less work for me :)

Latest update looks great! Glad it's even less work, too :)


> as @suomiy said, lets wait for API updates, we will change it then. I agree that if we should list all possible values into one dropdrown, it would be a huge one..

Makes sense to wait on API updates for this.


> checkboxes are tighter now. You can take a look if you want, storybook is updated.

Looks great! Thanks :)

src/components/wizards/import-vm/ImportVmWizard.js
```diff
@@ -0,0 +1,74 @@
...
+    wizardStepsImportVM = [
+      {
+        title: 'Basic Information',
+        render: () => <Form horizontal>
+          <FormGroup>
+            <Col sm={3} className="text-right">
+              <ControlLabel className="required-pf">Import Source Type</ControlLabel>
+            </Col>
+            <Col sm={4}>
+              <DropdownButton bsStyle="default" id="import-type-dropdown" title="--- Choose Import Source Type ---">
```
lizsurette: This should be "Select" rather than "Choose" to be consistent with the new VM side of things.

src/components/wizards/import-vm/ImportVmWizard.js
```diff
@@ -0,0 +1,74 @@
...
+  
+    wizardStepsImportVM = [
+      {
+        title: 'Basic Information',
+        render: () => <Form horizontal>
+          <FormGroup>
+            <Col sm={3} className="text-right">
+              <ControlLabel className="required-pf">Import Source Type</ControlLabel>
+            </Col>
+            <Col sm={4}>
```
lizsurette: Noticed this field isn't the same length as the rest of them in the list. Is it possible for them to align?

src/components/wizards/import-vm/ImportVmWizard.js
rawagner: Changed to Select

src/components/wizards/import-vm/ImportVmWizard.js
rawagner: length is the same now. The Import wizard is just a sketch, it has no functionality yet. Lets not spent any more time.

src/components/wizards/import-vm/ImportVmWizard.js
rawagner: Changed to Select

src/components/wizards/import-vm/ImportVmWizard.js
lizsurette: > length is the same now. The Import wizard is just a sketch, it has no functionality yet. Lets not spent any more time.

Understood. Thanks!!

src/components/buttons/ButtonWithIcon.js
```diff
@@ -0,0 +1,16 @@
...
+    <Button onClick={props.onClick} bsClass="btn btn-link btn-with-icon">
+      <Col md={12}>
+        <Icon type={props.iconType} name={props.icon} className="fa-5x"/>
+      </Col>
+      <Col md={12} className="label-column">
+        <span className="lead">{props.label}</span>
+      </Col>
+    </Button>
+  );
+}
```
ohadlevy: 1. please use extract the used props, imho it increases readability, you can use lebab to do a lot of this for you.
2. please add prop-types

the output should look something like
```js
import React from 'react';
import { Button, Icon, Col } from 'patternfly-react';
import './ButtonWithIcon.css';

export const ButtonWithIcon = ({ onClick, iconType, icon, label }) => (
  <Button onClick={onClick} bsClass="btn btn-link btn-with-icon">
    <Col md={12}>
      <Icon type={iconType} name={icon} className="fa-5x" />
    </Col>
    <Col md={12} className="label-column">
      <span className="lead">{label}</span>
    </Col>
  </Button>
);

ButtonWithIcon.propTypes = {
  onClick: PropTypes.func.isRequired,
  iconType: PropTypes.string.isRequired,
  icon: PropTypes.string.isRequired,
  label: PropTypes.string.isRequired
};
```

rawagner: @ohadlevy fixed

src/components/forms/FormFactory.js
```diff
@@ -0,0 +1,97 @@
...
+export const FormFactory = ({ fields, fieldsValues, onFormChange }) => {
+  const formGroups = Object.keys(fields)
+    .filter(key => !(fields[key].isVisible && fields[key].isVisible(fieldsValues) === false))
+    .map(key => {
+      let child;
+      const validMsg = get(fieldsValues[key], 'validMsg', null);
+      switch (fields[key].type) {
+        case 'textarea':
+          child = (
+            <FormControl
```
ohadlevy: does it make sense to extract each case statement to its own component? i could see reusing them in possibility in other context + i believe it would be easier to read?

ohadlevy:
can you please rebase? so travis tests can be enabled for the repo? thanks

ohadlevy:
opened/closed to trigger CI.

rawagner:
Added some enzyme tests.

@ohadlevy any suggestions on what should we use for code coverage metrics ?

ohadlevy:
@rawagner https://github.com/kubevirt/web-ui/pull/16

src/components/forms/FormControls.js
```diff
@@ -0,0 +1,62 @@
...
+  <FormControl componentClass="textarea" value={value} onChange={event => onChange(event.target.value, fieldKey)} />
+);
+
+TextAreaControl.propTypes = {
+  fieldKey: PropTypes.string.isRequired,
+  value: PropTypes.string.isRequired,
+  onChange: PropTypes.func.isRequired
+};
+
+export const DropdownControl = ({ fieldKey, value, choices, onChange }) => (
```
ohadlevy: I personally prefer a component in a file, it makes it easier to find, maybe just change them to forms/textArea, forms/checkBox etc, would be trivial to find ?

src/components/forms/__tests__/__snapshots__/FormFactory.test.js.snap
```diff
@@ -0,0 +1,186 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ohadlevy: I've found that snapshots over 50 lines is usually ignored, if some of the internal snaps are already part of other snapshots maybe they could be avoided?

rawagner: reduced to < 50 lines.`NewVmWizard.test.js.snap` still has 66 lines but I do not see a way to reduce it.

src/components/forms/controls/index.js
```diff
@@ -0,0 +1,6 @@
+import { TextAreaControl } from './TextAreaControl';
+import { TextControl } from './TextControl';
+import { DropdownControl } from './DropdownControl';
+import { CheckboxControl } from './CheckboxControl';
+
+export { TextAreaControl, TextControl, DropdownControl, CheckboxControl };
```
ohadlevy: you can also use the following syntax:
```js
export { TextAreaControl } from './TextAreaControl';
```

src/components/wizards/create-vm/CreateVmWizard.js
```diff
@@ -0,0 +1,233 @@
...
+  };
+
+  onFormChange = (newValue, target) => {
+    let validMsg;
+
+    if (this.basicFormFields[target].validate) {
+      validMsg = this.basicFormFields[target].validate(newValue);
+    }
+    if (this.basicFormFields[target].required && newValue.trim().length === 0) {
+      validMsg = 'is required';
```
ohadlevy: I wonder if we should do something with the strings to make it easier later on to i18n.

rawagner: maybe we could address that in followup patch ? :) Im not sure what solution for i18l we will want to use. It would be nice to keep aligned with openshift console but they do not have any i18l yet.

rawagner:
Latest PR updates code coverage configuration to exclude *.stories.js files

package.json
```diff
@@ -79,7 +82,8 @@
   "jest": {
     "collectCoverage": true,
     "collectCoverageFrom": [
-      "src/**/*.{js,jsx,mjs}"
+      "src/**/*.{js,jsx,mjs}",
+      "!src/**/*.stories.js"
```
ohadlevy: :+1: maybe we can just PR this right away?

ohadlevy:
@rawagner thanks, it looks like there is a snapshot out of date + css linting errors in the tests.

src/components/wizards/create-vm/CreateVmWizard.js
vojtechszocs: The `openshift/console` project (frontend part) currently doesn't seem to use any i18n mechanism, there are English strings hardcoded in their components.

I'd like to avoid adding complexity until we really need it. For now, I would focus on writing simple, reusable components (and break down bigger components into smaller bits for that matter) along with associated `.test.js` and `.stories.js` files.

src/components/forms/__tests__/__snapshots__/FormFactory.test.js.snap
vojtechszocs: > I've found that snapshots over 50 lines is usually ignored, if some of the internal snaps are already part of other snapshots maybe they could be avoided?

This is true for snapshots and other files too. If it looks big and complex, people are less inclined to review and maintain it, until it breaks the build or application itself.

This has already been discussed on `patternfly-react` project, see [my comment on Enzyme & snapshot testing](https://github.com/patternfly/patternfly-react/pull/268#issuecomment-375723540).

In particular:

> as @dmiller9911 pointed out, snapshots shouldn't include fully-resolved HTML tree, otherwise you'll end up with big snapshots and it's hard to differentiate the actual component's HTML bits

vojtechszocs:
I've got some infra changes coming up that will help formalize the project structure and conventions, including tools for validation and consistent usage of other tools (like using one Babel config for webpack, Jest and Storybook).

suomiy:
I refactored `FormFactory.js`, FormControls and `CreateVmWizard.js` in the [followup](#10). So if somebody notices something in there, we can discuss it in #10 :) and proceed with merging this PR.

rawagner:
deprecated by https://github.com/kubevirt/web-ui-components/pull/27

PR #6, opened 2018-08-30
Add JS and CSS linting support

ohadlevy:
JS linting is based on the patternfly-react eslint package

I also added an npm package for storybook deployer so we can easily share PR reviews.

This include fixing all the linting errors as well :)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,17 @@
+const path = require('path');
+
+module.exports = {
+  root: true,
+  extends: ['plugin:patternfly-react/recommended'],
+  rules: {
+    'import/first': 'off'
+  },
+  overrides: [
+    {
+      files: ['**/__mocks__/**', '**/Stories/**', '*.stories.js', '*.test.js'],
+      rules: {
+        'import/no-extraneous-dependencies': 'off'
+      }
+    }
+  ],
+};
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,8 @@
+{
+  "semi": true,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "none",
+  "useTabs": false,
+  "printWidth": 120
+}
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,0 +1,4 @@
+{
+  "defaultSeverity": "warning",
+  "extends": ["stylelint-config-standard"]
+}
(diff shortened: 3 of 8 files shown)


vojtechszocs:
cc @jelkosz @mareklibra @rawagner 

Looks good overall, I'd suggest to proceed with this PR as-is and do any adjustments in subsequent PRs.

I'm currently working on a PR that improves various aspects of the project infrastructure, I can rebase my work on top of this.

@ohadlevy some comments:

As for using [Prettier](https://prettier.io/), I don't think the added value (being able to automatically fix wrongly formatted code) is worth the added complexity (extra dependency, `.prettierrc` and ESLint config assumption), but I guess the main reason here is that "`eslint-plugin-patternfly-react` assumes Prettier, so we should use it too". Instead, I'd suggest using an [EditorConfig](https://editorconfig.org/#file-format-details) file, which is supported by many IDEs, but most importantly, which puts *humans* in charge of fixing wrongly formatted code. In my opinion, doing that upon PR merge by some automated tool is generally a bad idea.

Next up is `eslint-plugin-patternfly-react` [config](https://github.com/patternfly/patternfly-react/blob/master/packages/eslint-plugin-patternfly-react/lib/config/recommended.js) which overlays [StandardJS](https://standardjs.com/) rules with [AirBnb](https://github.com/airbnb/javascript) rules. To me, this seems a bit weird, considering both of these are designed to form a *baseline* of app-specific linting rules. We'd like to use StandardJS code style in this project, so we'd have to declare it on top of AirBnb, which sits on top of StandardJS. There's also the issue of `eslint-plugin-patternfly-react` making assumption about using Prettier. Overall, I'd just take the relevant bits out of `eslint-plugin-patternfly-react` instead of using it directly. (And maybe consider breaking patternfly-react's config into smaller, modular bits to improve reusability and remove hidden assumptions like Prettier.)

`storybook-deployer` looks like a nice tool, so +1 on that one!

Also note that `create-react-app` generated project structure executes ESLint as part of webpack build :smile: this will be changed in my PR that I'm working on. Different concerns shouldn't be tangled together, for example, you should be able to lint the sources *without* running webpack build. Same thing for running tests etc.

ohadlevy:
> Looks good overall, I'd suggest to proceed with this PR as-is and do any adjustments in subsequent PRs.

rawagner:
LGTM

vojtechszocs:
OK, let's merge this and discuss any further improvements in follow-up PRs.

Issue #9, opened 2018-08-31
Double click needed for changing focused cells in tables


suomiy:
Reactabular always creates new element when value changes in the cell (React key doesn't help). 

Buggy sequence: 
- edit row
- focus text input
- write value
- click on different cell
- onBlur sets react state
- reactabular rerenders inputs and creates brand new elements and they do not get focus from the click

https://github.com/reactabular/reactabular/issues/346, probably will be fixed in [reactabular 9](https://github.com/reactabular/reactabular/pull/345)

suomiy:
This could be fixed by remembering which cell is edited and capturing onClick events and using autofocus in our inputs. It is quite hacky and it would not be possible to use existing components much (like pf-react FormControl).

this approach is used in this repo: https://github.com/reactabular/react-edit/blob/master/src/input.js

Issue #8, opened 2018-08-31
Edit Buttons positioned incorrectly in modal tables


suomiy:
caused by https://github.com/patternfly/patternfly-react/issues/535

suomiy:
Issue still present and in web-ui, buttons are on the wrong side

PR #10, opened 2018-08-31
Create vm wizard - Storage

suomiy:
This is build on top of #4 and there are currently two unfixable bugs for the table: #8 , #9
- [published storybook](https://clever-allen-639701.netlify.com/?selectedKind=Wizards&selectedStory=New%20VM%20Wizard&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

I created EditableDraggableTable component which can also be used for the Network; supports:
- inline editing with custom inner components
- drag and drop
- renders custom definitions from `columns.renderConfigs`


suomiy:
@rawagner 
- I needed to use FormFactory in other way so I refactored it. dunno if you want the changes from that commit into #4 or if it is enough to merge just with this one

- can you plan better handling and passing the values and validation to `CreateVmWizard`. I  am just passing disks and validation back to `CreateVmWizard` for now. I think it would be better to move the first step to its own component

suomiy:
@lizsurette
- Deleting wasn't described in the design doc. Please look at the storybook and let me know what you think about it. I just added delete button for each row (which is disabled when editing)

note: ConfirmButtons [X] is only used to discard edit changes 

- also we need to let user somehow know that the first disk in the first row will be bootable. Can we add some label behind it's name or annother column?

jelkosz:
@suomiy there should be some validation which disallows you to create a disk without the required params.

lizsurette:
> Deleting wasn't described in the design doc. Please look at the storybook and let me know what you think about it. I just added delete button for each row (which is disabled when editing)

Good catch! I like what you've done, but I think to stay consistent with the other list/table views in the app we should probably use the kebab with an option to Delete in each row. Thoughts on this?
screen shot 2018-08-31 at 1 36 14 pm


> also we need to let user somehow know that the first disk in the first row will be bootable. Can we add some label behind it's name or annother column?

Yeah, I think simply adding something like "(bootable)" after the name would be a good start. I am showing this in the mockup above, too. What do you think?

ohadlevy:
@suomiy can you rebase please?

suomiy:
@lizsurette I fixed bootable and delete  kebap. Can you please look at it? 

validations are not fixed yet

suomiy:
fixed all the issues (except tests):
- validations and transition between steps (jumping between steps)
- results page (not sure if needed after seeing the docs).
- refactoring
- requests.js details

Bumped up `"patternfly-react": "^2.17.3"` to include `previousStepDisabled`

Updated storybook; 
- note: create vm with name: `fail` to see the mockup request fail

rawagner:
@suomiy would it be possible to show some validation error when disk is missing required fields ? (name/size is empty). It only disables the Create Virtual Machine button without any explanation.

Also when editing disk, I can click on menu button (three dots) but Remove Disk menu item is disabled. It would be nicer if the whole menu button is disabled.

suomiy:
> @suomiy would it be possible to show some validation error when disk is missing required fields ? (name/size is empty). It only disables the Create Virtual Machine button without any explanation.

@lizsurette where would be the best to show such validation? We need to let the user know which row and column is wrong

> Also when editing disk, I can click on menu button (three dots) but Remove Disk menu item is disabled. It would be nicer if the whole menu button is disabled.

@lizsurette  should we disable it or is there a possibility to show other action when editing?

src/components/wizards/create-vm/ResultTab.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Alert } from 'patternfly-react';
+
+import './ResultTab.css';
+
+const ResultTab = ({ result, successfull }) => {
```
rawagner: successfull -> successful. Other occurrences has also this typo.

rawagner: or just call it success, its shorter :)

lizsurette:
> @lizsurette where would be the best to show such validation? We need to let the user know which row and column is wrong

I think having a similar style to our normal inline validation error would be good. It would make the row grow a bit vertically, but I think that's fine. Something like:
screen shot 2018-09-13 at 10 13 43 am


> @lizsurette should we disable it or is there a possibility to show other action when editing?

I'd say remove the kebab all together when editing.

lizsurette:
> @lizsurette I fixed bootable and delete kebap. Can you please look at it?

@suomiy I'm not sure if I'm looking in the wrong place, but the storybook still shows the trash icon rather than a kebab for me... would you mind sharing an updated link if it's changed?

suomiy:
@lizsurette 
I am not sure what is wrong, it is working here. Can you try to open it in incognito window? 

link: https://clever-allen-639701.netlify.com/

suomiy:
> I think having a similar style to our normal inline validation error would be good. It would make the row grow a bit vertically, but I think that's fine. Something like:

Ok, I will make these validations visible even when editing is off

ohadlevy:
it looks like css is a bit off 
image

src/components/forms/Number.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormControl } from 'patternfly-react';
+import './Dropdown.css';
+import { eventValueHandler } from './utils';
+
+const onKeyDown = e => {
+  if (!((e.keyCode > 95 && e.keyCode < 106) || (e.keyCode > 47 && e.keyCode < 58) || e.keyCode === 8)) {
```
ohadlevy: please extract keyboard codes to consts, and use meaningful numbers, also I'm pretty sure PF-react has already some of these defined (maybe can import it)

src/components/forms/Number.js
```diff
@@ -0,0 +1,44 @@
...
+    e.preventDefault();
+    return false;
+  }
+  return true;
+};
+
+export const Number = ({ id, value, defaultValue, onChange, onBlur, positive }) => (
+  <FormControl
+    id={id}
+    type="number"
```
ohadlevy: I assume this mostly works on newer browsers? (not that I mind) but just wondering what is the minimum browser we support?

ohadlevy: also, maybe this makes sense to have this component in pf-react.

src/components/forms/utils.js
```diff
@@ -0,0 +1,7 @@
+export const valueHandler = callback => (typeof callback === 'function' ? value => callback(value) : null);
+
+export const eventValueHandler = (callback, defaultValue = '') =>
+  typeof callback === 'function' ? event => callback(event.target.value || defaultValue) : null;
+
+export const checkboxHandler = (callback, defaultValue = false) =>
```
ohadlevy: does it makes sense to document these utils?

src/components/table/EditableDraggableTable.css
```diff
@@ -0,0 +1,11 @@
+table.pf-table-inline-edit tbody tr.editing td {
+  background-color: #def3ff; /* color-pf-blue-50 */
```
ohadlevy: looks like we need scss support in this repo?

src/components/table/EditableDraggableTable.js
```diff
@@ -0,0 +1,319 @@
...
+//   Stateless function components cannot be given refs. Attempts to access this ref will fail.
+// eslint-disable-next-line react/prefer-stateless-function
+class InlineEditRow extends React.Component {
+  render() {
+    return <Table.InlineEditRow {...this.props} />;
+  }
+}
+
+// eslint-disable-next-line react/no-multi-comp
+class EditableDraggableTable extends React.Component {
```
ohadlevy: please one react component per file?

lizsurette:
> @lizsurette
> I am not sure what is wrong, it is working here. Can you try to open it in incognito window?
> 
> link: https://clever-allen-639701.netlify.com/
All is well now! Maybe a caching issue? Thanks for the help :)

This is looking great and I just wanted to add a few other comments/suggestions. Here is a screenshot of what I'm seeing at the moment:
screen shot 2018-09-13 at 1 43 01 pm

1) I'm noticing the same CSS issue that Ohad pointed out with the save/cancel buttons. Seems like depending on where you scroll it will display in a different spot each time.

2) After choosing to Create a Disk, clicking the "Cancel" action should actually remove the line from the table rather than add a blank one.

3) Instead of "Select" could we change the text to "Select Storage" when attaching storage?

src/components/table/EditableDraggableTable.css
suomiy: @vojtechszocs

src/components/forms/utils.js
suomiy: I don't think that is necessary

suomiy:
everything should be fixed now

@lizsurette 

> 1. I'm noticing the same CSS issue that Ohad pointed out with the save/cancel buttons. Seems like depending on where you scroll it will display in a different spot each time.

depends on #8

> 2. After choosing to Create a Disk, clicking the "Cancel" action should actually remove the line from the table rather than add a blank one.

I think that would be confusing. Cancel action should discard the current changes. We have a delete action for that. Patternfly [examples](https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?knob-Bordered=true&knob-Edit%20whole%20table=true&selectedKind=patternfly-react%2FContent%20Views%2FTable%20View&selectedStory=Table%20With%20Inline%20Edit%20Row&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) behave in the same way

> 3. Instead of "Select" could we change the text to "Select Storage" when attaching storage?

done

suomiy:
updated storybook

suomiy:
> "Add an item to this list, no wait, I don't want to do that...cancel" which I think should result in the row going away. We can do some further user testing on it if you still feel like it's a weird interaction. 

I think user also wants to edit the vm  and then decides that he/she would like to discard the changes.

> o wait, I don't want to do that...cancel

I don't think that it is such a problem to click on a delete action after that + it is more consistent behaviour, when user sees these buttons somewhere else and is used to discarding functionality

lizsurette:
> I think user also wants to edit the vm and then decides that he/she would like to discard the changes.

Agreed. The way this works today is great.

> I don't think that it is such a problem to click on a delete action after that + it is more consistent behaviour, when user sees these buttons somewhere else and is used to discarding functionality

I'm fine with this to start. If we get user feedback that it is confusing, we can update later.

suomiy:
completed tests

suomiy:
I tried passing the generated output to the CLI, here are the results:

### Attaching Storage 

- Works fine

### Creating Disk (DataVolume) 

- Problematic, because it seems there has to be a source for the volume (`HTTP` or `S3`)
- It works fine with the source. 

- tested on  `kubevirt v0.8.0`
- it gives no errors without a source, but the the persistent volume claim is not created and VM not started. 

### Sources
- [User Guide](https://kubevirt.io/user-guide/#/workloads/virtual-machines/disks-and-volumes?id=datavolume)
- See `source`=`required` in [v1alpha1.DataVolumeSpec](https://kubevirt.io/api-reference/master/definitions.html#_v1alpha1_datavolumespec). 

I don't know if this is gonna change in future releases, but for now it seems we need to also select the source in the ui when creating new disks.  Does anybody have info on that? Or did I miss something?

src/components/forms/Number.js
suomiy: We should be alligned with tectonic console. This is what their github page says:

```
Supported Browsers
We support the latest versions of the following browsers:

Edge
Chrome
Safari
Firefox
IE 11 and earlier is not supported.
```

So I guess we are supporting just the latest browsers.

suomiy: I am working on the pf-react component

suomiy:
- rebased
- added final refactoring
- split to more commits
- fixed tests
- disabled Create Disk button (until blank disk feature is supported by kubevirt)

suomiy:
Cosmos preview available here: https://clever-allen-639701.netlify.com/

package.json
```diff
@@ -33,8 +33,8 @@
   "peerDependencies": {
     "patternfly-react": "2.x",
     "prop-types": "15.x",
-    "react": "16.5.x",
-    "react-dom": "16.5.x"
+    "react": "16.3.x",
```
priley86: hi @suomiy - are you running into challenges running React `16.5.x` against `patternfly-react`? If that is the case, it would be nice to bubble these issues upstream. Otherwise, what is the problem w/ `16.5`?

src/components/Form/Integer.js
```diff
@@ -0,0 +1,130 @@
...
+};
+
+export const Integer = ({ id, value, defaultValue, onChange, onBlur, positive, nonNegative }) => {
+  let allowedKeys;
+  let validRegex;
+  let fixAfterValue;
+  let min;
+
+  if (positive) {
+    allowedKeys = NON_NEGATIVE_INTEGER_KEYS;
```
priley86: can i ask the original source of these functions? I am just noting that there is some additional validations here that may not be present in other libraries (so it would be nice to know the distinctions made). For example, [Formik](https://github.com/jaredpalmer/formik/blob/master/src/utils.ts#L98) has very simple assertion functions. Does it make sense to self implement these or use a 3rd party? I'm fine w/ this, but would just like to know the API surface is well supported.

priley86: My primary concern w/ form validations / form HOCs such as these is that they will:

1. Be flexible to different variation
2. Will support local state management. Many times in the past I have the feeling that redux-form is creating an unnecessary burden, and that form state should be locally managed (like you have done here).

src/components/Table/EditableDraggableTable.js
```diff
@@ -0,0 +1,341 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { DragDropContext } from 'react-dnd';
```
priley86: this is a nice enhancement. It may make sense to contribute this to PF React in the future. thanks for the work on this!

suomiy:
I split this into two other PRs #37 and #38

src/components/Wizard/NewVmWizard/fixtures/NewVmWizard.fixture.js
```diff
@@ -20,6 +20,50 @@ export const namespaces = [
...
+    name: 'AzureDisk',
+    id: 'azuredisk'
+  },
+  {
+    name: 'Local',
+    id: 'local'
+  }
+];
+
+export const storages = [
```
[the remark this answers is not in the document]
suomiy: yes, fixed now - and integration with web-ui is here (https://github.com/suomiy/console/tree/disks-tab-integration)

package.json
suomiy: There is no problem with 16.5. We just need to allign it with https://github.com/openshift/console/blob/master/frontend/yarn.lock because these react versions are incompatible and can crash.

src/components/Form/Integer.js
suomiy: I started from some SO question like this [one](https://stackoverflow.com/questions/5535449/javascript-keycode-allow-number-and-plus-symbol-only) (don't remember which exact one was it) and wrote it into the current form. 

I looked at Formik but it doesn't have anything for this use-case. 
- I wanted to have a field which would allow user to press only the valid keys without the need to show errors (i.e. positive integer without 0). 
- We could use formik components and validation lifecycle but I am not sure if it is worth it. Formik doesn't help with the validation itself.

I could write the component in more generic way. There was a suggestion to submit it to pf-react. Do you think that is a good idea?

btw: you can try the component [here](https://clever-allen-639701.netlify.com/?component=Form%2FG&fixture=Integer&editor=true)

suomiy:
@lizsurette 
could you please look at the result tab in [Create VM Wizard](https://clever-allen-639701.netlify.com/?component=Wizard%2FCreateVmWizard%2FB&fixture=CreateVmWizard) and tell me what do you think about the design?

There was one opinion that we could show the VM JSON collapsed under "Created VM" or something and show it on demand.

lizsurette:
@suomiy I like that the results step has something on it rather than just a "success" message. I do wonder what the use case for having this JSON would be. I'm assuming a user could take this and either use it via the CLI or share it with someone else to do the same? A few additional comments:

- Minor typo on "successful", drop the last "l".
- If we keep the code snippet, we should probably label this box. Should this really be YAML rather than JSON for future use when creating VMs via the UI? 
- If the use case is to copy/paste to a YAML file, should we alternatively just allow the user to "download the YAML" or at the very least allow for an easy Copy function?
- I don't think the "Cancel" or "Back" actions make sense to keep at this point. Any way we can remove them for this step?

Here are a few mockups with these ideas:
results 1

results 2


Happy to chat further, but I like the direction!

suomiy:
> @suomiy I like that the results step has something on it rather than just a "success" message. I do wonder what the use case for having this JSON would be. I'm assuming a user could take this and either use it via the CLI or share it with someone else to do the same? A few additional comments:

> If the use case is to copy/paste to a YAML file, should we alternatively just allow the user to "download the YAML" or at the very least allow for an easy Copy function?


I think the most common use case for a user would be to look at the JSON/YAML to undestand what was actually created and not passing it somewhere else.

- There was a concern that the generated JSON could look like an error message from afar. That is why I asked about having it collapsed by default.
- We also don't want the user to think that he has to copy it somewhere, because it was already created.

> If we keep the code snippet, we should probably label this box. Should this really be YAML rather than JSON for future use when creating VMs via the UI?

We could support YAML in a followup


> Minor typo on "successful", drop the last "l".
> I don't think the "Cancel" or "Back" actions make sense to keep at this point. Any way we can remove them for this step?

Thanks, I will look at it.

rawagner:
> > @suomiy I like that the results step has something on it rather than just a "success" message. I do wonder what the use case for having this JSON would be. I'm assuming a user could take this and either use it via the CLI or share it with someone else to do the same? A few additional comments:
> 
> > If the use case is to copy/paste to a YAML file, should we alternatively just allow the user to "download the YAML" or at the very least allow for an easy Copy function?
> 
> I think the most common use case for a user would be to look at the JSON/YAML to undestand what was actually created and not passing it somewhere else.
> 
> * There was a concern that the generated JSON could look like an error message from afar. That is why I asked about having it collapsed by default.
> * We also don't want the user to think that he has to copy it somewhere, because it was already created.
> 
> > If we keep the code snippet, we should probably label this box. Should this really be YAML rather than JSON for future use when creating VMs via the UI?
> 
> We could support YAML in a followup
> 
> > Minor typo on "successful", drop the last "l".
> > I don't think the "Cancel" or "Back" actions make sense to keep at this point. Any way we can remove them for this step?
> 
> Thanks, I will look at it.

typo is already fixed by https://github.com/kubevirt/web-ui-components/pull/41

src/components/Form/Integer.js
suomiy: I agree. I will start looking into the table first then.

suomiy:
added a possibility to see and remove disks from the user template. I dissabled editing for these disks (because they are part of the template), the only thing which can be changed is boot order

Can you please give me a final round of review, so we can merge it and also release new version of https://github.com/kubevirt/web-ui

@mareklibra @rawagner @vojtechszocs   

@lizsurette I am not sure how to highlight rows from user template (which are not editable)
I don't know if displaying the name in a `disk (Bootable) (From Template)` would be the right way?

rawagner:
@suomiy 
this flow is not ok IMHO:
1. provision source Template
2. go to storage tab
3. storage is correctly read from chosen template
4. go back and change provision source to something else (ie PXE)
5. go to storage tab
6 storage read from template is still there

rawagner:
In case of provision source URL, i think we should require that at least one storage will be defined. URL provision source uses DataVolumes and the datavolume spec in VM has PVC spec hardcoded see https://github.com/kubevirt/web-ui-components/blob/master/src/k8s/request.js#L139 . We should require that this disk/PVC will be defined in Storage Tab and remove the hardcoded stuff.

src/components/Wizard/CreateVmWizard/StorageTab.js
```diff
@@ -0,0 +1,404 @@
...
+              actionType: DELETE_ACTION,
+              text: 'Remove Disk'
+            }
+          ],
+          visibleOnEdit: false
+        })
+      }
+    ];
+
+    // TODO: add create disk button once the create blank disk feature is implemented in kubevirt
```
rawagner: shouldnt we allow to create the disk ? Kubevirt already supports blank disks in latest release

src/components/Wizard/CreateVmWizard/StorageTab.js
```diff
@@ -0,0 +1,404 @@
...
+  }
+}
+
+StorageTab.defaultProps = {
+  initialDisks: []
+};
+
+StorageTab.propTypes = {
+  storageClasses: PropTypes.array.isRequired,
+  storages: PropTypes.array.isRequired,
```
rawagner: can you rename it to pvcs/volumeclaims/persistentvolumeclaims so its clearer what it actually represents ? Applyes also for other places.

src/components/Wizard/CreateVmWizard/StorageTab.js
```diff
@@ -0,0 +1,404 @@
...
+    ];
+
+    // TODO: add create disk button once the create blank disk feature is implemented in kubevirt
+    /*
+     * <Button className="create-disk" onClick={this.createDisk} disabled={this.state.editing} id="create-disk-btn">
+     *   Create Disk
+     * </Button>
+    */
+    return (
+      <React.Fragment>
```
rawagner: is the TableFactory missing something ? Maybe you could use it and add missing stuff/improve it ?

src/components/Wizard/CreateVmWizard/StorageTab.js
suomiy: fixed

suomiy:
all fixed

src/components/Wizard/CreateVmWizard/StorageTab.js
suomiy: It still does not seem to work. I would leave for the followup

src/components/Wizard/CreateVmWizard/StorageTab.js
suomiy: I think that much abstraction is not necessary here.

suomiy:
> In case of provision source URL, i think we should require that at least one storage will be defined. URL provision source uses DataVolumes and the datavolume spec in VM has PVC spec hardcoded see https://github.com/kubevirt/web-ui-components/blob/master/src/k8s/request.js#L139 . We should require that this disk/PVC will be defined in Storage Tab and remove the hardcoded stuff.

That would be nice, but we would have to distinguish that particular disk in the ui - to somehow show that the disk has a URL source. @lizsurette  ?

anyway I would leave this functionality to the followup

lizsurette:
> > In case of provision source URL, i think we should require that at least one storage will be defined. URL provision source uses DataVolumes and the datavolume spec in VM has PVC spec hardcoded see https://github.com/kubevirt/web-ui-components/blob/master/src/k8s/request.js#L139 . We should require that this disk/PVC will be defined in Storage Tab and remove the hardcoded stuff.
> 
> That would be nice, but we would have to distinguish that particular disk in the ui - to somehow show that the disk has a URL source. @lizsurette ?

+1 to tying the URL provision source together closely with the disks(PVCs) defined in this step. As I'm digging into this stuff a bit more, it seems like there could be two main use cases...please correct me if I'm wrong since I'm very new to it all :) -

1 - The PVC with the provision source (ISO, Image, or URL) already exists and I just want to reference and use it.

2 - The PVC does not exist yet and I want to create it to store the data based on my provision source selection.

Maybe for the case where the PVC already exists we could automatically add it to the Storage table for the user? I'm not sure technically how difficult this would be to do :) The user could always remove it and add/create a different disk if they'd like. If the PVC doesn't exist yet and in the case where they remove a PVC that might have already been there, I think it would be nice to do the same error checking we do with PXE / Network step to be sure at least one PVC has been defined before letting the user continue.

rawagner:
URL and Registry imagesources are broken by this change. 
1. choose URL or Registry
2. add disk in storage tab
3. added disk is marked as bootable and has boot order: 1 which means VM wont boot from disk added by URL or Registry

src/k8s/request.js
```diff
@@ -238,10 +287,101 @@ const addCloudInit = (vm, basicSettings) => {
...
+  });
+
+  addBootableDisk(
+    vm,
+    networks,
+    {
+      name,
+      volumeName: name,
+      disk: {
+        bus: VIRTIO_BUS
```
rawagner: disk spec should be taken from template's default disk. See `ANNOTATION_DEFAULT_DISK`. If it exists, than you should use what is defined there (bus and possibly other stuff) and change only name, volumename and boot order. If the annotation is not there (for example in case of template from import-vm-apb) we could take the first found disk. If no disk is found than we could fallback to this.

rawagner: This can be done as followup

mareklibra:
Merged within #10

PR #22, opened 2018-09-14
Adapt project to be an npm package

mareklibra:
The project will hold just reusable components published as npm packages.

With this patch, the standalone application is removed from the project.
Storybook or test environment remains working.


mareklibra:
Published npm package: https://www.npmjs.com/package/kubevirt-web-ui-components

mareklibra:
NPM `dependencies` moved under `devDependencies` and `peerDependencies`, some of them are removed completely. 
Please take a closer look, if it makes sense this way.

mareklibra:
Prior merging this, I would like to hear opinion of others as well.

package.json
```diff
@@ -62,9 +73,14 @@
...
+  "peerDependencies": {
+    "patternfly-react": "^2.14.0",
+    "react": "^16.4.2",
+    "react-dom": "^16.4.2"
+  },
+  "dependencies": {},
   "scripts": {
-    "start": "node scripts/start.js",
-    "build": "node scripts/build.js",
+    "dist": "rm -rf dist; yarn lint && CI=nowatch yarn test && node scripts/dist.js",
```
ohadlevy: travis already does linting, maybe we should remove it as its included in this task?

mareklibra: The point is, this target is called right before publishing the npm package. So final check to be sure ...

ohadlevy: that's fine, I am saying we could remove it in the travis configuration as it does it twice.

package.json
```diff
@@ -62,9 +73,14 @@
     "webpack-manifest-plugin": "1.3.2",
     "whatwg-fetch": "2.0.3"
   },
+  "peerDependencies": {
+    "patternfly-react": "^2.14.0",
+    "react": "^16.4.2",
+    "react-dom": "^16.4.2"
+  },
+  "dependencies": {},
   "scripts": {
```
mareklibra: We can add `validations`, as @vojtechszocs proposed in a meeting.
Plus more explicit semver format for dependencies and update manually.

ohadlevy: :+1: but please do it on another PR that is specific for validations.

src/components/__tests__/HelloWorld.test.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
```
mareklibra: I just followed former structure of this project, but I'm for co-locating of tests with components

package.json
vojtechszocs: Long-term, the whole project should be build-able (lint, validate, test, actual build) via single script, using `pre`-scripts to implement the logical chain of build steps.

> I am saying we could remove it in the travis configuration as it does it twice

+1 on removing `yarn lint` from `.travis.yml`, however the main point here is project-specific build steps leaking into Travis config. In other words, in the long-term, `.travis.yml` should only contain `yarn build` plus any CI-specific stuff like Coveralls publish.

package.json
vojtechszocs: Thanks for bringing this up. To summarize, validations are meant to be small programs (exporting a function that returns `true` or `false`) to check cross-cutting concerns of the project structure, like "each React component has associated test" and "we use a consistent dependency semver format".

src/components/__tests__/HelloWorld.test.js
vojtechszocs: Right, this is the trend I've been observing in many React projects.

Co-locating `.test.js` and `.stories.js` and other React component "friends" makes it easier to navigate the project structure, improving developer productivity. We should also make it a rule that every React component has an associated unit test, this can be done through validations.

.gitignore
```diff
@@ -8,6 +8,7 @@
 
 # production
 /build
```
vojtechszocs: I think we should remove `/build` since it's where webpack-generated stuff goes, based on Create React App configs.

vojtechszocs:
LGTM, just one small comment.

FYI, Jest will soon support [custom snapshot paths](https://github.com/facebook/jest/pull/6143) so we can co-locate `.snap` files just like `.test.js` and others, avoiding hardcoded `__whatever__` assumptions imposed by the test framework.

.gitignore
vojtechszocs: Actually I don't think it's that big of a deal, we can change this later, resolving.

vojtechszocs:
I think we're good to go with this one.

rawagner:
LGTM

mareklibra:
To highlight identified follow-ups:
- add `validators`
- tighten-up semver in package.json
- co-locate tests and components 
- clean-up `.gitignore`
- clean-up `.travis.yml`

pcbailey:
Sorry I was late to the party. Luckily, I had nothing new to add and am good with the changes made here. =)

src/components/__tests__/HelloWorld.test.js
pcbailey: I'm glad we're doing this. I've seen it in other projects and really like it as opposed to maintaining a completely separate directory for tests.

PR #25, opened 2018-09-18
Babelize `dist` result

mareklibra:
By distributing babelized output instead of plain component sources,
we can use more JavaScript features of our choice.


mareklibra:
Fixed: copy .cs files to dist

mareklibra:
To verify, I bumped version of the project. So far, there's no harm with that since there are no real users of the project.

vojtechszocs:
@mareklibra I'm actually working on this too :smile: also I think we're missing some Babel config here.

Overall, our target code should be ES6 so there shouldn't be much difference between `src` and `dist` code.

vojtechszocs:
Also, components should be self-contained, i.e. importing the given component should also import the relevant bit of CSS.

mareklibra:
| I'm actually working on this too

yay :-)

| we're missing some Babel config here.

please feel free to add, the rusult seems to be good enough so far

| target code should be ES6

as long it is consumable with settings of kubevirt/web-ui which should stay untouched

| Also, components should be self-contained, i.e. importing the given component should also import the relevant bit of CSS.

Can we do this without bringing additional requirements on consuming webpack setting? If not, it would cause more troubles than benefits.

mareklibra:
Regarding the styling, I this we should use SASS.
by providing one "top-level" scss file importing all the bits down the structure, we can either
- copy `.scss` as it is from `src` to `dist`
- compile to single `.css`

In both cases, the consuming application will reference (import) just a single file.
The kubevirt/web-ui is already prepared for sass, so I like the first option.
If we want be even more generic, the later option is better.

vojtechszocs:
> Regarding the styling, I this we should use SASS.

Yes, but we shouldn't process it into plain CSS during `dist` build.

Our project is meant to be consumed on source level, i.e. component sources and any other assets (such as `.scss` files) are to be imported by the consuming application, which is responsible for setting up its own build process (webpack, Typescript, etc) that knows how to handle ES6, Sass, etc.

mareklibra:
|| Regarding the styling, I this we should use SASS.

| Yes, but we shouldn't process it into plain CSS during dist build.

Then the first option is ok.

vojtechszocs:
In other words, optimizing our JS & CSS & other assets will make it harder for the consuming application to do the "final" optimization, which brings our stuff and their stuff together in a single (webpack) compilation.

> by providing one "top-level" scss file importing all the bits down the structure
> In both cases, the consuming application will reference (import) just a single file.

This is what `patternfly-react` does, we can do that too.

mareklibra:
Closed in favor of #26

PR #26, opened 2018-09-18
Clean up & simplify the project

vojtechszocs:
# General

- use consistent semver range format for all dependencies: `A.x` or `A.B.x` or `A.B.C`
- use [`files`](https://docs.npmjs.com/files/package.json#files) to whitelist assets to be published as part of the npm package, remove `.npmignore` file because [whitelist is safer than blacklist](https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d)
- add dev dependency for each peer dependency to ensure that development environment has all the necessary modules installed in the `node_modules` directory (see related [Yarn discussion](https://github.com/yarnpkg/yarn/issues/5287))
- rework `scripts` to better reflect the general build & development workflow
    - project build phases: `lint`, `validate`, `test`, `build` (executed in order)
    - this means that `yarn build` is all that's necessary to build the whole project
  - add `node: >=8.6.x` to [`engines`](https://docs.npmjs.com/files/package.json#engines) (8.6+ because of [object rest/spread syntax](https://node.green/#ES2018-features-object-rest-spread-properties) support)
- update `.travis.yml` to [use the latest Node.js LTS version](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions) (`lts/*`)

# Project structure

- `/config`
  - contains all configuration (try to keep the project root clean)
- `/coverage`
  - contains coverage information (`lcov` format, ready for use with [Coveralls](https://github.com/nickmerwin/node-coveralls))
- `/dist`
  - contains build artifacts, everything here will be published as part of the npm package
- `/src`
  - contains React components & related frontend code
  - [co-locate tests](http://islovely.co/posts/colocating-unit-tests/) and other component "friends" like snapshots
- `/tools`
  - contains support tools

# Scripts

- to build the project, run `yarn build` (includes lint, validate and test phases)
- to run Jest in watch mode, run `yarn test-watch`

# Validations

- every module under `tools/validations` exports an `async` function
- example of a dummy validation that always succeeds:
```js
module.exports = async function () { return true }
```
- print *useful* and *actionable* message to developers when the validation fails, for example:

val1

- don't stop on the first problem, but collect and report all results in one go, for example:

val2

- don't print anything upon validation success, for example:

val3

# Linting

- use two separate ESLint configurations:
  - `eslint.browser.js` applies to `src`
  - `eslint.node.js` applies to `config` and `tools`
- use [babel-eslint](https://github.com/babel/babel-eslint) parser to support more recent language features (e.g. [class fields](https://github.com/tc39/proposal-class-fields))

# Testing

- use two separate Jest configurations (using the [projects](https://jestjs.io/docs/en/configuration.html#projects-array-string-projectconfig) feature):
  - `jsdom-env` for tests in `src`
  - `node-env` for tests in `tools`
- collect coverage information only for `src` directory (make testing `tools` possible while keeping our focus on `src` coverage)

# Babel

- use the latest stable 7.x version
- use `babel-core: 7.0.0-bridge.0` to bridge with tools that rely on older Babel version ([details](https://github.com/babel/babel-bridge))

# Component playground

- *WORK IN PROGRESS*
- evaluate [react-cosmos](https://github.com/react-cosmos/react-cosmos) which is more focused on the development workflow
  - ability to modify component props, state or context
  - ability to mock dependencies if necessary (`fetch` API, etc.)
  - developer UX should be close to running Jest in watch mode

# Notes on Storybook

- relies on older version of webpack ([version 2](https://storybook.js.org/configurations/custom-webpack-config/)) and Babel (version 6)
- tweaking Storybook-specific webpack config is error-prone, facing errors when trying to override the default `babel-loader` entry (both Jest and Storybook should use the same Babel config)
- lots of addons that should really be part of the core itself, which means lots of extra dependencies (and peer dependency warnings)


vojtechszocs:
Still facing some peer dependency warnings:
```
warning "patternfly-react > reactabular-table@8.14.0" has unmet peer dependency "lodash@>= 4.0.0 < 5.0.0".
warning "patternfly-react > sortabular@1.6.0" has unmet peer dependency "lodash@>= 4.0.0 < 5.0.0".
warning "patternfly-react > table-resolver@3.3.0" has unmet peer dependency "lodash@>= 3.0.0 < 5.0.0".
warning "patternfly-react > table-resolver@3.3.0" has unmet peer dependency "redux@>= 3.0.0 < 4.0.0".
```

@priley86 I guess those are unavoidable due to Yarn installing all optional dependencies by default, right?

vojtechszocs:
There are currently some weird trace messages during the build:
```
Trace: The node type SpreadProperty has been renamed to SpreadElement
```
It's due to /me using older `babel-plugin-*` versions, I'm going to fix that.

priley86:
> Still facing some peer dependency warnings:
> 
> ```
> warning "patternfly-react > reactabular-table@8.14.0" has unmet peer dependency "lodash@>= 4.0.0 < 5.0.0".
> warning "patternfly-react > sortabular@1.6.0" has unmet peer dependency "lodash@>= 4.0.0 < 5.0.0".
> warning "patternfly-react > table-resolver@3.3.0" has unmet peer dependency "lodash@>= 3.0.0 < 5.0.0".
> warning "patternfly-react > table-resolver@3.3.0" has unmet peer dependency "redux@>= 3.0.0 < 4.0.0".
> ```
> @priley86 I guess those are unavoidable due to Yarn installing all optional dependencies by default, right?

hey @vojtechszocs - i believe these are some upstream issues with the version of reactabular. I can try upgrading PF React to latest to see if it resolves. If not, I can follow up in the upstream w/ @bebraw and maybe make a contribution to reactabular to resolve it. Apologies, this has been outstanding for a bit but I am happy to take a look.

package.json
```diff
@@ -8,136 +8,71 @@
...
-    "webpack": "3.8.1",
-    "webpack-dev-server": "2.9.4",
-    "webpack-manifest-plugin": "1.3.2",
-    "whatwg-fetch": "2.0.3"
+  "files": [
+    "/dist"
+  ],
+  "scripts": {
+    "lint": "yarn lint-js && yarn lint-css",
+    "lint-js": "yarn lint-js-browser && yarn lint-js-node",
```
mareklibra: I find `lint:js` more common than `lint-js` (colon vs. dash)

package.json
```diff
@@ -8,136 +8,71 @@
...
+    "lint-css": "stylelint 'src/**/*.css' --config config/stylelint.config.js",
+    "prevalidate": "yarn lint",
+    "validate": "node tools/runValidations.js",
+    "pretest": "yarn validate",
+    "test": "rimraf coverage && jest --config config/jest.config.js",
+    "test-watch": "yarn test --watch",
+    "prebuild": "yarn test --ci",
+    "build": "rimraf dist && yarn build-js && yarn build-css",
+    "build-js": "babel src --config-file ./config/babel.config.js --out-dir dist --ignore '**/*.test.js,**/setupTest.js'",
+    "build-css": "cd src && find . -name '*.css' -exec cp {} ../dist/{} \\;",
```
mareklibra: How are `.css` imported by consuming application?

Considering project context, I think we should not import styling from components.
Use of sass and hierarchical imports with single entry-point seems to be better approach.

config/eslint.node.js
```diff
@@ -0,0 +1,29 @@
+'use strict'
```
mareklibra: We should use semi-colons `;` to stay aligned with other related projects.

src/components/HelloWorld/HelloWorld.test.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
```
mareklibra: I like this co-location of `test.js` files.

package.json
```diff
@@ -8,136 +8,71 @@
...
+    "/dist"
+  ],
+  "scripts": {
+    "lint": "yarn lint-js && yarn lint-css",
+    "lint-js": "yarn lint-js-browser && yarn lint-js-node",
+    "lint-js-browser": "eslint src --config config/eslint.browser.js",
+    "lint-js-node": "eslint config tools --config config/eslint.node.js",
+    "lint-css": "stylelint 'src/**/*.css' --config config/stylelint.config.js",
+    "prevalidate": "yarn lint",
+    "validate": "node tools/runValidations.js",
```
mareklibra: nice

package.json
```diff
@@ -8,136 +8,71 @@
...
-    "stylelint": "^8.4.0",
-    "stylelint-config-standard": "^18.2.0",
-    "sw-precache-webpack-plugin": "0.11.4",
-    "url-loader": "0.6.2",
-    "webpack": "3.8.1",
-    "webpack-dev-server": "2.9.4",
-    "webpack-manifest-plugin": "1.3.2",
-    "whatwg-fetch": "2.0.3"
+  "files": [
+    "/dist"
```
mareklibra: I like this approach more.

config/eslint.node.js
pcbailey: +1

vojtechszocs:
> There are currently some weird trace messages during the build

It's fixed now.

ESLint uses `babel-eslint` parser for parity with Babel-supported syntax. Babel uses the minimal set of plugins for successful code compilation (currently only [`@babel/plugin-proposal-class-properties`](https://babeljs.io/docs/en/next/babel-plugin-proposal-class-properties.html)).

Recent [Travis build log](https://travis-ci.com/kubevirt/web-ui-components/builds/85302206) is clean, without any errors or warnings.

vojtechszocs:
> hey @vojtechszocs - i believe these are some upstream issues with the version of reactabular. I can try upgrading PF React to latest to see if it resolves. If not, I can follow up in the upstream w/ @bebraw and maybe make a contribution to reactabular to resolve it. Apologies, this has been outstanding for a bit but I am happy to take a look.

Thank you @priley86 and @bebraw for looking into this. It's not a big issue - I just stumbled upon these peer dependency warnings when updating our project's dependencies.

Using `patternfly-react@2.17.3` and looking at our `node_modules` directory, `reactabular-table` has peer dependency on both `react` and `lodash`. If I understand correctly, `patternfly-react` uses `reactabular-table` to implement table-like components, so peer dependency propagation seems to be justified here.

It's just that `lodash` in particular is a utility library - taking `eslint-plugin-import` as an example, it has peer dependency on `eslint` and normal dependency on `lodash`. This is what `reactabular-table` should do too, in my opinion.

As for the other two peer dependency warning sources, `sortabular` and `table-resolver`, these are declared by `patternfly-react` as optional dependencies. While you can do `yarn install --ignore-optional` this impacts *all* optional dependencies within your dependency tree. I'd really like to see Yarn allowing us to ignore *specific* optional dependencies (see [this issue](https://github.com/yarnpkg/yarn/issues/5251)), but in the meantime, we'll have to live with `sortabular` and `table-resolver` pulled through `patternfly-react`.

package.json
vojtechszocs: Does this mean you also prefer e.g. `lint:js:browser` over `lint-js-browser`?

package.json
vojtechszocs: > Use of sass and hierarchical imports with single entry-point seems to be better approach.

This is what `patternfly-react` does too, I'm OK with that. We can have a single (s)css entry point to be imported by the consuming application, in order to pull in all the necessary styling.

In our components, we can simply refer to CSS class names via [`className` prop](https://reactjs.org/docs/faq-styling.html).

Adding Sass to the project is on my task list.

config/eslint.node.js
vojtechszocs: Our ESLint configuration is currently based on [StandardJS](https://standardjs.com/) which eliminates the need for semicolons. Technically, semicolons in JavaScript are unnecessary [except a few edge cases](https://standardjs.com/rules.html#semicolons) due to the ASI (automatic semicolon insertion) mechanism being part of ECMAScript language spec ([link](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-rules-of-automatic-semicolon-insertion)).

We already have a good experience using StandardJS as the code style baseline in oVirt projects like [ovirt-web-ui](https://github.com/oVirt/ovirt-web-ui) and [ovirt-engine-ui-extensions](https://gerrit.ovirt.org/gitweb?p=ovirt-engine-ui-extensions.git;a=tree).

When it comes to contributing our components to other projects (like `patternfly-react`), we can use their [Prettier](https://prettier.io/) config associated with the given project. Typically, this involves copying the component into such project and using Prettier-aware editor to save & automatically format the code.

To summarize, I don't think that not using semicolons creates that big of a difference between us and other projects like `patternfly-react`. Also, check out [`eslint-plugin-patternfly-react` recommended config](https://github.com/patternfly/patternfly-react/blob/master/packages/eslint-plugin-patternfly-react/lib/config/recommended.js) - it's actually based on StandardJS, but also applies AirBnb on top of it (this one re-adds the semicolons).

vojtechszocs:
> resolve usage of `.css`. IMO, sass is better plus consuming application should handle import and compilation.

Right. `kubevirt-web-ui-components` is a library, producing optimized JS/CSS outputs doesn't make much sense - it's the job of the consuming application to bring all JS/CSS/whatever sources together and produce an optimized bundle for execution in the browser.

I'm all for using Sass, where we import `patternfly-react` styling (which in turn imports `patternfly` styling), then add our own project-specific styling, and produce either `.scss` or `.css` entry file to be imported by the consuming application. If the target application already uses Sass, we can omit CSS compilation and simply use the Sass sources.

As for Babel, we should only apply the basic transformations that produce ES6 compatible code. Again, the target application already supports ES6 syntax and our components are to be imported as source code, becoming part of the application's module tree/bundle.

config/eslint.node.js
priley86: +1 for consistent lint configs. I'm not sure what warrants the new config here, but would like to stay consistent w/ patternfly and [console](https://github.com/openshift/console) where possible. My guess is we will eventually introduce a tslint config in pf4 react, but for now, eslint-plugin-patternfly-react is the closest config in PF if you'd like to share any configs introduced there. Not overly pressing, but consistency is nice as we share components across projects.

tools/common.js
```diff
@@ -0,0 +1,6 @@
+'use strict'
+
+module.exports = {
+  checkMark: '✓',
```
ohadlevy: should we use non utf8 chars? e.g. 
```html
&#10003;
```

package.json
mareklibra: If we really need such deep hierarchy and so many targets, than yes. *But this is nitpicking.* Just pointing to what is more common elsewhere.

My other preference is to keep the number of targets (scripts) at reasonably low number - more or less just those which are really called manually or as a "code sharing" if content of targets overlaps.
I find it nice if I can understand what to call from `scripts` at the first glance.

config/eslint.node.js
mareklibra: Having independent project, I would not like to follow all of the pf-react linting rules.
But since this is not our case, switching of context in our heads will cause much more stress than getting used to it.

config/paths.js
```diff
@@ -1,55 +1,19 @@
...
-  appNodeModules: resolveApp('node_modules'),
-  publicUrl: getPublicUrl(resolveApp('package.json')),
-  servedPath: getServedPath(resolveApp('package.json')),
-};
+  projectRoot,
+  packageJson: pathTo('package.json'),
+  // nodeModules: pathTo('node_modules'),
+  src: pathTo('src'),
+  components: pathTo('src/components'),
+  // storybook: pathTo('storybook'),
```
pcbailey: Why the commented out code here?

package.json
vojtechszocs: > If we really need such deep hierarchy and so many targets, than yes.

OK, I'll go with `lint:js:browser` format. [patternfly-react](https://github.com/patternfly/patternfly-react/blob/master/package.json) project uses it too.

> My other preference is to keep the number of targets (scripts) at reasonably low number - more or less just those which are really called manually or as a "code sharing" if content of targets overlaps.

Agreed.

vojtechszocs:
> only big concern is that over time we will digress from ui components developed in console and ui components developed in patternfly-react. Hopefully you have a strategy in mind to keep these in sync as we go forward. It seems that patternfly-react is the lowest common denominator at the moment.

Right now, all of these:
- `openshift/console` / `kubevirt/web-ui` (*)
- `kubevirt/web-ui-components`
- `patternfly-react`
use ESLint.

(*) this one doesn't seem to use neither StandardJS and/or AirBnb base config, just `eslint:recommended`.

It really seems that `patternfly-react` is the lowest common denominator and we should stay close to its code style. Once `patternfly-react` starts using TSLint, that can be seen as another reason to simply adopt TypeScript in `kubevirt/web-ui-components`.

> do you see typescript as a future possibility in this repo or in kubevirt-web-ui ?

My understanding is that `@patternfly/react-core` (based on next-gen PatternFly v4) is still under development, so we're currently using `patternfly-react` (based on PatternFly v3).

Once we start consuming `@patternfly/react-core` components, and assuming that project already converged to TypeScript & TSLint, we can adopt TypeScript and reuse that project's TSLint config. Alternatively, we can adopt TypeScript right now, just for the sake of type checking - I'm not against it. Also note that `openshift/console` already uses TypeScript (and webpack) but with ESLint.

vojtechszocs:
> I would like to keep pf-react linting rules, if needed fix the lint rules there.

I will modify our ESLint config to be as close to `patternfly-react` as possible.

> I would like to keep prettier, I think we are going to save a lot of time this way.

I will re-add Prettier, both in terms of config file and in terms of making ESLint config aware of it.

> I would also suggest that a lot of the common infra here, can be shared, and useful to many other projects, I wonder if we can make it an npm package and maybe include it in pf-react (e.g. validators) /cc @priley86 ?

I can contribute it as a separate package to `patternfly-react` monorepo, but that's out of scope for this PR.

tools/common.js
vojtechszocs: Is it safe to assume that our sources are encoded in UTF8 charset? If so, why should we avoid using UTF8 chars?

config/paths.js
vojtechszocs: Thanks for noticing, I will remove these commented-out lines eventually.

vojtechszocs:
@ohadlevy @mareklibra  @jelkosz - we had a good discussion today about open issues in this PR, here's my summary:

- ESLint config will be equal to `eslint-plugin-patternfly-react` minus the super opinionated stuff (`rules` overrides) and minus `node` env (which is unrelated to client-side code), this also means that we will
  - keep using Prettier
  - keep using semicolons
  - follow-up: break `patternfly-react/recommended` into smaller bits so we can just use the "minimal" config, for example
```js
{
  plugins: ['plugin:patternfly-react/minimal'],
  rules: [] // tweak stuff if necessary
}
```

- regarding test co-location, our component tree will look like
```
/components/FancyWizard/tests/FancyWizard.test.js
/components/FancyWizard/tests/FancyWizardBody.notest
/components/FancyWizard/FancyWizard.js
/components/FancyWizard/FancyWizardBody.js
/components/FancyWizard/index.js
```

where `FancyWizard` is the main logical component.

- we will keep existing validations to ensure consistency and correctness of `package.json` and make the presence/absence of unit tests explicit

tools/validations/dependencyPeerAsDev.js
```diff
@@ -0,0 +1,32 @@
+// Go through peerDependencies declared in package.json file and check that corresponding
+// entries exist in devDependencies. This is to ensure that development environment has all
+// the necessary modules installed in the node_modules directory.
+
+const chalk = require('chalk');
```
priley86: I think this validation can also help in PF React. It would be nice to ensure `patternfly-react` and `patternfly` and `patternfly-next` dependencies are all in sync on the monorepo. I will create a follow up issue on this.

priley86:
updates look good to me @vojtechszocs - after Travis tests pass I am happy to see it merge if others are.

vojtechszocs:
Thanks @priley86 for your review. As for Yarn & Travis, I'll update `.travis.yml` to use a more recent Yarn version (1.9.x) according to [Travis docs](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#using-a-specific-yarn-version).

I'll also post a brief summary of recent changes to get everyone on the same page before we merge this PR.

vojtechszocs:
Travis build is fixed. Coverage decreased from 100% to ~31% due to `tools/validations` not completely covered by tests. (This is OK for now, we can improve that later on.)

vojtechszocs:
I see the following in Travis log:

```
PASS   jsdom-env  src/components/HelloWorld/tests/HelloWorldMessage.test.js
  ● Console
    console.log node_modules/react-cosmos-config/dist/log.js:17
      [Cosmos] No config file found, using defaults
```
and I'd like to fix that before we merge this PR. Strangely enough, above doesn't occur on local devel. env.

vojtechszocs:
[Recent Travis build](https://travis-ci.com/kubevirt/web-ui-components/builds/85766100) is clean, no more `[Cosmos] No config file found, using defaults` entries :tada:

pcbailey:
The updates look good to me, as well. Great job, Vojtech!

vojtechszocs:
Hey everyone, here's a brief summary of the recent changes:

- adapt to ESLint configuration declared in [`patternfly-react/recommended`](https://github.com/patternfly/patternfly-react/blob/master/packages/eslint-plugin-patternfly-react/lib/config/recommended.js)
  - PF-React specific rule configs extracted into `eslint.rules.*.js` files
- add [`.prettierrc`](https://prettier.io/docs/en/configuration.html) to support Prettier-aware editors
- add [`.browserslistrc`](https://github.com/browserslist/browserslist#queries) for use with [`@babel/preset-env`](https://babeljs.io/docs/en/babel-preset-env#browserslist-integration) as well as other tools
- use [Cosmos](https://github.com/react-cosmos/react-cosmos#getting-started) as our UI component playground technology
  - run `yarn cosmos` to start the Cosmos web server
  - every component should have a corresponding `fixtures/<ComponentName>.fixture.js`
  - Cosmos effectively *emulates* the consuming application, rendering each component in isolation
- current Babel configuration is minimal on purpose
  - produce ES6 compatible output
  - optimization & module bundling is the responsibility of the consuming application
- collect coverage for `src/components/**/*.js` and `tools/validations/*.js`
- use Sass for styling our components
  - `sass/_components.scss` should be imported to apply KubeVirt UI component specific styling
  - `sass/_dependencies.scss` is used to pull in dependant styling (Bootstrap, PatternFly core)
  - currently, there's no `.scss => .css` compilation step, just copy `sass` into `dist` as-is
- Travis CI: install the latest version of Yarn via [`before_install`](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#using-a-specific-yarn-version) phase
- miscellaneous
  - use [`shx`](https://github.com/shelljs/shx) for better script portability
  - some `devDependencies` are added due to using Cosmos: `@babel/polyfill`, `babel-loader`, `html-webpack-plugin`, `webpack`

Current `dist` content:
```
$ tree dist
dist
|-- js
|   |-- components
|   |   `-- HelloWorld
|   |       |-- HelloWorld.js
|   |       |-- HelloWorldMessage.js
|   |       `-- index.js
|   `-- index.js
`-- sass
    |-- components
    |   `-- _HelloWorld.scss
    |-- _components.scss
    `-- _dependencies.scss

5 directories, 7 files
```

Follow-up action items:
- (gap) update Cosmos configuration to load our `.scss` files
- (gap) when PR gets merged into `master` branch, deploy static Cosmos application
- (enhancement) modularize `eslint-plugin-patternfly-react` and use it in this project
- (enhancement) add `docs` directory containing useful information for developers

PR #32, opened 2018-09-27
Update templates to support custom flavor

rawagner:


src/k8s/mock_templates/fedora28.template.js
```diff
@@ -41,11 +41,13 @@ export const fedora28 = {
...
               cpu: {
-                cores: 2
+                // eslint-disable-next-line no-template-curly-in-string
+                cores: '${{CPU_CORES}}'
               },
               resources: {
                 requests: {
-                  memory: '2G'
+                  // eslint-disable-next-line no-template-curly-in-string
+                  memory: '${MEMORY}G'
```
mareklibra: Are single curly brackets ok here?
Can we include unit (`M`, `G`, ...) in the parameter?

rawagner: CPU_CORES is using double brackets to express that value is integer
MEMORY is string, so single brackets are used.

I moved the M,G to parameter value.

And added test

rawagner:
@mareklibra can you please take a look again ?

PR #33, opened 2018-10-01
Deploy Cosmos to gh-pages branch upon pushing to master branch

vojtechszocs:
## Goal

When pushing changes to `master` branch (typically upon merging a pull request), build and deploy the static Cosmos application to `gh-pages` branch. Once deployed, the Cosmos application is available at https://kubevirt.github.io/web-ui-components.

## Steps taken

1) Generate deploy key (represented as public/private rsa key pair):
```
$ ssh-keygen -t rsa -b 4096 -C "kubevirt-builder@redhat.com"
```

2) In project settings / deploy keys section, add new key (public one) named `kubevirt-builder` with repo write access.

3) Use [Travis CLI](https://github.com/travis-ci/travis.rb) to encrypt actual deploy key data:
```
$ travis encrypt-file deploy_key -r kubevirt/web-ui-components --org
...
encrypting deploy_key for kubevirt/web-ui-components
storing result as deploy_key.enc
storing secure env variables for decryption
```

4) Add `deploy_key.enc` to git repo, remove public/private keys since they're not needed anymore.
```
$ git add deploy_key.enc
$ rm deploy_key deploy_key.pub
```

5) Use Travis `before_script` build step to decrypt `deploy_key.enc` as `deploy_key` and add it to `ssh-agent`.

6) Use Travis `after_success` build step to export Cosmos application and deploy it to `gh-pages` branch, based on PF-React [`release.sh`](https://github.com/patternfly/patternfly-react/blob/master/.travis/release.sh) script.

## Notes and ideas

- Generated Cosmos application includes a `build-info.txt` file that looks like this:
```
Package version: 0.1.2
Travis build URL: http://foobar
Built on: Tue, 09 Oct 2018 17:09:25 GMT
```
- Try to avoid writing shell scripts: use JavaScript to implement specific CI workflow steps while still being able to execute shell commands.
- Not using `sudo: required` in `.travis.yml` - is it really necessary? :thinking: Note that enabling sudo causes Travis to run the build in a VM (instead of a Docker container), increasing the worker machine boot time.
- The `KubeVirt Builder <kubevirt-builder@redhat.com>` user is made up for the purpose of publishing to `gh-pages` branch.
- Console colors: changed `white` to `blue` on some places. That's because Travis web UI treats "no color" as white within the build log.


vojtechszocs:
@jelkosz I'd like to configure [GitHub pages source](https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/) for this repo, but I can't access project settings:



In our case, GitHub pages source should be set to "gh-pages branch":

vojtechszocs:
@mareklibra `git branch -a` shows a remote branch `mareklibra-patch-1` with latest commit from Sep 13, 2018. Do we still need this branch?

vojtechszocs:
@priley86 @jelkosz I think we'll need to do the following first:

- generate [GitHub personal access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/) - *but which user's account should we use for that?*
- on Travis settings page of `kubevirt/web-ui-components` repo, [add new encrypted variable](https://docs.travis-ci.com/user/environment-variables#Defining-Variables-in-Repository-Settings) containing the GitHub token

From here, we have two options:
1. use [GitHub Pages deployment provider](https://docs.travis-ci.com/user/deployment/pages/) to do the publishing
2. use [`after_success`](https://docs.travis-ci.com/user/customizing-the-build/#the-build-lifecycle) step that calls our script to do the publishing (i.e. use [gh-pages](https://github.com/tschaub/gh-pages) library)

Option 2. is what `patternfly-react` guys are doing at the moment - check out their [release.sh](https://github.com/patternfly/patternfly-react/blob/master/.travis/release.sh) script which deploys Storybook (based on PF-React v3) to `gh-pages` branch.

priley86:
Hey @vojtechszocs - I think using github personal access tokens or deploy tokens is OK for deploying to gh-pages (although deploy tokens is more secure).

We recently moved to surge.sh though so that we could support automatic PR deploys as well (maybe you saw this PR merge recently: https://github.com/patternfly/patternfly-react/pull/653)

I also like Now/Zeit which accomplishes the same thing: https://zeit.co/blog/now-for-github

Netlify does this as well, but we found monorepo support to be poor.

I would say pick whichever option best supports your efforts...

vojtechszocs:
Hey @priley86 that's a great feedback, thanks! I'll make sure to check out all of these options.

priley86:
this looks very nice @vojtechszocs - i have no issues w/ the approach now and nice work on the conversion to JS scripts vs bash (much easier to diagnose for many of us). 👍

vojtechszocs:
Thanks @priley86 I've tried to avoid shell scripts since I'd like us to (re)use JavaScript whenever possible.

We are still executing shell commands (via [exec-sh](https://github.com/tsertkov/exec-sh) library) which are async by nature, so the individual "tasks" at `tools/travis/tasks` are effectively async functions that return `true` or `false` to indicate their success or failure. This is very similar to the concept of validation functions (minus the async part) that implement smaller bits of the bigger workflow.

vojtechszocs:
Note that publishing Cosmos for each open PR will be done as a follow-up, once we have a working deploy to GitHub Pages upon PR merge.

.travis.yml
```diff
@@ -1,11 +1,28 @@
...
   - yarn install
+before_script:
+  - node tools/travis/beforeScript.js
 script:
   - yarn build
   - yarn coveralls
+after_success:
+  - node tools/travis/afterSuccess.js
+branches:
+  only:
```
mareklibra: Can we keep testing on open/force=pushed PRs?
If done in a follow-up, let's have it prepared already before merging this. It would block other PRs otherwise ...

mareklibra: This should work for our recent workflow. I think we can adjust later, if ever needed:

https://stackoverflow.com/questions/31882306/how-to-configure-travis-ci-to-build-pull-requests-merges-to-master-w-o-redunda

vojtechszocs:
Hi @mareklibra

> LGTM but I suggest to block merge until patch enabling tests on PR is available.

To clarify, in `.travis.yml`, the `branches: only: master` was meant to address the issue where Travis CI runs **two** builds whenever a PR gets created/updated, which is confusing and redundant. In this case:

* [Build 117](https://travis-ci.com/kubevirt/web-ui-components/builds/87374270), Commit 003fd22, Started October 9, 2018 19:20:40
* [Build 158](https://travis-ci.org/kubevirt/web-ui-components/builds/439247939), Commit 003fd22, Started October 9, 2018 19:20:44

According to [this answer](https://stackoverflow.com/a/31882307), when using the `branches: only: master` setting:

> it's configured to build all pull requests (regardless of the base branch), but limited to only build pushes to `master`

[An alternative](https://stackoverflow.com/a/41715756) is to do something like `branches: except: <regex>` like in [patternfly-react](https://github.com/patternfly/patternfly-react/blob/master/.travis.yml).

---

TL;DR for now, I'll remove the `branches: only: master` setting, assuming the CI build redundancy is not a big issue for us.

vojtechszocs:
> TL;DR for now, I'll remove the `branches: only: master` setting, assuming the CI build redundancy is not a big issue for us.

OK, I see the PR has been merged, so never mind :smiley: as you wrote, we can adjust later on as necessary.

PR #41, opened 2018-10-10
Add networks tab and PXE boot

rawagner:


src/components/Form/FormFactory.js
```diff
@@ -20,7 +20,15 @@ export const getFormElement = props => {
...
       );
     case 'dropdown':
-      return <Dropdown id={id} value={value || defaultValue} onChange={onChange} onBlur={onBlur} choices={choices} />;
+      return (
+        <Dropdown
+          id={id}
+          value={value || defaultValue}
+          onChange={onChange}
+          onBlur={onBlur}
+          choices={typeof choices === 'function' ? choices() : choices}
```
suomiy: is there a real need for choices to be a function?

src/components/Table/EditableDraggableTable.js
```diff
@@ -132,11 +132,12 @@ class EditableDraggableTable extends React.Component {
   };
 
   getActionButton = (type, additionalData, id) => {
+    const renderConfig = this.getRenderConfig(additionalData);
     switch (type) {
       case DELETE_ACTION:
         return (
           <MenuItem id={id} key={id} onSelect={() => this.crudController.onDelete(additionalData)}>
-            Remove Disk
+            {renderConfig.text}
```
suomiy: please make it more configurable under renderConfig.actions. Then you can also pass it as a parameter

src/components/Table/EditableDraggableTable.js
```diff
@@ -258,7 +259,7 @@ class EditableDraggableTable extends React.Component {
     rowId: rowData.id,
     role: 'row',
     onMove: ({ sourceRowId, targetRowId }) => {
-      if (!this.state.editing) {
+      if (!this.state.editing && targetRowId !== 1 && sourceRowId !== 1) {
```
suomiy: why is this needed? Please resolve it in the user of ON_MOVE callback

suomiy: or please at least make it configurable

src/components/Table/EditableDraggableTable.js
```diff
@@ -146,28 +147,28 @@ class EditableDraggableTable extends React.Component {
...
   getRenderConfig = additionalData => {
-    const { renderConfigs } = additionalData.column;
-    const renderConfigIdx = additionalData.rowData.renderConfig;
-
-    if (renderConfigs && Number.isInteger(renderConfigIdx) && renderConfigIdx < renderConfigs.length) {
-      return renderConfigs[renderConfigIdx];
-    }
-    return null;
+    const { renderConfig } = additionalData.column;
+    return typeof renderConfig === 'function' ? renderConfig(additionalData.rowData) : renderConfig;
```
suomiy: can we also have the original declarative approach with this functional one?

src/components/Table/EditableDraggableTable.js
```diff
@@ -181,7 +182,7 @@ class EditableDraggableTable extends React.Component {
       result = value;
     }
 
-    if (!editable && renderConfig && renderConfig.hasAddendum) {
+    if (!editable && additionalData.column.hasAddendum) {
```
suomiy: why not keep it how it was (dependent on renderConfig)?

src/components/Table/tests/InlineEditRow.test.js
```diff
@@ -7,7 +7,7 @@ const testProp = 'testProp';
 
 const testInlineEditRow = () => <InlineEditRow testProp={testProp} />;
 
-describe('<EditableDraggableTable />', () => {
+describe('<InlineEditRow />', () => {
```
suomiy: thanks :)

src/components/Wizards/CreateVmWizard/CreateVmWizard.js
```diff
@@ -15,6 +18,19 @@ export class CreateVmWizard extends React.Component {
...
+        value: {
+          networks: [
+            {
+              id: 1,
+              name: 'eth0',
+              mac: '',
+              network: POD_NETWORK,
+              edit: false
+            }
+          ]
```
suomiy: please add default value for `valid`

src/components/Table/EditableDraggableTable.js
rawagner: I have column without renderConfig (its not editable) but I want to add addendum to it

src/components/Table/EditableDraggableTable.js
rawagner: sure, lets do this in a follow-up

src/components/Form/FormFactory.js
rawagner: not anymore :)

src/components/Wizards/CreateVmWizard/CreateVmWizard.js
rawagner: added

lizsurette:
@rawagner Any Cosmos link available to see this in action without having a dev environment setup? :)

rawagner:
@lizsurette [storybook](http://pensive-williams-3779f8.netlify.com/?component=Wizards%2FCreateVmWizard%2FT&fixture=namespaces) 
A few observations I had during creating the networks tab:
1. NIC name is not visible anywhere in created VM, so it does not need to be editable, we can just generate the values (Im using eth<number> or maybe i should change it to nic<number> ? ).
2. there is always one Pod network, that is just always present in VM, so I do not allow user to remove it.
3. Pod network cannot be bootable
4. There are no Model, Type and Link State columns as these are not yet configurable

yarn.lock
```diff
@@ -5,6 +5,7 @@
 "@babel/cli@7.x":
   version "7.1.0"
   resolved "https://registry.yarnpkg.com/@babel/cli/-/cli-7.1.0.tgz#a9429fd63911711b0fa93ae50d73beee6c42aef8"
+  integrity sha512-+OdtGZcJNH92CnDqwaPxh7P7gddFyhoiHV3GBzgKpYbxIJlQ4WDEiC8m+AMcueYzlI+bXqrYlIU/Pp17NaC0hg==
```
mareklibra: Please remove yarn.lock from this PR

mareklibra:
Merging to speed things up. Let's don't skip the follow-ups and especially UX review.

rawagner:
@lizsurette 

> @lizsurette [storybook](http://pensive-williams-3779f8.netlify.com/?component=Wizards%2FCreateVmWizard%2FT&fixture=namespaces)
> A few observations I had during creating the networks tab:
> 
> 1. NIC name is not visible anywhere in created VM, so it does not need to be editable, we can just generate the values (Im using `eth<number>` or maybe i should change it to `nic<number>` ? ).

It probably makes sense to generate the name but allow user to change it (he will see it in web UI when editing VM)
> 2. there is always one Pod network, that is just always present in VM, so I do not allow user to remove it.

Not true, I found out how to disable the default pod network. Question is, should be show it in UI by default but allow to remove it ?
> 3. Pod network cannot be bootable
> 4. There are no Model, Type and Link State columns as these are not yet configurable

lizsurette:
@rawagner 

> 1. NIC name is not visible anywhere in created VM, so it does not need to be editable, we can just generate the values (Im using `eth<number>` or maybe i should change it to `nic<number>` ? ).

I'm assuming that the user might want to control these names themselves, right? They will end up seeing this in the details area and be able to view details on them when they access the console, right?

> > 1. there is always one Pod network, that is just always present in VM, so I do not allow user to remove it.
> 
> Not true, I found out how to disable the default pod network. Question is, should be show it in UI by default but allow to remove it ?

Hmm I guess since technically you can disable it, we should allow the user to remove it. But I still like the idea of it being there by default. I think that maps well to what happens today.



> * Pod network cannot be bootable

Makes sense, I think what you are doing today works well where the first one that's created is bootable, and there is an error message, and the user can't continue until they add another. Makes it very clear.



> 2\. There are no Model, Type and Link State columns as these are not yet configurable

No problem! We will keep them in the designs for now and can revisit as more things are available.

Everything else looks great to me!

PR #50, opened 2018-10-11
Use default network model, if defined

rawagner:


src/k8s/request.js
```diff
@@ -161,6 +161,12 @@ const setSourceType = (vm, basicSettings) => {
 };
 
 const setNetworks = (vm, basicSettings, networks) => {
+  const defaultInterface = getDefaultInterface(vm, basicSettings);
+  const interfaceModel = defaultInterface ? defaultInterface.model : undefined;
+
+  delete vm.spec.template.spec.domain.devices.interfaces;
```
suomiy: please don't delete networks and interfaces when using user templates

rawagner: fixed

PR #51, opened 2018-10-15
Add tests for NetworksTab

rawagner:


src/components/Wizards/CreateVmWizard/tests/CreateVmWizard.test.js
```diff
@@ -126,6 +127,23 @@ describe('<CreateVmWizard />', () => {
     expect(component.find(WizardPattern).props().nextStepDisabled).toBeTruthy();
   });
 
+  it('first step renders <BasicSettinsTab />', () => {
```
suomiy: there is a mistake at https://github.com/kubevirt/web-ui-components/pull/51/commits/a449f3935fea5c89e884da1a64c5e0db21091534#diff-1007c24f3f684e98ae9b3e620fd4c85eR43: this isn't merged yet and shouldn't belong to `e09e902 Add networks tab and PXE boot (#41)` 

Please could you try to incorporate these two use cases into the testWalkThrough method if it is feasible?

src/components/Wizards/CreateVmWizard/tests/NetworksTab.test.js
```diff
@@ -1,13 +1,116 @@
...
+  { id: 3, isBootable: false, name: 'nic3', mac: 'mac3', network: 'network3', errors: [] }
+];
 
 describe('<NetworksTab />', () => {
   it('renders correctly', () => {
-    const component = shallow(testNetworksTab());
+    const component = testNetworksTab();
     expect(component).toMatchSnapshot();
   });
+  it('passes existing networks to FormFactory', () => {
```
suomiy: this is implementation detail and IMO shouldn't be tested in unit tests

src/components/Wizards/CreateVmWizard/tests/NetworksTab.test.js
```diff
@@ -1,13 +1,116 @@
...
+    expect(component.instance().checkPxeBootNetwork(rows.slice(0, 1))).toBeFalsy();
+  });
+  it('resolveBootability', () => {
+    const props = { ...NetworksTabFixture.props };
+    props.networks = defaultNetworks;
+    const component = testNetworksTab(props);
+
+    const { rows } = component.state();
+    // change rows order
+    let newRows = component.instance().resolveBootability([rows[0], rows[2], rows[1]]);
```
suomiy: can also be tested outside of the component

src/components/Wizards/CreateVmWizard/tests/NetworksTab.test.js
```diff
@@ -1,13 +1,116 @@
...
+    expect(passedRows[2].isBootable).toBeFalsy();
+    expect(passedRows[2].renderConfig).toEqual(0);
+  });
+  it('checkPxeBootNetwork', () => {
+    const props = { ...NetworksTabFixture.props };
+    props.networks = defaultNetworks;
+    const component = testNetworksTab(props);
+
+    const { rows } = component.state();
+    expect(component.instance().checkPxeBootNetwork(rows)).toBeFalsy();
```
suomiy: I think it is better to put the function outside of the component and test it separately

src/components/Wizards/CreateVmWizard/tests/NetworksTab.test.js
```diff
@@ -1,13 +1,116 @@
...
+    component
+      .find('#create-network-btn')
+      .at(1)
+      .simulate('click');
+
+    expect(component.state().editing).toBeTruthy();
+    expect(component.state().nextId).toEqual(nextId + 1);
+    expect(component.state().rows).toHaveLength(rows.length + 1);
+
+    const newRow = component.state().rows[rows.length];
```
suomiy: I don't think it is necessary to look at the inner state of the component when we can just check the `onChange` callback

src/components/Wizards/CreateVmWizard/tests/NetworksTab.test.js
mareklibra: Agree. Will help readability

src/components/Wizards/CreateVmWizard/tests/NetworksTab.test.js
mareklibra: Agree, it expects particular inner implementation.
Anyway, failing test here might point to a serious issue. How much complicated it would be to test the same "inner flow" as a black-box, so passing props and observing behavior as black-box? If feasible, this approach would bring additional value to us.

PR #52, opened 2018-10-15
Allow disabling VM's pod network

rawagner:


src/k8s/tests/request.test.js
```diff
@@ -300,4 +302,20 @@ describe('request.js', () => {
...
       return vm;
     }));
+  it('no pod network', () =>
+    createVM(k8sCreate, templates, customFlavor, networks).then(vm => {
+      expect(vm.spec.template.spec.domain.devices.autoattachPodInterface).toBeFalsy();
+      expect(vm.spec.template.spec.domain.devices.interfaces).toBeUndefined();
+      expect(vm.spec.template.spec.networks).toBeUndefined();
+      return vm;
+    }));
+  it('only multus network', () =>
```
suomiy: please use more verbose descriptions, it is not obvious what it should test

src/k8s/tests/request.test.js
```diff
@@ -246,6 +246,7 @@ describe('request.js', () => {
     createVM(k8sCreate, templates, vmPXE, pxeNetworks).then(vm => {
       expect(vm.metadata.name).toBe(basicSettings.name.value);
       expect(vm.metadata.namespace).toBe(basicSettings.namespace.value);
+      expect(vm.spec.template.spec.domain.devices.autoattachPodInterface).toBeUndefined();
```
suomiy: Just a thought: shouldn't we set it to true for the sake of verbosity? Even if it defaults to it.

src/k8s/tests/request.test.js
rawagner: is it better now ?

src/k8s/tests/request.test.js
rawagner: Hard to say...I would vote for not defining when not needed.
- We do specify pod network under devices/interfaces so its clear that the VM has pod network
- resulting vm definition is shorter
- docs do not mention this and it might me confusing ?  . 

@mareklibra @jelkosz  any preference ?

src/k8s/tests/request.test.js
suomiy: ok

src/k8s/tests/request.test.js
suomiy: I know docs set it to true by default. It could be just an info for the user what happened, when the generated json is displayed

src/components/Wizards/CreateVmWizard/NetworksTab.js
```diff
@@ -196,10 +198,13 @@ export class NetworksTab extends React.Component {
...
           type: 'dropdown',
-          choices: this.props.networkConfigs.map(networkConfig => networkConfig.metadata.name).filter(networkConfig => {
-            const result = this.state.rows.find(r => r.network === networkConfig);
-            return !result;
-          }),
+          choices: this.props.networkConfigs
+            .map(networkConfig => networkConfig.metadata.name)
+            .concat(POD_NETWORK)
+            .filter(networkConfig => {
+              const result = this.state.rows.find(r => r.network === networkConfig);
```
mareklibra: Can this be one-liner?

rawagner: @mareklibra changed to one-liner and rebased on master

PR #54, opened 2018-10-16
Remove cloudinit disk and volume if cloudinit is not used

rawagner:


src/k8s/request.js
```diff
@@ -238,14 +239,36 @@ const addCloudInit = (vm, basicSettings) => {
...
-  disks.push(diskSpec);
-  devices.disks = disks;
-  domain.devices = devices;
   vm.spec.template.spec.domain = domain;
+  return domain;
+};
+
+const getDevices = vm => {
+  const devices = get(getDomain(vm), 'devices', {});
+  vm.spec.template.spec.domain.devices = devices;
```
suomiy: I think a condition would look better here, so we don't have to re-assign the variable when it's found. Applies for the other function as well.

src/k8s/tests/request.test.js
```diff
@@ -300,4 +300,18 @@ describe('request.js', () => {
...
     }));
+  it('every disk has volume', () =>
+    createVM(k8sCreate, templates, basicSettings, networks).then(vm => {
+      vm.spec.template.spec.domain.devices.disks.forEach(disk => {
+        expect(vm.spec.template.spec.volumes.filter(volume => volume.name === disk.volumeName)).toHaveLength(1);
+      });
+      return vm;
+    }));
+  it('every disk has volume - cloud init', () =>
+    createVM(k8sCreate, templates, basicSettingsCloudInit, networks).then(vm => {
```
suomiy: repeated code, maybe make a function out of this?

rawagner: fixed

src/k8s/request.js
```diff
@@ -195,13 +195,14 @@ const setNetworks = (vm, basicSettings, networks) => {
...
-  get(vm.spec.template.spec.domain.devices, deviceType, []).find(device => device.name === deviceName);
-
 const addCloudInit = (vm, basicSettings) => {
   // remove existing config
-  const volumes = get(vm.spec.template.spec, 'volumes', []);
-  remove(volumes, volume => volume.hasOwnProperty('cloudInitNoCloud'));
+  const volumes = get(getSpec(vm), 'volumes', []);
+  const existingVolume = volumes.find(volume => volume.hasOwnProperty('cloudInitNoCloud'));
+  if (existingVolume) {
+    remove(getDisks(vm), disk => disk.volumeName === existingVolume.name);
```
mareklibra: Maybe I miss something, but is this flow unit-tested?

mareklibra:
Comments of @suomiy have been addressed, we can merge. If something missing, let's fix it in a follow-up.

PR #55, opened 2018-10-16
Choose bootable network with dropdown

rawagner:
depends on https://github.com/kubevirt/web-ui-components/pull/52

[storybook](https://agitated-fermi-950e49.netlify.com/?component=Wizards%2FCreateVmWizard%2Fj&fixture=namespaces)

### DIFF
--- a/src/components/Table/TableFactory.js
+++ b/src/components/Table/TableFactory.js
@@ -34,7 +34,7 @@ export const TableFactory = props => {
 
   const onTableChange = (rows, data) =>
     onChange(rows, data, props.onRowUpdate, props.onRowDeleteOrMove, props.onRowActivate);
-  const error = props.error !== '' ? <Alert>{props.error}</Alert> : undefined;
+  const error = props.error !== undefined ? <Alert>{props.error}</Alert> : undefined;
 
   return (
     <React.Fragment>
@@ -45,12 +45,17 @@ export const TableFactory = props => {
   );
 };
 
+TableFactory.defaultProps = {
+  error: undefined,
+  actionButtons: []
+};
+
 TableFactory.propTypes = {
-  actionButtons: PropTypes.array.isRequired,
+  actionButtons: PropTypes.array,
   onRowUpdate: PropTypes.func.isRequired,
   onRowDeleteOrMove: PropTypes.func.isRequired,
   onRowActivate: PropTypes.func.isRequired,
-  error: PropTypes.string.isRequired,
+  error: PropTypes.string,
   columns: PropTypes.array.isRequired,
   rows: PropTypes.array.isRequired
 };
--- a/src/components/Table/TableFactory.js
+++ b/src/components/Table/TableFactory.js
@@ -34,7 +34,7 @@ export const TableFactory = props => {
 
   const onTableChange = (rows, data) =>
     onChange(rows, data, props.onRowUpdate, props.onRowDeleteOrMove, props.onRowActivate);
-  const error = props.error !== undefined ? <Alert>{props.error}</Alert> : undefined;
+  const error = props.error && <Alert>{props.error}</Alert>;
 
   return (
     <React.Fragment>
--- a/src/components/Wizards/CreateVmWizard/NetworksTab.js
+++ b/src/components/Wizards/CreateVmWizard/NetworksTab.js
@@ -1,10 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { FormFactory } from '../../Form';
 import { TableFactory } from '../../Table/TableFactory';
 import { ACTIONS_TYPE, DELETE_ACTION } from '../../Table/constants';
 import { POD_NETWORK } from '../../../constants';
-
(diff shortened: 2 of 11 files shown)


src/components/Wizards/CreateVmWizard/NetworksTab.js
```diff
@@ -239,20 +217,67 @@ export class NetworksTab extends React.Component {
...
+  };
+
   render() {
     const columns = this.getColumns();
     const actionButtons = this.getActionButtons();
 
+    let pxeForm;
+    if (this.props.pxeBoot) {
+      const pxeNetworks = this.state.rows.filter(row => row.network !== POD_NETWORK && row.network !== '');
+      const bootableNetwork = pxeNetworks.find(row => row.isBootable);
```
suomiy: I think it would be better to disable the dropdown if there are no nics to select from

suomiy: @lizsurette what is your opinion on this?

rawagner:
> I think this is buggy behaviour:
> 
> Steps:
> 
> * create PXE bootable nic
> * select it as PXE nic
> * remove the nic
> * create another PXE bootable nic
> * that nic is automatically selected without user clicking on the select

That is intentional. I will select the first PXE-capable network, so user can click less (in some cases) :) WDYT @lizsurette ?

lizsurette:
> That is intentional. I will select the first PXE-capable network, so user can click less (in some cases) :) WDYT @lizsurette ?

Looks like maybe you updated it? If I create a PXE-capable network is it selected automatically in the dropdown (even the first time). I like it!

src/k8s/request.js
```diff
@@ -196,7 +200,13 @@ const setNetworks = (vm, basicSettings, networks) => {
 };
```
suomiy: why are we deleting disks and volumes at https://github.com/kubevirt/web-ui-components/pull/55/files#diff-b6fca8acbec69ecc7320bf2aa4787625R195 when using PXE?

rawagner:
> > Looks like maybe you updated it? If I create a PXE-capable network is it selected automatically in the dropdown (even the first time). I like it!
> 
> hm maybe I made a mistake last time I tried that, sorry about that
> 
> another scenario:
> 
> 1. select PXE in basic settings
> 2. create 2 pxe networks
> 3. select one
> 4. change the selected one to Pod Network
> 5. pxe is unselected in the dropdown + vm can be created
> 6. you can also remove the other pxe network (pxe is still unselected in the dropdown) + vm can be created

good catch ! Will have to fix that

rawagner:
> > > Looks like maybe you updated it? If I create a PXE-capable network is it selected automatically in the dropdown (even the first time). I like it!
> > 
> > 
> > hm maybe I made a mistake last time I tried that, sorry about that
> > another scenario:
> > 
> > 1. select PXE in basic settings
> > 2. create 2 pxe networks
> > 3. select one
> > 4. change the selected one to Pod Network
> > 5. pxe is unselected in the dropdown + vm can be created
> > 6. you can also remove the other pxe network (pxe is still unselected in the dropdown) + vm can be created
> 
> good catch ! Will have to fix that

fixed

src/k8s/request.js
rawagner: This is just temporary as we do not have Storage tab yet. I need to make sure that there is no disk/volume when PXE is used as it might cause troubles. The disk/volume deletion should be removed once Storage Tab is merged.

suomiy: ok, I removed it in #10

src/components/Table/TableFactory.js
```diff
@@ -34,7 +34,7 @@ export const TableFactory = props => {
 
   const onTableChange = (rows, data) =>
     onChange(rows, data, props.onRowUpdate, props.onRowDeleteOrMove, props.onRowActivate);
-  const error = props.error !== '' ? <Alert>{props.error}</Alert> : undefined;
+  const error = props.error !== undefined ? <Alert>{props.error}</Alert> : undefined;
```
mareklibra: Shouldn't it be `props.error ? <Alert/> : null` ?

Or even better - change L43:
```
{props.error && <Alert> ...}
```

src/components/Wizards/CreateVmWizard/NetworksTab.js
mareklibra: IMO, by disabling the dropdown, we would denote that the flow does not allow the value to be selected.
It's different from: there is nothing to be selected from.

I know it's too much for now, but eventually I would better render an action for the user to allow him to change such an "empty" state.

src/components/Table/TableFactory.js
rawagner: @mareklibra fixed

rawagner:
@mareklibra fixed props/alert stuff fixed and rebased on master. Can you take a look again ?

mareklibra:
@suomiy , I believe your comments have been addressed. If not, please open an issue or so ...

PR #59, opened 2018-10-16
Prettier: allow trailing commas

suomiy:


mareklibra:
I like this. But as this is a matter of preference, let's see what others think too.

pcbailey:
Personally, I don't use trailing commas, but I also don't have a strong opinion on the issue, either way. So, I'm good with whatever the majority decides.

rawagner:
Im fine with merging

vojtechszocs:
With trailing commas, the diff can be smaller, so I'm OK with having them.

If we do so, let's stay consistent with our comma usage.

PR #66, opened 2018-10-24
Make constants out of pxe annotations

rawagner:


src/constants/index.js
```diff
@@ -31,6 +31,9 @@ export const PROVISION_SOURCE_TEMPLATE = 'Template';
 
 export const POD_NETWORK = 'Pod Networking';
 
+export const ANNOTATION_PXE_INTERFACE = 'cnv.ui.pxeInterface';
```
mareklibra: nitpick: Is this constant explicitly defined somewhere in documentation (or so)?
Does it make sense to add reference link as a comment here?

Same as the constants above.

mareklibra: nitpick: can we order the lines alphabetically or at least group them by context (like `TEMPLATE_` or `ANNOTATION_`)?

rawagner: no, I just made it up. This is used only by us (UI) for later VM edit

rawagner: will do!

ohadlevy:
:+1: does it make sense to extract other things such as virtio etc?

src/constants/index.js
rawagner: For other constants..some of them are used in common-templates, but there are no docs, we could link to the common-templates repo ..?
Others are used in kubevirt itself (like API_VERSION)..we could link to user guide's VM example.. but that does not seem correct..

rawagner:
@ohadlevy virtio is already a constant. I added a few more from strings that are used in `k8s/request.js`. If you have any suggestions let me know.

src/constants/index.js
rawagner: done

PR #67, opened 2018-10-24
Move template util methods to utils/templates

rawagner:
Rename utils/template and utils/validation to utils/templates and utils/validations


rawagner:
@mareklibra rebased on master

PR #69, opened 2018-10-24
Use integer input type for CPU and Memory fields

rawagner:


src/components/Form/Integer.js
```diff
@@ -48,6 +48,17 @@ const fixInt = (value, oldValue, keyCode) => {
   return '0';
 };
 
+const isInputValid = (allowedKeys, keyCode, targetValue, validRegex) => {
+  if (allowedKeys.includes(keyCode)) {
```
mareklibra: Please let's discuss this offline.
If the semantics is as expected, this function can be simplified.
If not, the overall validation flow needs to be explained, seems to be over-complicated and possible hard to unit test.

src/components/Form/Integer.js
```diff
@@ -48,6 +48,17 @@ const fixInt = (value, oldValue, keyCode) => {
...
+    return !(
+      targetValue === '' &&
+      (keyCode === KEY_CODES[0] || keyCode === KEY_CODES.NUMPAD[0]) &&
```
suomiy: This is just a specific optimization (only first value of positive int), and does not concern other use cases.
What is the desired use case of this? This functionality is also tackled by timeouted validRegex (is it missing something?).

rawagner:
@suomiy PR updated, please take a look

PR #87, opened 2018-11-07
import-vm-apb template is using datavolumes not pvc

rawagner:


mareklibra:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1646947

PR #89, opened 2018-11-07
NewVmWizard: fix fixture

suomiy:
I am not sure if this component is still needed, because it is not used in kubevirt/we-ui? Maybe we should deprecate it.


rawagner:
@suomiy I agree, we should remove it, as it is not used anywhere. I will merge this and we can remove it in followup

PR #91, opened 2018-11-09
Add configurable URL Disk size

rawagner:


rawagner:
dv_size

lizsurette:
> default value is 10G, user can edit it if he wants. Field allows entering only positive number.
> 
> @jelkosz @lizsurette @mareklibra

@rawagner Would the user be able to see this as a disk and edit the size in the Storage step, too?

rawagner:
@lizsurette no. It will show only in Basic Settings step. But I plan to reimplement this for 1.4. We will have ability to create disk (not just attach it) so I will require, in case of URL image source, that user will create at least one disk.

lizsurette:
> @lizsurette no. It will show only in Basic Settings step. But I plan to reimplement this for 1.4. We will have ability to create disk (not just attach it) so I will require, in case of URL image source, that user will create at least one disk.

@rawagner Sounds great! Thanks for explaining :)

PR #93, opened 2018-11-09
Set firstBoot annotation to false if VM should start after creation

rawagner:
https://bugzilla.redhat.com/show_bug.cgi?id=1648245

PR #94, opened 2018-11-09
Set firstBoot annotation to false if VM should start after creation

rawagner:
Backport https://github.com/kubevirt/web-ui-components/pull/93 to v1.3


rawagner:
not planned for 1.3 yet, so I will close it for now.

PR #95, opened 2018-11-12
Add PXE boot patch util method

rawagner:


src/utils/utils.js
```diff
@@ -15,3 +19,37 @@ export const setNativeValue = (element, value) => {
...
+  if (annotations[ANNOTATION_FIRST_BOOT]) {
+    if (annotations[ANNOTATION_FIRST_BOOT] === 'true') {
+      patch.push({
+        op: 'replace',
+        path: `/metadata/annotations/${ANNOTATION_FIRST_BOOT}`,
+        value: 'false',
+      });
+    } else {
+      // find bootable disk and change boot order
+      const bootableDiskIndex = getDisks(vm).findIndex(disk => disk.bootOrder === 2);
```
suomiy: 1 and 2 could be shared constants between utils.test.js

src/utils/utils.js
```diff
@@ -15,3 +19,37 @@ export const setNativeValue = (element, value) => {
     valueSetter.call(element, value);
   }
 };
+
+export const getPxeBootPatch = vm => {
```
suomiy: Why is this method needed and where is it going to be used?

src/utils/utils.js
```diff
@@ -15,3 +19,37 @@ export const setNativeValue = (element, value) => {
...
+          {
+            op: 'replace',
+            path: `/spec/template/spec/domain/devices/interfaces/${bootableInterfaceIndex}/bootOrder`,
+            value: 2,
+          }
+        );
+      }
+    }
+  }
+  return patch;
```
suomiy: dunno if `patches` would describes it better

src/utils/utils.js
suomiy: nevermind, I found it in web-ui/

src/utils/utils.js
```diff
@@ -15,3 +19,37 @@ export const setNativeValue = (element, value) => {
...
+        patch.push(
+          {
+            op: 'replace',
+            path: `/spec/template/spec/domain/devices/disks/${bootableDiskIndex}/bootOrder`,
+            value: 1,
+          },
+          {
+            op: 'replace',
+            path: `/spec/template/spec/domain/devices/interfaces/${bootableInterfaceIndex}/bootOrder`,
+            value: 2,
```
suomiy: are we sure we want to even include the  vm interface in boot order?

rawagner: I would like to include the boot order. When you try to pxe boot and fail the installation for some reason, next time you run the VM it will try to boot from disk (which wont work because installation failed) and it will fallback to second boot device which will be PXE. Seems like a nice "feature" to me. Its better than failing the boot altogether because if you have boot order defined, the devices without boot order are not considered at all.

suomiy: Is it really the case? I have not tried that, but the usual behavior is that the boot order falls back to the PXE only when the disk is missing.

rawagner: I guess you are right :) Updated PR is now removing the bootOrder prop from NICs.

src/utils/utils.js
rawagner: renamed to patches

src/utils/utils.js
rawagner: created constants out of 1 and 2 and they are now used in utils/utils.test/request

PR #101, opened 2018-11-14
Remove tools folder from coverate report

rawagner:


suomiy:
Even though we have tests in tools I agree that we do not have to collect coverage on that.

rawagner:
I wanted to collect coverage from the bits that we are actually shipping, but you are right, this is being used during the build so we should re-enable coverage report from `tools` dir.

PR #102, opened 2018-11-14
Cloud-init fields are not required

rawagner:


src/k8s/request.js
```diff
@@ -234,15 +234,20 @@ const addCloudInit = (vm, getSetting) => {
     };
     addDisk(vm, cloudInitDisk);
 
-    const userDataObject = {
-      users: [
+    const userDataObject = {};
+
+    if (getSetting(AUTHKEYS_KEY)) {
```
suomiy: is cloud init useful with any combination of these?

PR #105, opened 2018-11-15
Add VM name validation against DNS1123 subdomain spec

rawagner:
part of https://github.com/kubevirt/web-ui/issues/92 -  we will also need to add check if VM already exists

### DIFF
--- a/src/components/Form/FormFactory.js
+++ b/src/components/Form/FormFactory.js
@@ -72,7 +72,7 @@ export const FormFactory = ({
     .map(key => {
       const field = fields[key];
       const values = fieldsValues[key];
-      const validMsg = get(values, 'validMsg', null);
+      const validation = get(values, 'validation');
       const value = get(values, 'value');
 
       const child = getFormElement({
@@ -85,7 +85,7 @@ export const FormFactory = ({
       return (
         <FormGroup
           key={key}
-          validationState={validMsg ? 'error' : null}
+          validationState={validation ? validation.type : null}
           className={field.noBottom ? 'form-group-no-bottom' : undefined}
         >
           <Col sm={labelSize} className={textPosition}>
@@ -98,7 +98,7 @@ export const FormFactory = ({
           </Col>
           <Col sm={controlSize}>
             {child}
-            <HelpBlock>{validMsg}</HelpBlock>
+            <HelpBlock>{validation ? validation.message : null}</HelpBlock>
           </Col>
         </FormGroup>
       );
--- a/src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
+++ b/src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
@@ -4,6 +4,7 @@ import { get } from 'lodash';
 import { FormFactory } from '../../Form/FormFactory';
 import { getName } from '../../../utils/selectors';
 import { getTemplate } from '../../../utils/templates';
+import { validateDNS1123SubdomainValue, getValidationObject } from '../../../utils/validations';
 import {
   getFlavors,
   getOperatingSystems,
@@ -19,6 +20,7 @@ import {
   PROVISION_SOURCE_TEMPLATE,
   PROVISION_SOURCE_URL,
   TEMPLATE_TYPE_VM,
+  VALIDATION_ERROR_TYPE,
 } from '../../../constants';
 import {
   NAME_KEY,
@@ -39,6 +41,7 @@ import {
   AUTHKEYS_KEY,
   IMAGE_URL_SIZE_KEY,
 } from './constants';
+import { ERROR_IS_REQUIRED } from './strings';
 
 export const getFormFields = (basicSettings, namespaces, templates, selectedNamespace) => {
   const workloadProfiles = getWorkloadProfiles(basicSettings, templates);
@@ -63,6 +66,7 @@ export const getFormFields = (basicSettings, namespaces, templates, selectedName
       id: 'vm-name',
       title: 'Name',
       required: true,
+      validate: validateDNS1123SubdomainValue,
     },
     [DESCRIPTION_KEY]: {
       id: 'vm-description',
@@ -206,7 +210,10 @@ const validateWizard = (formFields, values) => {
 
   // check if all fields are valid
   for (const key in values) {
-    if (get(values[key], 'validMsg') && (formFields[key].isVisible ? formFields[key].isVisible(values) : true)) {
+    if (
+      get(values[key], 'validation.type') === VALIDATION_ERROR_TYPE &&
+      (formFields[key].isVisible ? formFields[key].isVisible(values) : true)
+    ) {
       wizardValid = false;
       break;
     }
@@ -215,9 +222,9 @@ const validateWizard = (formFields, values) => {
   return wizardValid;
 };
 
-const asValueObject = (value, validMsg) => ({
+const asValueObject = (value, validation) => ({
   value,
-  validMsg,
+  validation,
 });
 
 const publish = ({ basicSettings, namespaces, templates, selectedNamespace, onChange }, value, target, formFields) => {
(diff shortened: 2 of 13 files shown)


src/utils/validations.js
```diff
@@ -1 +1,22 @@
...
+  }
+  if (!value.charAt(value.length - 1).match('[a-zA-Z0-9]')) {
+    return DNS1123_END_ERROR;
+  }
+  for (let i = 1; i < value.length - 1; i++) {
+    if (!value.charAt(i).match('[-a-zA-Z0-9.]')) {
+      return `${DNS1123_CONTAINS_ERROR} ${value.charAt(i)}`;
+    }
+  }
+  return undefined;
```
mareklibra: Can we return i.e. `null` or `false`?

src/utils/validations.js
```diff
@@ -1 +1,22 @@
+import { DNS1123_START_ERROR, DNS1123_END_ERROR, DNS1123_CONTAINS_ERROR, DNS1123_EMPTY_ERROR } from './strings';
+
 export const isPositiveNumber = value => value && value.match(/^[1-9]\d*$/);
+
+// DNS-1123 subdomain
+export const validateDNS1123SubdomainValue = value => {
+  if (!value) {
```
mareklibra: what about maximal length? (253 for dns1123 but do you know limit for k8s name?)

src/utils/validations.js
```diff
@@ -1 +1,22 @@
...
+    return DNS1123_EMPTY_ERROR;
+  }
+  if (!value.charAt(0).match('[a-zA-Z0-9]')) {
```
mareklibra: Looking at implementation of k8s DNS1123 validation, upper case characters are not allowed.
I think it's ok for VM name let's write a comment if we reference 1123 in the function name.

src/utils/validations.js
rawagner: null it is :)

src/utils/validations.js
rawagner: good catch, I will add the length check

src/utils/validations.js
rawagner: backend will actually convert the name to lowercase, maybe I could show warning that such conversion will happen ? WDYT ?

mareklibra: Warning would be great if it appears once the user enters upper case letter.

src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
```diff
@@ -63,6 +64,7 @@ export const getFormFields = (basicSettings, namespaces, templates, selectedName
       id: 'vm-name',
       title: 'Name',
       required: true,
+      validate: currentValue => validateDNS1123SubdomainValue(currentValue),
```
suomiy: unnecessary lambda

src/utils/validations.js
lizsurette: > backend will actually convert the name to lowercase, maybe I could show warning that such conversion will happen ? WDYT ?

I feel like most UI users either won't see these names of the backend or will understand that it needed to be changed. Just wondering if a warning would be concerning to these folks. I'm trying to think of something subtle, but don't have any good ideas yet :)

rawagner: They will see `myvirtualmachine` instead of `MyVirtualMachine`  in VM List for example. Maybe just info instead of warning ? Or nothing, im fine with either. Its your call @lizsurette

lizsurette: @rawagner Ah sorry I misunderstood then :) Yes, definitely makes sense to highlight this to the user as to what's happening since it will change how it's represented in the UI. Info makes sense as a level of notification to me rather than warning.

src/utils/validations.js
```diff
@@ -1 +1,22 @@
...
+    return DNS1123_START_ERROR;
+  }
+  if (!value.charAt(value.length - 1).match('[a-zA-Z0-9]')) {
+    return DNS1123_END_ERROR;
+  }
+  for (let i = 1; i < value.length - 1; i++) {
```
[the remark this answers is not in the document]
rawagner: hm, i dont think so. That way I would not validate next-to-last character

src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
rawagner: fixed

src/utils/validations.js
```diff
@@ -1 +1,22 @@
...
+
+// DNS-1123 subdomain
+export const validateDNS1123SubdomainValue = value => {
+  if (!value) {
+    return DNS1123_EMPTY_ERROR;
+  }
+  if (!value.charAt(0).match('[a-zA-Z0-9]')) {
+    return DNS1123_START_ERROR;
+  }
+  if (!value.charAt(value.length - 1).match('[a-zA-Z0-9]')) {
```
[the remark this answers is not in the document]
rawagner: extracted

src/utils/validations.js
rawagner: Latest PR update adds possibility to specify the type of validation message (error/warning/info). 

info message about uppercase to lowercase conversion
validation

error
validation1


and I do not have warning example since we do not have any warnings in this case :) just error/info

src/components/Form/FormFactory.js
```diff
@@ -98,7 +98,7 @@ export const FormFactory = ({
           </Col>
           <Col sm={controlSize}>
             {child}
-            <HelpBlock>{validMsg}</HelpBlock>
+            <HelpBlock>{valid ? valid.message : null}</HelpBlock>
```
ohadlevy: shouldn't this be invalid? (assuming its a validation error hence its not valid?)

src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
```diff
@@ -4,6 +4,7 @@ import { get } from 'lodash';
 import { FormFactory } from '../../Form/FormFactory';
 import { getName } from '../../../utils/selectors';
 import { getTemplate } from '../../../utils/templates';
+import { validateDNS1123SubdomainValue, getValidationObject } from '../../../utils/validations';
```
ohadlevy: do we know if there is any similar validation on the api side? just wondering if we are creating a mix UX where some validations are enforced only in the UI.

src/components/Form/FormFactory.js
rawagner: validation can return warning or info too. So the result does not always has to erroneous. So the `errors` does not really fit here IMHO.

src/components/Wizard/CreateVmWizard/strings.js
```diff
@@ -6,6 +6,7 @@ export const STEP_RESULT = 'Result';
 export const NEXT = 'Next';
```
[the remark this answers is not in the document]
rawagner: I agree, since the strings.js is already in the code, I will fix it in follow up https://github.com/kubevirt/web-ui-components/issues/107

src/components/Form/FormFactory.js
ohadlevy: do you have examples in mind? the only usage case I could think of would be a weak password? regardless, in that case, we should have messages that are errors, and others that are info/warn? in sounds to me we need to know which level the field validation are, so we can highlight in red / yellow the entire field for errors?

rawagner: example is here https://github.com/kubevirt/web-ui-components/pull/105#discussion_r234536393 , the message is actually highlighted red/yellow/grey (error/warning/info) by https://github.com/kubevirt/web-ui-components/pull/105/files#diff-a64432c062870774c8559d8af0c30ff7R88 (from pf-react)

rawagner: renamed `valid` to `validation`

src/utils/validations.js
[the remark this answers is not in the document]
suomiy: ah, sorry about that

rawagner:
@suomiy  or @mareklibra  please take a look again

src/utils/validations.js
andybraren: The info message and behavior look good to me. I think adding `text-transform: lowercase;` to the `input` so that it behaves like [this](https://codepen.io/abraren/pen/MzxRJL?editors=1100) would be a nice future enhancement. If this is added, the info message would probably change to "Name must be lowercase".

rawagner: followup issue https://github.com/kubevirt/web-ui-components/issues/129

PR #108, opened 2018-11-19
Introduce VmStatus component

mareklibra:
The component and related `getVmStatusDetail()`, `getVmStatus()` functions provides visualization and mapping of a VM to it's status information.

Features:
- `Error` status renders with a link to pod/vm event page (based on error context)
- text status info rendered along its icon


src/components/VmStatus/VmStatus.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  if (get(vm, 'status.created', false)) {
+    // created but not yet ready
+    if (pod) {
+      // pod created, so check for it's potential error
+      const notReadyCondition = get(pod, 'status.conditions', []).find(condition => condition.status !== 'True');
+      if (notReadyCondition) {
+        // at least one pod condition not met, let the user analyze issue via Pod events
+        return { status: VM_STATUS_POD_ERROR, message: notReadyCondition.message };
```
rawagner: Can we also include check for VMI errors ?

mareklibra: My thinking here is, if the VM `status.ready` is`true`, the `VMI` should not report any error.

And as the user is interested in just a limited (means "status") information, we do not need to complicate things by adding too many details.
If a VM reports an error, the UI leads the user for appropriate pages to resolve/analyze.

suomiy:
Can we also display migration status in this component? It needs to be aware of Migration object.
Or maybe another column? VM could be running(scheduling) and also migrating at the same time.

related issue: https://github.com/kubevirt/web-ui-components/issues/111

mareklibra:
We can reflect that in the single Status.

mareklibra:
Failed importer flow:
- in Create VM Wizard, the user selects `URL` as `Provision Source` and provides incorrect value
- once the VM is started, the VM gets to `VM Error` state and the user can be navigated to the `VM Events` page for additional details

With https://github.com/kubevirt/web-ui/issues/85, the user gets adequate information for this state as the issue is listed among in `VM` events.

I think we do not need to go in farther detail in the `VmStatus`, especially we do not need to analyze the importer pod.
@rawagner , wdyt?

mareklibra:
I've added extra handling of the importer pod.

PR #110, opened 2018-11-20
add migration dialog

suomiy:


rawagner:
@suomiy Im getting error when trying to migrate VMI

`{"response":{},"json":{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Internal error occurred: Object 'apiVersion' is missing in 'object has no apiVersion field'","reason":"InternalError","details":{"causes":[{"message":"Object 'apiVersion' is missing in 'object has no apiVersion field'"}]},"code":500}}`

suomiy:
followup issue: #111

Issue #77, opened 2018-10-31
Do not use hardcoded indexes


rawagner:
Storage/NetworkTab errors and tabs settings in CreateVmWizard

PR #113, opened 2018-11-23
Use object instead of array in CreateVmWizard stepData. #77

rawagner:
https://github.com/kubevirt/web-ui-components/issues/77


rawagner:
@mareklibra added test, please review

PR #115, opened 2018-11-27
Introduce VmConsoles component

mareklibra:
Based on `AccessConsoles` component from `paternfly/react-consoles` project.

### DIFF
--- a/src/components/VmConsoles/SerialConsoleConnector.js
+++ b/src/components/VmConsoles/SerialConsoleConnector.js
@@ -11,7 +11,7 @@ const { debug, info, error } = console;
  * Kubevirt serial console is accessed via WebSocket proxy in k8s API.
  * Protocol used is "plain.kubevirt.io", means binary and single channel - forwarding of unix socket only (see subresource.go in kubevirt).
  */
-class SerialConcolseConnector extends React.Component {
+class SerialConsoleConnector extends React.Component {
   state = {
     status: LOADING,
     passKeys: false,
@@ -128,7 +128,7 @@ class SerialConcolseConnector extends React.Component {
     );
   }
 }
-SerialConcolseConnector.propTypes = {
+SerialConsoleConnector.propTypes = {
   vmi: PropTypes.object.isRequired,
   host: PropTypes.string.isRequired,
   path: PropTypes.string.isRequired,
@@ -136,4 +136,4 @@ SerialConcolseConnector.propTypes = {
   WSFactory: PropTypes.func.isRequired, // reference to OKD utility
 };
 
-export default SerialConcolseConnector;
+export default SerialConsoleConnector;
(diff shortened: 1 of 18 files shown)


src/components/VmConsoles/SerialConsoleConnector.js
```diff
@@ -0,0 +1,150 @@
+import React from 'react';
```
rawagner: can you rename VmConsoles to VmConsole (singular) ? We are using singular for also for Wizard etc..

src/components/VmConsoles/LoadingMock.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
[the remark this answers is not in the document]
mareklibra: This would require changes to our test environment since file like `tests/tests/LoadingMock.test.js` would be required.
To fix, the component is moved to `VmConsoles.fixture.js`.

src/components/VmConsoles/SerialConsoleConnector.js
```diff
@@ -0,0 +1,150 @@
...
+ */
+class SerialConcolseConnector extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      status: LOADING,
+      passKeys: false,
+    };
+
+    this.onBackendDisconnected = this.onBackendDisconnected.bind(this);
```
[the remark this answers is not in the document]
mareklibra: Not anymore. I just took my already tested code from other project. Will rewrite it.

src/components/VmConsoles/SerialConsoleConnector.js
mareklibra: I can, it's nothing I would fight for. But the component wraps access to multiple consoles. I thing the plural express it's meaning better in this case. 
The `Wizard` is just a folder - container for other (unrelated) components. The `VmConsoles` is the component itself focused on managing multiple sub-components.

rawagner: you are right. Lets keep it as it is.

src/components/VmConsoles/SerialConsoleConnector.js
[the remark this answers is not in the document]
mareklibra: done - the component uses class properties now.

src/components/VmConsoles/tests/SerialConsoleConnector.test.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
```
rawagner: having -2% on code coverage is kind of sad :( Do you think its possible to write some meaningful tests for SerialConsoleConnector ?

mareklibra: I've covered VmConsoles, so improved slightly.

The SerialConsoleConnector would require to mock backend by writing much more code than for the implementation. And this code will need to be maintained. I doubt it would bring any real benefit to us.
Let's postpone testing of this till integration tests are ready, so we will be able to connect to a real backend.

Issue #136, opened 2018-12-05
Fix dependency issues


vojtechszocs:
While working on [KubeVirt extension of OpenShift Console](https://github.com/openshift/console/pull/888), I've found some dependency issues.

---

#### 1) symptoms observed in `kubevirt-web-ui-components` library

```bash
$ git log -1 --format="%h %s"
bc2770f Add Templates Wizard and integration with Create Vm Wizard (#123)

$ rm -rf ./node_modules && yarn install
..
warning " > @patternfly/react-console@1.9.1" has unmet peer dependency "patternfly@^3.58.0".
warning " > @patternfly/react-console@1.9.1" has unmet peer dependency "react-bootstrap@^0.32.1".
..
```

Above two warnings come from `@patternfly/react-console` package which has peer deps on `patternfly` and `react-bootstrap` ([`react-console` package.json](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/react-console/package.json)).

However:

```bash
$ cd /path/to/patternfly-react/packages/patternfly-3/react-console
$ grep -r "from 'patternfly" src # shows only imports from "patternfly-react"
$ grep -r "from 'react" src      # shows only imports from "react"
```

To me, it seems that `patternfly` and `react-bootstrap` shouldn't be defined as peer deps of `@patternfly/react-console` package. This package already has peer dep on `patternfly-react`, which in turn depends on `patternfly` and `react-bootstrap` via __normal deps__ ([`patternfly-react` package.json](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/package.json)).

In other words, `patternfly-react` doesn't treat `patternfly` and `react-bootstrap` as peer deps, but as normal deps. Any package depending on `patternfly-react` shouldn't complicate existing peer dep requirements as it causes warnings in target applications.

__Proposed fix:__ remove `patternfly` and `react-bootstrap` peer deps in `@patternfly/react-console` package.

---

#### 2) symptoms observed in `openshift/console` target application

With [KubeVirt extension PoC](https://github.com/openshift/console/pull/888) PR applied:

```bash
$ git log -1 --format="%h %s"
9575c7b0b KubeVirt extension PoC

$ cd frontend
$ rm -rf ./node_modules && yarn install
..
warning " > kubevirt-web-ui-components@0.1.7" has incorrect peer dependency "react@16.3.x".
[...]

PR #139, opened 2018-12-06
Update dependencies and bump React version

vojtechszocs:
This is a follow-up to issue #136.

`react` and `react-dom` peer dependencies have their versions loosened from `16.3.x` to `16.x`. This aligns with Console UI currently depending on `react-dom@16.x`.

`react` and `react-dom` version locks bumped to `16.6.3`. This aligns with Console UI currently depending on `react@16.6.3`.

Following dependencies are now treated as regular `dependencies`:

- `@patternfly/react-console`
- `js-yaml`
- `lodash`
- `react-dnd`
- `react-dnd-html5-backend`
- `reactabular-dnd`

Everything in `devDependencies` is now truly needed for build/development only (except for entries that must match `peerDependencies`).

@mareklibra @rawagner Console UI [package.json](https://github.com/openshift/console/blob/master/frontend/package.json) already contains `react-dnd` and `react-dnd-html5-backend`. It seems that we could treat these as peer dependencies from `kubevirt-web-ui-components` perspective - what do you think?


mareklibra:
@vojtechszocs , I agree with these changes.
There might be a conflict with the patternfly/react-consoles when consuming by web-ui due to hacks I needed to do to keep the changes to OKD core at bare minimum. But let's resolve it once they arise.

PR #147, opened 2018-12-17
WIP: Convert datavolume name to lowercase.

rawagner:
https://bugzilla.redhat.com/show_bug.cgi?id=1659347

### DIFF
--- a/src/k8s/request.js
+++ b/src/k8s/request.js
@@ -433,7 +433,7 @@ const addDataVolume = (vm, storage, getSetting) => {
     name: storage.name,
     dataVolume: {
       ...(storage.templateStorage ? storage.templateStorage.volume.dataVolume : {}),
-      name: `${storage.name}-${getSetting(NAME_KEY)}`,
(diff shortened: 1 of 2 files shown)


suomiy:
wouldn't it be better to move this into the storage tab together with DNS-1123 validation?

rawagner:
Disk can contain uppercase letters, problem is with datavolumes and as I just found out with volumes too. I could convert the volumes to lowercase too.. but there is https://github.com/kubevirt/kubevirt/pull/1570 where they plan to remove volumeName altogether and reference by name only which would mean that disk name with uppercase will not be valid in the future . Not sure when the PR is going to be merged but we could go ahead and add DNS-1123 validation to storage tab as you suggested

suomiy:
> Disk can contain uppercase letters

I am not sure if we have to be 100% compliant with the API, but I think DNS-1123 validation shouldn't hurt anything and the UI would be more consistent.

rawagner:
replaced by https://github.com/kubevirt/web-ui-components/pull/169

PR #148, opened 2018-12-17
export isBeingMigrated function

suomiy:
needed for cancel migration pull request in kubevirt/web-ui


suomiy:
also fixes proptypes warning

Issue #104, opened 2018-11-15
Add custom cloud-init script option to cloud-init settings


rawagner:
@andybraren we need to allow user to add his own cloud-init yaml script see https://bugzilla.redhat.com/show_bug.cgi?id=1648317 can you please design something ? User should be able to choose if he wants to use his own script and copy/paste into some text area or use already existing basic hostname/ssh-key config. He cannot use both.

andybraren:
Here's what can be built quickly into the existing wizard:

<img width="531" alt="screen shot 2018-11-15 at 10 44 32 am" src="https://user-images.githubusercontent.com/9122899/48563957-755b8900-e8c3-11e8-81b7-44801f4462e0.png">

A new "Use custom script" checkbox should put above Hostname and Authenticated SSH Keys. It is off by default and has an info button next to its label (to assist users who may be unfamiliar).

<img width="510" alt="screen shot 2018-11-15 at 10 42 22 am" src="https://user-images.githubusercontent.com/9122899/48563861-47764480-e8c3-11e8-9b43-c687baefaf10.png">

If the user turns it on, the Hostname and Authenticated SSH Keys fields are hidden entirely and replaced with a resizable "Custom Script" text field for the user to copy/paste their script into.

andybraren:
I should also mention that we have a work-in-progress "Advanced" Step 4 for the Create VM wizard that would contain settings for Cloud-Init/Sysprep and other future features.

This idea is still rough so some fields may not be applicable if "Use custom script" is enabled, but here's what that flow might look like:

![cloud-init-advanced-concept](https://user-images.githubusercontent.com/9122899/48565146-d84e1f80-e8c5-11e8-96bb-81de8827bb98.png)

suomiy:
This is great! Moving some logic from the first tab will make it less cluttered and more organized.

PR #152, opened 2018-12-22
add custom cloud-init script to Create VM Dialog

suomiy:
implements #104 

no possibility to add files for now; only copy-paste text

- uploading a file (or multiple files) can be added in a followup

### DIFF
--- a/src/k8s/request/cloudInit.js
+++ b/src/k8s/request/cloudInit.js
@@ -1,0 +1,70 @@
+import { safeDump } from 'js-yaml';
+
+import { CLOUDINIT_DISK } from '../../constants';
+
+const resolveScriptHeader = userData => {
+  const validScriptHeaders = [
+    'Content-Type',
+    '#include',
+    '#!',
+    '#cloud-config',
+    '#upstart-job',
+    '#part-handler',
+    '#cloud-boothook',
+  ];
+
+  if (
+    typeof userData !== 'string' ||
+    userData.length === 0 ||
+    validScriptHeaders.find(header => userData.startsWith(header))
+  ) {
+    return userData;
+  }
+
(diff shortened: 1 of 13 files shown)


src/k8s/request/cloudInit.js
```diff
@@ -0,0 +1,69 @@
...
+  setUserData(userData) {
+    this.volume.cloudInitNoCloud.userData = userData;
+  }
+
+  setPredefinedUserData({ hostname, sshAuthorizedKeys }) {
+    const userDataObject = {};
+    if (sshAuthorizedKeys) {
+      userDataObject.users = [
+        {
+          name: 'default',
```
suomiy: note: I chose to use default account intstead of root which might not be enabled by default

sass/components/_CreateVmWizard.scss
```diff
@@ -1,3 +1,7 @@
 .modal .wizard-pf-main {
   padding-top: 10px;
 }
+
+#cloud-init-custom-script {
```
mareklibra: I am not strictly against it, but prefer use of classes instead of ids

src/k8s/request/cloudInit.js
mareklibra: lgtm
Can you please add similar comment to the code?

sass/components/_CreateVmWizard.scss
rawagner: if I understand it correctly, BEM also prefers classes instead of IDs. In any case, please also add kubevirt prefix

suomiy: +1, I will update this PR to the recent standard

suomiy:
fixed

suomiy:
rebased

PR #158, opened 2019-01-04
fix isBeingMigrated function

suomiy:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1661870

was not tried in real environment but should be correct according to https://github.com/kubevirt/kubevirt/blob/master/pkg/api/v1/types.go


rawagner:
looks good, merging. But we should try migration on upshift maybe.

PR #159, opened 2019-01-04
link pod in vm running status

suomiy:


mareklibra:
Do we need this? The launcher pod is referenced in another column, close to this status.

The original idea behind these status-field links was to lead the user to the point of
- current processing
- or error details

src/utils/okduris.js
```diff
@@ -1,10 +1,11 @@
 import { get } from 'lodash';
 
 export const getSubPagePath = (apiObj, model, subPage) => {
-  if (!apiObj) {
+  if (!apiObj || !model) {
```
mareklibra: How can this happen? The model is statically provided at build-time. I guess it's a bug if it is missing here.

jelkosz:
> Do we need this? The launcher pod is referenced in another column, close to this status.

not anymore :) https://github.com/kubevirt/web-ui/pull/155
The idea is to make the listing to have fewer columns so it will be a easier to navigate.

src/utils/okduris.js
suomiy: I guess it would be a bug if it wasn't there. I was just looking at it as a block and adding more validation to this function.

suomiy:
note: also, the link to pod is still in the vm detail (so it is there twice)

PR #165, opened 2019-01-08
link to pod when a container is in a CrashLoopBackOff

suomiy:


src/components/VmStatus/VmStatus.js
```diff
@@ -60,22 +65,59 @@ const StateStarting = ({ ...props }) => (
...
 );
 const StateError = ({ children, ...props }) => (
   <StateValue iconClass="pficon pficon-error-circle-o" {...props}>
     {children}
   </StateValue>
 );
-StateError.propTypes = {
-  children: PropTypes.string.isRequired,
+
+export const VmStatuses = props => {
```
rawagner: I do not like to have VmStatuses and VmStatus components. Cant we just modify VmStatus to handle IMPORTING and IMPORT_ERROR better ? 
Maybe Im wrong but it seems that we could just do `importerPods.map` in https://github.com/kubevirt/web-ui-components/blob/f4509456c5b2a5aa36fac532dcee8b4902ccfe05/src/components/VmStatus/VmStatus.js#L132 and also `verbose` would be based on number of importer pods ( importerPods.length > 1 then verbose is true).
Same applies for IMPORT_ERROR https://github.com/kubevirt/web-ui-components/blob/f4509456c5b2a5aa36fac532dcee8b4902ccfe05/src/components/VmStatus/VmStatus.js#L147

suomiy: This cannot be done because we are using VmStatus in VmRow where we show errors/statuses from all importer pods in hover text. VmDetail has more space so we show all the statuses. I think having two components makes more sense semantically.

rawagner: ah, ok. Thanks for explanation. Please rebase and I will merge it.

suomiy: rebased

PR #166, opened 2019-01-08
Add 'dev' script to do a development build

vojtechszocs:
Addressing comment https://github.com/kubevirt/web-ui-components/pull/150#issuecomment-451887728

- add `dev` script to do a development build (one-time) without any pre-checks
- ensure that regular production build (`build` script) has `NODE_ENV` set accordingly
- updated developer guideline

Q: why not use `--watch` for the `dev` script?
A: our build is split into two parts, JS and Sass. JS files are processed by Babel, Sass files are copied as-is. While Babel supports `--copy-files` to include files that weren't part of the JS compilation, this would require changing `--out-dir` and transforming file paths, for which I didn't find any suitable Babel plugin. We could use webpack and its watch mode on top of Babel, but I'm reluctant to add such complexity for now.

Q: is there actually a difference between production and development build?
A: not right now :smiley: it seems like `NODE_ENV` has no effect on our build outputs at the moment:

```sh
$ yarn build && mv dist dist_prod
$ yarn dev && mv dist dist_dev
$ diff -r dist_prod dist_dev # no results
```


suomiy:
Watch mode would be nice to have though. Mainly because web-ui dev watch mode breaks when we delete our dist. Then web-ui dev needs to be restarted.

web-ui dev output:

```
ERROR in ./public/style.scss
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from ./node_modules/thread-loader/dist/cjs.js):
Thread Loader (Worker 9)

ERROR in ..-components/dist/js/index.js
Module build failed (from ./node_modules/thread-loader/dist/cjs.js):
Thread Loader (Worker 8)
ENOENT: no such file or directory, open '$AT_LOCATION/kube-web-ui-components/dist/js/index.js'

```

PR #170, opened 2019-01-09
Show full  OS names.

rawagner:
https://bugzilla.redhat.com/show_bug.cgi?id=1648335

### DIFF
--- a/src/components/Form/Dropdown.js
+++ b/src/components/Form/Dropdown.js
@@ -2,43 +2,43 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { ButtonGroup, DropdownButton, MenuItem, noop } from 'patternfly-react';
 
-export const Dropdown = ({ id, value, disabled, onChange, onBlur, choices }) => (
-  <ButtonGroup justified key={id}>
-    <DropdownButton
-      id={id}
-      bsStyle="default"
-      className="kubevirt-dropdown"
-      title={value}
-      onSelect={onChange}
-      onBlur={onBlur}
-      disabled={disabled}
-    >
-      {choices.map(choice => {
-        const isObject = typeof choice === 'object';
-        const key = isObject ? choice.id || choice.name : choice;
-        const val = isObject ? choice.name : choice;
+export const Dropdown = ({ id, value, disabled, onChange, choices }) => {
+  const title = typeof value === 'object' ? value.name || value.id : value;
+  return (
+    <ButtonGroup justified key={id}>
+      <DropdownButton
+        id={id}
+        bsStyle="default"
+        className="kubevirt-dropdown"
+        title={title}
+        disabled={disabled}
+        onSelect={onChange}
+      >
+        {choices.map(choice => {
+          const isObject = typeof choice === 'object';
+          const key = isObject ? choice.id || choice.name : choice;
+          const val = isObject ? choice.name : choice;
 
-        return (
-          <MenuItem key={key} eventKey={key}>
-            {val}
-          </MenuItem>
-        );
-      })}
-    </DropdownButton>
-  </ButtonGroup>
-);
+          return (
+            <MenuItem key={key} eventKey={choice}>
+              {val}
+            </MenuItem>
+          );
+        })}
+      </DropdownButton>
+    </ButtonGroup>
+  );
+};
 
 Dropdown.defaultProps = {
   onChange: noop,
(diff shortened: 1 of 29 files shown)


src/components/Form/Dropdown.js
```diff
@@ -2,31 +2,28 @@ import React from 'react';
...
-      title={value}
-      onSelect={onChange}
-      onBlur={onBlur}
-      disabled={disabled}
-    >
-      {choices.map(choice => {
-        const isObject = typeof choice === 'object';
-        const key = isObject ? choice.id || choice.name : choice;
-        const val = isObject ? choice.name : choice;
+export const Dropdown = ({ id, value, disabled, onChange, onBlur, choices }) => {
```
suomiy: I am not in a favor of this approach. This will create unnecessary amount of callback handlers. Can't we rather lookup the name somewhere higher, e.g. in `request.js`?

src/components/Form/Dropdown.js
```diff
@@ -2,31 +2,28 @@ import React from 'react';
...
+  const title = typeof value === 'object' ? value.name || value.id : value;
+  return (
+    <ButtonGroup justified key={id}>
+      <DropdownButton id={id} bsStyle="default" className="kubevirt-dropdown" title={title} disabled={disabled}>
+        {choices.map(choice => {
+          const isObject = typeof choice === 'object';
+          const key = isObject ? choice.id || choice.name : choice;
+          const val = isObject ? choice.name : choice;
+          const onSelect = isObject ? () => onChange(key, val) : () => onChange(key);
+          const blur = isObject ? () => onBlur(key, val) : () => onBlur(key);
```
suomiy: The blur is not much useful here. I guess we could remove the callback altogether.

src/components/Form/Dropdown.js
rawagner: apparently we can pass object to MenuItem's eventKey so Im passing choice as whole. With this I can keep onSelect on DropdownButton.

suomiy: nice

src/k8s/selectors.js
```diff
@@ -21,8 +22,13 @@ import {
 
 export const settingsValue = (basicSettings, key, defaultValue) => get(basicSettings, [key, 'value'], defaultValue);
 
-export const getLabel = (basicSettings, labelPrefix, value) =>
-  has(basicSettings, value) ? `${labelPrefix}/${settingsValue(basicSettings, value)}` : undefined;
+export const getLabel = (basicSettings, labelPrefix, value) => {
+  if (!has(basicSettings, value)) {
+    return undefined;
+  }
+  const val = settingsValue(basicSettings, value);
```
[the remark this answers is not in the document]
rawagner: I tried to make the function safer, take a look again please.

src/components/Wizard/CreateVmWizard/tests/__snapshots__/BasicSettingsTab.test.js.snap
```diff
@@ -96,19 +96,58 @@ exports[`<BasicSettingsTab /> for Create VM Template renders correctly 1`] = `
...
-          "fedora25",
-          "fedora24",
-          "fedora23",
-          "rhel7.0",
-          "ubuntu18.04",
-          "win2k12r2",
-          "win2k8r2",
-          "win2k8",
-          "win10",
+          Object {
```
fabiand: NACK from me.

This needs to be parsed from some libosinfo source.

/cc @MarSik

MarSik: Yes, the templates get the OS names from libosinfo. See here: https://github.com/kubevirt/common-templates/blob/master/templates/_linux.yaml#L17

fabiand: My point is: The array you create in the code, this should not happen, instead you should bundle the libosinfo db and parse teh contents from there. IMO

MarSik: Fabian, those are **test data**. You do not want to depend on 3rd party libraries there, the test inputs and outputs must be 100% predictable.

fabiand:
Oh. Thanks for the clarification.

Issue #168, opened 2019-01-09
If the common-templates are not installed, the VM details if failing on errors


jelkosz:
"templates.template.openshift.io "fedora-generic" not found" is shown in the VM details if the VM has been based on the mocked templates and not the ones coming from common-templates

rawagner:
fixed with https://github.com/kubevirt/web-ui-components/pull/171

PR #171, opened 2019-01-10
Fix VM Details form validity detection

rawagner:
Fallback to mocked vm templates if given template is not found in kubernetes. #168


src/components/Details/Flavor/Flavor.js
```diff
@@ -39,17 +40,28 @@ export class Flavor extends React.Component {
       const getTemplatePromise = this.props.k8sGet(TemplateModel, template.name, template.namespace);
       getTemplatePromise
         .then(result => {
```
suomiy: I think this would be better to move to another file and get templates here in one final promise with loadError in catch clause

PR #179, opened 2019-01-14
Improve URL provision source help text.

rawagner:


rawagner:
@aglitke is the help text correct for datavolume url source ? thanks!

PR #184, opened 2019-01-17
Clear IP addresses & FQDN when VM is off

pcbailey:
This PR fixes a bug [1] wherein the VMI isn't immediately deleted when the VM is shutdown, leading to the IP addresses still being displayed when the VM is in an off state. It also prevents the FQDN from being displayed if the pod isn't immediately deleted to avoid the potential of a similar issue occurring. 

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1665901.

### DIFF
--- a/src/components/Details/VmDetails/tests/__snapshots__/VmDetails.test.js.snap
+++ b/src/components/Details/VmDetails/tests/__snapshots__/VmDetails.test.js.snap
@@ -418,9 +418,7 @@ exports[`<VmDetails /> renders IP addresses correctly 1`] = `
             <dt>
               FQDN
             </dt>
-            <dd>
-              ---
-            </dd>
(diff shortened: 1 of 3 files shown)


src/components/Details/VmDetails/tests/__snapshots__/VmDetails.test.js.snap
```diff
@@ -418,9 +418,7 @@ exports[`<VmDetails /> renders IP addresses correctly 1`] = `
             <dt>
               FQDN
             </dt>
-            <dd>
-              ---
-            </dd>
+            <dd />
```
mareklibra: For consistency, this should be still `DASHES` if the FQDN is missing. But it's not a big deal, we can add in a follow-up.

PR #186, opened 2019-01-18
feature: Typeahead component

priley86:
adds the Typeahead component and an example in CreateVmWizard

* i am still getting comfortable w/ the semantics of this project, please critique 😄

### DIFF
--- a/src/components/Form/FormFactory.js
+++ b/src/components/Form/FormFactory.js
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 import { FormGroup, Col, ControlLabel, HelpBlock, Form, FieldLevelHelp } from 'patternfly-react';
 import { get } from 'lodash';
 
-import { Dropdown, Checkbox, Text, TextArea, Integer } from '.';
+import { Dropdown, Checkbox, Text, TextArea, Integer, TypeAhead } from '.';
 import { VALIDATION_INFO_TYPE, VALIDATION_ERROR_TYPE } from '../../constants';
 import { getValidationObject } from '../../utils/validations';
 import { ERROR_IS_REQUIRED } from '../Wizard/CreateVmWizard/strings';
@@ -19,9 +19,12 @@ export const getFormElement = props => {
     onChange,
     onBlur,
     choices,
+    choicesLabelKey,
     defaultValue,
     isControlled,
     disabled,
+    placeholder,
+    multiple,
     className,
     LoadingComponent,
   } = props;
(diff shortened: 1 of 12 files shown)


jelkosz:
@priley86 hey, could you please also add some screenshot how it looks?

jelkosz:
> Do we expect the users to always select the option or is just typing/copying the value not expected of them?

do you ask if there will be a situation that the user can enter a value which is not present in the list box yet? That is a great point! In oVirt we have couple of places where there are default values in a type ahead but the user is free to enter whatever he wants. An example of it is CPU type (IIRC) which oVirt lets you to select one populated from the database but, if your host has something else you want to utilize, than you can enter a custom value.

suomiy:
I see. That would be good in a case user has not updated their database (templates) with the newest OSes (or the OS is missing from the DB altogether), although we would still need the OS id. I am not sure if specifying custom OS like that is supported.

My question was much simpler. 
In a current state, The field will not validate if we just copy valid OS name. I think it would make more sense for this case to pass the validation. 

Or we can require the users to always pick the OS item. I guess this is the default use case anyway.

jelkosz:
> I see. That would be good in a case user has not updated their database (templates) with the newest OSes (or the OS is missing from the DB altogether), although we would still need the OS id. I am not sure if specifying custom OS like that is supported.

no, that is not supported. The combination of os/flavort/profile makes up a whole template from common-templates. We can not have just the name.
In current UI we would not use the support of having a free text the select box, but maybe later. But yeah, we can implement it once faced with the issue.

> 
> My question was much simpler.
> In a current state, The field will not validate if we just copy valid OS name. I think it would make more sense for this case to pass the validation.

Ah, I understand, thank you for showing it to me live :)
So the steps are:
1: paste a correct OS name into the typeahead
2: click somewhere out of it - it will be marked as validation error
But, if instead of clicking out you press the tab key, it will work correctly.

BTW a general design question - I dont really like the fact that it looks like a text field until you enter it. If we replaced all list boxes by it, it would look like we have bunch of text fields which don't help the user at all.

@andybraren, @lizsurette @matthewcarleton what do you think?

> 
> Or we can require the users to always pick the OS item. I guess this is the default use case anyway.

matthewcarleton:
I agree @jelkosz it would be better to indicate that it is more than just a stand text field. In Patternfly 4 we have an arrow indicating there will be a dropdown, if the user starts typing the options will appear with the option to delete the text. 
screen shot 2019-01-18 at 9 12 15 am
screen shot 2019-01-18 at 9 12 27 am

priley86:
added some notes in the [issue](https://github.com/patternfly/patternfly-react/issues/1215#issuecomment-455586419) for the upcoming pf4 typeahead

PR #198, opened 2019-01-29
Add BEM block to VmStatus component

vojtechszocs:
We need this for kubevirt/web-ui#189 - make element selectors within integration tests less dependent on rendered DOM structure.

cc @rhrazdil

### DIFF
--- a/docs/DevGuide.md
+++ b/docs/DevGuide.md
@@ -114,6 +114,12 @@ yarn test
 yarn dev
 ```
 
+### Run tests without any pre-checks?
+
+```sh
+yarn jest
+```
+
 ### Run specific tests?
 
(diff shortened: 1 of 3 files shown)


rawagner:
@vojtechszocs LGTM, but please fix the failing snapshot test

vojtechszocs:
Sorry guys, I'll fix the failing build right away.

vojtechszocs:
Please note, I've modified `VmStatus` to use `kubevirt-vm-status__link` class on the link element, which is what we really need in kubevirt/web-ui#189. (So instead of `.kubevirt-vm-status a` we can simply use `.kubevirt-vm-status__link`.)

PR #199, opened 2019-01-30
Add view-only boot order component to VM details dialog

pcbailey:
This PR is to add a boot order component to the VM details dialog. Editing functionality will be added in a separate PR in the future.

Depends on: https://github.com/kubevirt/web-ui/pull/207

### DIFF
--- a/sass/components/VmDetails/boot-order.scss
+++ b/sass/components/VmDetails/boot-order.scss
@@ -1,0 +1,4 @@
+.kubevirt-boot-order__list {
+  padding-left: 0;
+  list-style-position: inside;
+}
--- a/src/components/Details/BootOrder/BootOrder.js
+++ b/src/components/Details/BootOrder/BootOrder.js
@@ -1,0 +1,14 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const BootOrder = props => {
+  const { bootableDevices } = props;
+  const sortedBootableDevices = bootableDevices.sort((a, b) => a.bootOrder - b.bootOrder);
+  const listItems = sortedBootableDevices.map((dev, idx) => <li key={`${idx}-${dev.name}`}>{dev.name}</li>);
+
+  return <ol className="kubevirt-boot-order__list">{listItems}</ol>;
+};
+
+BootOrder.propTypes = {
+  bootableDevices: PropTypes.array.isRequired,
+};
--- a/src/components/Details/BootOrder/BootOrder.js
+++ b/src/components/Details/BootOrder/BootOrder.js
@@ -1,10 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-export const BootOrder = props => {
-  const { bootableDevices } = props;
-  const sortedBootableDevices = bootableDevices.sort((a, b) => a.bootOrder - b.bootOrder);
-  const listItems = sortedBootableDevices.map((dev, idx) => <li key={`${idx}-${dev.name}`}>{dev.name}</li>);
+export const BootOrder = ({ bootableDevices }) => {
+  const listItems = bootableDevices.map((dev, idx) => <li key={`${idx}-${dev.value.name}`}>{dev.value.name}</li>);
 
   return <ol className="kubevirt-boot-order__list">{listItems}</ol>;
 };
--- a/src/components/Details/VmDetails/VmDetails.js
+++ b/src/components/Details/VmDetails/VmDetails.js
@@ -30,7 +30,7 @@ import { Description } from '../Description';
 import { Loading } from '../../Loading';
 import { DESCRIPTION_KEY, FLAVOR_KEY } from '../common/constants';
 import { BootOrder } from '../BootOrder';
-import { getBootableDevices } from '../../../k8s/vmBuilder';
+import { getBootableDevicesInOrder } from '../../../k8s/vmBuilder';
 
 export class VmDetails extends React.Component {
   constructor(props) {
@@ -149,7 +149,7 @@ export class VmDetails extends React.Component {
     const fqdn = vmIsOff || !hostName ? DASHES : hostName;
     const template = getVmTemplate(vm);
     const id = getId(vm);
-    const bootableDevices = getBootableDevices(vm);
+    const sortedBootableDevices = getBootableDevicesInOrder(vm);
     const editButton = (
       <Fragment>
         {!vmIsOff && (
@@ -245,7 +245,9 @@ export class VmDetails extends React.Component {
                   <dd>{PodResourceLink ? <PodResourceLink /> : DASHES}</dd>
 
                   <dt>Boot Order</dt>
-                  <dd>{bootableDevices.length > 0 ? <BootOrder bootableDevices={bootableDevices} /> : DASHES}</dd>
+                  <dd>
+                    {sortedBootableDevices.length > 0 ? <BootOrder bootableDevices={sortedBootableDevices} /> : DASHES}
+                  </dd>
                 </dl>
               </Col>
 
--- a/src/components/Details/VmDetails/VmDetails.js
+++ b/src/components/Details/VmDetails/VmDetails.js
@@ -29,6 +29,8 @@ import { Flavor } from '../Flavor';
 import { Description } from '../Description';
 import { Loading } from '../../Loading';
 import { DESCRIPTION_KEY, FLAVOR_KEY } from '../common/constants';
+import { BootOrder } from '../BootOrder';
+import { getBootableDevices } from '../../../k8s/vmBuilder';
 
 export class VmDetails extends React.Component {
   constructor(props) {
@@ -147,7 +149,7 @@ export class VmDetails extends React.Component {
     const fqdn = vmIsOff || !hostName ? DASHES : hostName;
     const template = getVmTemplate(vm);
     const id = getId(vm);
-
+    const bootableDevices = getBootableDevices(vm);
     const editButton = (
       <Fragment>
         {!vmIsOff && (
@@ -240,6 +242,10 @@ export class VmDetails extends React.Component {
 
                   <dt>Pod</dt>
                   <dd id={prefixedId(id, 'pod')}>{PodResourceLink ? <PodResourceLink /> : DASHES}</dd>
+                  <dd>{PodResourceLink ? <PodResourceLink /> : DASHES}</dd>
+
+                  <dt>Boot Order</dt>
+                  <dd>{bootableDevices.length > 0 ? <BootOrder bootableDevices={bootableDevices} /> : DASHES}</dd>
                 </dl>
               </Col>
 
--- a/src/components/Details/VmDetails/fixtures/VmDetails.fixture.js
+++ b/src/components/Details/VmDetails/fixtures/VmDetails.fixture.js
@@ -23,7 +23,27 @@ export const vmFixtures = {
   },
   runningVm: {
     metadata,
-    spec: { running: true },
+    spec: {
+      template: {
+        spec: {
+          domain: {
+            devices: {
+              rng: {},
+              interfaces: [
+                { bridge: {}, name: 'podNetworkName', bootOrder: 2 },
+                { bridge: {}, name: 'pxeNetworkName', bootOrder: 1 },
+              ],
+              disks: [
+                { disk: {}, name: 'disk-one', bootOrder: 3 },
+                { disk: {}, name: 'disk-two', bootOrder: 4 },
+                { disk: {}, name: 'disk-three', bootOrder: 5 },
+              ],
+            },
+          },
+        },
+      },
+      running: true,
+    },
     status: {
       ready: true,
       created: true,
--- a/src/components/Details/VmDetails/fixtures/VmDetails.fixture.js
+++ b/src/components/Details/VmDetails/fixtures/VmDetails.fixture.js
@@ -37,9 +37,18 @@ export const vmFixtures = {
                 { disk: {}, name: 'disk-one', bootOrder: 3 },
                 { disk: {}, name: 'disk-two', bootOrder: 4 },
                 { disk: {}, name: 'disk-three', bootOrder: 5 },
+                { disk: { bus: 'virtio' }, name: 'cloudinitdisk' },
               ],
             },
           },
+          volumes: [
+            {
+              cloudInitNoCloud: {
+                userData: '# configure default password\npassword: fedora\nchpasswd: { expire: False }',
+              },
+              name: 'cloudinitdisk',
+            },
+          ],
         },
       },
       running: true,
--- a/src/constants/index.js
+++ b/src/constants/index.js
@@ -17,6 +17,9 @@ export const KUBEVIRT_API_VERSION = 'kubevirt.io/v1alpha3';
 export const OS_LABEL = 'kubevirt.io/os';
 export const POD_NETWORK = 'Pod Networking';
 
+export const DEVICE_TYPE_INTERFACE = 'interfaces';
+export const DEVICE_TYPE_DISK = 'disks';
+
 export const PROVISION_SOURCE_PXE = 'PXE';
 export const PROVISION_SOURCE_CONTAINER = 'Container';
 export const PROVISION_SOURCE_URL = 'URL';
--- a/src/constants/index.js
+++ b/src/constants/index.js
@@ -17,8 +17,15 @@ export const KUBEVIRT_API_VERSION = 'kubevirt.io/v1alpha3';
 export const OS_LABEL = 'kubevirt.io/os';
 export const POD_NETWORK = 'Pod Networking';
 
-export const DEVICE_TYPE_INTERFACE = 'interfaces';
-export const DEVICE_TYPE_DISK = 'disks';
+export const DEVICE_TYPE_INTERFACE = 'interface';
+export const DEVICE_TYPE_DISK = 'disk';
+export const INTERFACE_PATH_KEY = 'interfaces';
+export const DISK_PATH_KEY = 'disks';
+
+export const deviceTypeToPathKey = {
+  [DEVICE_TYPE_INTERFACE]: INTERFACE_PATH_KEY,
+  [DEVICE_TYPE_DISK]: DISK_PATH_KEY,
+};
 
 export const PROVISION_SOURCE_PXE = 'PXE';
 export const PROVISION_SOURCE_CONTAINER = 'Container';
--- a/src/k8s/request.js
+++ b/src/k8s/request.js
@@ -99,6 +99,7 @@ import {
   addVolume,
   removeDisksAndVolumes,
   getDataVolumeTemplateSpec,
+  sequentializeBootOrderIndexes,
 } from './vmBuilder';
 
 import { getImportProviderSecretObject } from '../components/Wizard/CreateVmWizard/providers/vmwareProviderPod';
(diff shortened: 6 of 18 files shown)


src/k8s/request.js
```diff
@@ -362,6 +362,33 @@ const getDevices = vm => {
...
+  const disks = getDisks(vm);
+  const nics = getInterfaces(vm);
+  let largestIdx = -1;
+
+  disks.forEach(disk => {
+    const bootOrder = get(disk, 'bootOrder', -1);
+    largestIdx = bootOrder > largestIdx ? bootOrder : largestIdx;
+  });
+
+  nics.forEach(nic => {
```
suomiy: I would rather not set explicitly boot order for networks when creating VMs. Except the ones the user chooses. I know this concerns editing the VMs detail, so maybe we should have an option to discard some nics (or disks) from the boot order? Otherwise this behavior could have security ramifications.

suomiy: maybe there could be one component for boot order which could be used in VM details and in the create VM dialog? It could have its own step (something like BIOS?). So we would not be forced to set bootorder to all devices - just the ones user chooses. Thoughts?

pcbailey: > maybe there could be one component for boot order which could be used in VM details and in the create VM dialog? It could have its own step (something like BIOS?). So we would not be forced to set bootorder to all devices - just the ones user chooses. Thoughts?

+1... I like this idea. @jelkosz @rawagner What do you think?

rawagner: sounds good to me!

sass/components/VmDetails/boot-order.scss
```diff
@@ -0,0 +1,4 @@
+.kubevirt-boot-order__ol {
```
rawagner: I think that its incorrect to use html tag `ol` in css selector. Correct would be something like `.kubevirt-boot-order__list`

src/components/Details/BootOrder/BootOrder.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const BootOrder = props => {
+  const { bootableDevices } = props;
+  const sortedBootableDevices = bootableDevices.sort((a, b) => a.bootOrder - b.bootOrder);
+  const listItems = [];
```
rawagner: can be reduced to `listItems = sortedBootableDevices.map((dev, idx) => <li key={`${idx}-${dev.name}`}>{dev.name}</li>)`;

src/k8s/request.js
suomiy: After a discussion, it was decided that everything would be bootable. The boot order will be the order of devices in CreateDialog. Afterwards, it can be edited in VM detail with possibility to change or remove the bootOrder.

suomiy: Boot order should be set when creating VMs. Created relevant issue: #230

src/components/Details/BootOrder/BootOrder.js
pcbailey: Nice and clean! Done!

sass/components/VmDetails/boot-order.scss
pcbailey: Ah... I wasn't aware of that. Thanks! Done!

src/k8s/vmBuilder.js
```diff
@@ -282,3 +282,86 @@ export const getDataVolumeTemplateSpec = (storage, dvSource) => {
...
+
+  // assigned indexes start at two as the first index is assigned directly by the user
+  return largestIdx !== -1 ? largestIdx + 1 : BOOT_ORDER_SECOND;
+};
+
+export const sequentializeBootOrderIndexes = vm => {
+  const sortedBootableDevices = getBootableDevicesInOrder(vm);
+
+  let index;
+  for (index = 0; index <= sortedBootableDevices.length; index++) {
```
suomiy: .forEach((device, index) => {}) is more readable IMO

src/k8s/vmBuilder.js
```diff
@@ -282,3 +282,86 @@ export const getDataVolumeTemplateSpec = (storage, dvSource) => {
...
+  });
+
+  nics.forEach(nic => {
+    const bootOrder = get(nic, 'bootOrder', null);
+    if (bootOrder) {
+      bootableDevices.push(nic);
+    }
+  });
+
+  return bootableDevices;
```
suomiy: please use shorthands 

`return [..nics.filter(predicate), ...disks.filter(predicate)]`

src/k8s/vmBuilder.js
```diff
@@ -282,3 +282,86 @@ export const getDataVolumeTemplateSpec = (storage, dvSource) => {
...
+};
+
+export const getBootableDevicesInOrder = vm => getBootableDevices(vm).sort((a, b) => a.bootOrder - b.bootOrder);
+
+export const isNic = device => !!get(device, 'bridge', null);
```
suomiy: I do not think this is good enough to identify disks/nics. Everything except name is optional (https://kubevirt.io/api-reference/master/definitions.html#_v1_interface). Please restructure the code so this is not needed or make a wrapper object.

src/k8s/vmBuilder.js
```diff
@@ -282,3 +282,86 @@ export const getDataVolumeTemplateSpec = (storage, dvSource) => {
...
+
+export const isDisk = device => !!get(device, 'disk', null);
+
+export const updateBootOrder = (vm, device, bootOrder) => {
+  if (isNic(device)) {
+    getInterfaces(vm).forEach(nic => {
```
suomiy: why do we have to find that device again anyway? isn't nic and device the same exact object?

src/k8s/request.js
```diff
@@ -193,6 +194,8 @@ export const createVm = async (k8sCreate, templates, basicSettings, networks, st
 
   const vm = selectVm(processedTemplate.objects);
 
+  sequentializeBootOrderIndexes(vm);
```
suomiy: I think this should be also applied for createVmTemplate

src/k8s/vmBuilder.js
```diff
@@ -282,3 +282,86 @@ export const getDataVolumeTemplateSpec = (storage, dvSource) => {
...
+  const disks = getDisks(vm);
+  const nics = getInterfaces(vm);
+  let largestIdx = -1;
+
+  disks.forEach(disk => {
+    const bootOrder = get(disk, 'bootOrder', -1);
+    largestIdx = bootOrder > largestIdx ? bootOrder : largestIdx;
+  });
+
+  nics.forEach(nic => {
```
suomiy: could be also written as `Math.max(...nics.map())`

src/k8s/vmBuilder.js
```diff
@@ -18,9 +18,9 @@ export const addDisk = (vm, defaultDisk, storage, getSetting) => {
   };
   if (storage.isBootable) {
     const imageSource = getSetting(PROVISION_SOURCE_TYPE_KEY);
-    diskSpec.bootOrder = imageSource === PROVISION_SOURCE_PXE ? BOOT_ORDER_SECOND : BOOT_ORDER_FIRST;
+    diskSpec.bootOrder = imageSource === PROVISION_SOURCE_PXE ? assignBootOrderIndex(vm) : BOOT_ORDER_FIRST;
   } else {
-    delete diskSpec.bootOrder;
```
suomiy: I think this should not include cloud init disks.

src/utils/patches.js
```diff
@@ -387,3 +392,46 @@ export const addPrefixToPatch = (prefix, patch) => ({
   ...patch,
   path: `${prefix}${patch.path}`,
 });
+
+const getDeviceType = device => {
+  let type;
+  if (isNic(device)) {
```
suomiy: again, we should already know here what kind of device we have

suomiy: I guess easiest way is to wrap the disks/nic in a object with type attribute.

src/utils/tests/patches.test.js
```diff
@@ -378,34 +380,39 @@ describe('utils.js tests', () => {
...
     const vmWithoutNetworks = getVM(false);
 
-    let patch = getAddNicPatch(vmWithoutNetworks, nic);
+    const patch = getAddNicPatch(vmWithoutNetworks, nic);
     expect(patch).toHaveLength(2);
     comparePatch(patch[0], '/spec/template/spec/domain/devices/interfaces/0', intface);
     comparePatch(patch[1], '/spec/template/spec/networks', [network]);
-
+  });
+  it('AddNicPatch - VM without interfaces', () => {
```
suomiy: +1 for splitting the tests

src/components/Details/VmDetails/VmDetails.js
```diff
@@ -231,6 +234,9 @@ export class VmDetails extends React.Component {
 
                   <dt>Pod</dt>
                   <dd>{PodResourceLink ? <PodResourceLink /> : DASHES}</dd>
+
+                  <dt>Boot Order</dt>
```
suomiy: this could make a sense also for VmTemplateDetails

src/k8s/vmBuilder.js
pcbailey: Done!

src/k8s/vmBuilder.js
pcbailey: Done!

src/k8s/vmBuilder.js
pcbailey: Done!

src/k8s/vmBuilder.js
pcbailey: This was me being stupid. Thanks for making me realize it. It allowed me to clean things up significantly.

src/utils/patches.js
pcbailey: Done!

src/utils/tests/patches.test.js
pcbailey: =)

src/components/Details/VmDetails/VmDetails.js
pcbailey: Agreed! Done!

src/constants/index.js
```diff
@@ -17,6 +17,9 @@ export const KUBEVIRT_API_VERSION = 'kubevirt.io/v1alpha3';
 export const OS_LABEL = 'kubevirt.io/os';
 export const POD_NETWORK = 'Pod Networking';
 
+export const DEVICE_TYPE_INTERFACE = 'interfaces';
```
suomiy: I know this is just an id but does it have to be plural?

src/k8s/vmBuilder.js
```diff
@@ -282,3 +289,43 @@ export const getDataVolumeTemplateSpec = (storage, dvSource) => {
...
+  }
+  return bootOrder;
+};
+
+const getNextAvailableBootOrderIndex = vm => {
+  const sortedBootableDevices = getBootableDevicesInOrder(vm);
+  const numBootableDevices = sortedBootableDevices.length;
+
+  // assigned indexes start at two as the first index is assigned directly by the user
+  return numBootableDevices > 0 ? sortedBootableDevices[numBootableDevices - 1].value.bootOrder + 1 : BOOT_ORDER_SECOND;
```
suomiy: can we return 0 instead of BOOT_ORDER_SECOND ? and do this when we use it : 

`getNextAvailableBootOrderIndex(vm) || BOOT_ORDER_SECOND`

this way, the function will contain less magic

src/utils/patches.js
```diff
@@ -387,3 +392,50 @@ export const addPrefixToPatch = (prefix, patch) => ({
   ...patch,
   path: `${prefix}${patch.path}`,
 });
+
+const getWrappedDeviceType = device => {
+  let type;
+  if (isNic(device)) {
+    type = DEVICE_TYPE_INTERFACE;
+  } else if (isDisk(device)) {
```
suomiy: I know this was needed historically but `isNic` and `isDisk` are not needed anymore. You can just call `get(device, 'type')` instead of this method,

src/utils/patches.js
```diff
@@ -387,3 +392,50 @@ export const addPrefixToPatch = (prefix, patch) => ({
...
+      const deviceType = getWrappedDeviceType(device);
+      const currDevIndex = getDeviceIndex(vm, device.value.name, deviceType);
+      // Account for index change once device is removed
+      const deviceIndex =
+        deviceType === removedDeviceType && currDevIndex > removedDeviceIndex ? currDevIndex - 1 : currDevIndex;
+
+      if (deviceIndex !== -1 && deviceType) {
+        const patch = {
+          op: 'replace',
+          path: `/spec/template/spec/domain/devices/${deviceType}/${deviceIndex}/bootOrder`,
```
suomiy: Oh! I see now, why is that needed. I think it would be probably better to decouple the type from the path fragment. I would rather use the singular for type and then do the mapping of the type to the path fragment. Quick and readable way is to use object for mappings for example.

this way, the functionality will not break when somebody decides to refactor the type

src/constants/index.js
suomiy: please look at the comment in `patches.js`

src/utils/tests/patches.test.js
```diff
@@ -151,6 +152,7 @@ const podNic = {
 
 const intface = {
```
suomiy: btw, does it have to be called intface ?

src/k8s/vmBuilder.js
pcbailey: Done!

src/constants/index.js
pcbailey: Agreed and done!

src/utils/patches.js
pcbailey: Good point. Done!

src/utils/patches.js
pcbailey: Agreed and done! Could you please check the way I implemented your suggestion to make sure this is what you were thinking?

src/utils/tests/patches.test.js
pcbailey: I don't believe I named this variable, but I'm pretty sure the reason they went with "intface" instead of "interface" is because "interface" is a reserved keyword in JavaScript. I typically use "iface" when shortening it, but it's not much better. Any suggestions?

src/k8s/vmBuilder.js
pcbailey: I think it makes more sense as it's currently written. The function is just expected to return the next available index, so the caller wouldn't expect to have to handle a special case like that when using it. I just reused the existing constant to avoid having a magic number and to increase readability. 

I tweaked the last line a small bit to make it a bit more understandable at first glance. Thoughts?

src/utils/patches.js
suomiy: yup, looks good! 

Although I would rather name it DEVICE_TYPE_INTERFACE or something like that. INTERFACE/DISK seems a bit generic.

src/k8s/vmBuilder.js
suomiy: ok, your call :) 

We have just one caller of the method and I would rather move the hacky logic higher up the chain. I was more focused on the method's name (`getNextAvailableBootOrderIndex`) which doesn't adhere to the reality. 
Nevertheless both are valid approaches.

src/utils/tests/patches.test.js
suomiy: I see, I forgot about that. You can leave it then, if you want..

pcbailey: I'll update it in the future if I think of a better alternative.

src/utils/patches.js
pcbailey: Agreed and done! =)

src/k8s/vmBuilder.js
pcbailey: Understood. I think the name is valid, though. All the hacky logic does is keep the method from returning an index of 1, which isn't technically an available index since it's assigned explicitly. I think I'll leave it as-is for the time being. We can always change it later if it becomes problematic.

suomiy: ok

src/k8s/tests/request.test.js
```diff
@@ -87,9 +87,9 @@ const testDataVolumeStorage = (
 
 const testStorage = (vm, storageIndex, bootOrder, expectedName) => {
   expect(vm.spec.template.spec.domain.devices.disks[storageIndex].name).toBe(expectedName);
-  expect(vm.spec.template.spec.domain.devices.disks[storageIndex].bootOrder).toBe(
-    bootOrder !== -1 ? bootOrder : undefined
-  );
+  vm.spec.template.spec.domain.devices.disks.forEach(disk => {
```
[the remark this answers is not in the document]
pcbailey: Done!

src/components/Details/VmDetails/fixtures/VmDetails.fixture.js
```diff
@@ -23,7 +23,27 @@ export const vmFixtures = {
...
+        spec: {
+          domain: {
+            devices: {
+              rng: {},
+              interfaces: [
+                { bridge: {}, name: 'podNetworkName', bootOrder: 2 },
+                { bridge: {}, name: 'pxeNetworkName', bootOrder: 1 },
+              ],
+              disks: [
+                { disk: {}, name: 'disk-one', bootOrder: 3 },
```
suomiy: can you add a cloud init to the vm ? so that we can see the snapshot does not contain it

src/utils/tests/patches.test.js
```diff
@@ -436,3 +456,45 @@ describe('utils.js tests', () => {
...
+  const interfaces = [
+    { bridge: {}, name: 'nic-one', bootOrder: 4 },
+    { bridge: {}, name: 'nic-two', bootOrder: 5 },
+    { bridge: {}, name: 'nic-three', bootOrder: 6 },
+  ];
+
+  let vm;
+
+  beforeEach(() => {
+    vm = cloneDeep(getVM(false));
```
suomiy: doesn't this function already return brand new object?

src/utils/tests/patches.test.js
```diff
@@ -436,3 +456,45 @@ describe('utils.js tests', () => {
     comparePatch(patch[0], '/spec', { running: false });
   });
 });
+
+describe('getDeviceBootOrderPatch() tests', () => {
+  const disks = [
+    { disk: {}, name: 'disk-one', bootOrder: 1 },
+    { disk: {}, name: 'disk-two', bootOrder: 2 },
+    { disk: {}, name: 'disk-three', bootOrder: 3 },
```
suomiy: can you please add the non bootable cloud init here as well?

src/k8s/vmBuilder.js
```diff
@@ -18,14 +26,21 @@ export const addDisk = (vm, defaultDisk, storage, getSetting) => {
   };
   if (storage.isBootable) {
     const imageSource = getSetting(PROVISION_SOURCE_TYPE_KEY);
-    diskSpec.bootOrder = imageSource === PROVISION_SOURCE_PXE ? BOOT_ORDER_SECOND : BOOT_ORDER_FIRST;
+    diskSpec.bootOrder = imageSource === PROVISION_SOURCE_PXE ? assignBootOrderIndex(vm) : BOOT_ORDER_FIRST;
+  } else if (!isCloudInitDisk(vm, diskSpec)) {
```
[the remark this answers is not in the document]
pcbailey: You're definitely right. I should have caught that. Thanks! =)

src/components/Details/VmDetails/fixtures/VmDetails.fixture.js
pcbailey: Done!

src/utils/tests/patches.test.js
pcbailey: It does... lol. Fixed!

src/utils/tests/patches.test.js
pcbailey: Done!

src/components/Details/BootOrder/BootOrder.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const BootOrder = props => {
+  const { bootableDevices } = props;
```
mareklibra: Nitpicking: Rest of the code follows syntax:
```
const MyComponent = ({ bootableDevices }) => {};
```

pcbailey: Fixed and rebased. =)

PR #203, opened 2019-02-01
Move mocked templates to openshift namespace.

rawagner:
Use ProjectModel instead of NamespaceModel when requesting resources.

### DIFF
--- a/src/k8s/mock_templates/fedora28.mock.js
+++ b/src/k8s/mock_templates/fedora28.mock.js
@@ -3,7 +3,7 @@ export const fedora28 = {
   kind: 'Template',
   metadata: {
     name: 'fedora-generic',
-    namespace: 'default',
+    namespace: 'openshift',
     annotations: {
       'openshift.io/display-name': 'Fedora 23+ VM',
       description:
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -9,6 +9,7 @@ import {
   TEMPLATE_FLAVOR_LABEL,
 } from '../constants';
 import { NETWORK_TYPE_POD } from '../components/Wizard/CreateVmWizard/constants';
+import { NamespaceModel, ProjectModel } from '../models';
 
 export function prefixedId(idPrefix, id) {
   return idPrefix && id ? `${idPrefix}-${id}` : null;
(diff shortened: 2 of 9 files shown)


src/k8s/mock_templates/fedora28.mock.js
```diff
@@ -3,7 +3,7 @@ export const fedora28 = {
   kind: 'Template',
   metadata: {
     name: 'fedora-generic',
-    namespace: 'default',
+    namespace: 'openshift',
```
rawagner: common-templates are installed into openshift namespace so I want to use the same namespace here too.

src/utils/utils.js
```diff
@@ -400,12 +401,18 @@ export const getStartStopPatch = (vm, start) => {
...
   model,
-  { name, namespaced = true, namespace, isList = true, matchLabels, matchExpressions } = {
+  { name, namespaced = true, namespace, isList = true, matchLabels, matchExpressions, prop } = {
     namespaced: true,
     isList: true,
   }
 ) => {
-  const res = { kind: model.kind, namespaced, namespace, isList, prop: model.kind };
+  const res = {
+    kind: model === NamespaceModel ? ProjectModel.kind : model.kind,
```
rawagner: non-admin user cannot list all namespaces (it wont return only namespaces available to user but 403 forbidden). Instead we need to use ProjectModel which will return available projects/namespaces

src/utils/utils.js
```diff
@@ -400,12 +401,18 @@ export const getStartStopPatch = (vm, start) => {
...
+    namespaced,
+    namespace,
+    isList,
+    prop: prop || model.kind,
```
rawagner: adding possibility to define own prop value which comes handy in case we have two queries for the same kind.

suomiy: is this going to work even with a ProjectModel?

src/utils/utils.js
rawagner: Sure, comment added to code

src/utils/utils.js
rawagner: > wouldn't it be better to remove or at least limit the usage of getResource helper for now as discussed in [openshift/console#888](https://github.com/openshift/console/pull/888) ?

Yes, i guess we should but lets deal with it later. This PR is also going to be backported to 1.4 so its better to keep the changes minimal

rawagner: > is this going to work even with a ProjectModel?

Yes, it is. The way how `withResources` interacts with `firehose` requires to use the same `prop` as key in `resourceMap`. Example:
`templates: {
      resource: getResource(TemplateModel)
}`

then key `templates` is the same as `prop` from `getResources`

but if you have two queries for the same kind ..something like

`userTemplates: {
      resource: getResource(TemplateModel, {namespace: activeNamespace, prop: 'userTemplates', matchLabels: {[TEMPLATE_TYPE_LABEL]: TEMPLATE_TYPE_VM}}),
    },
    commonTemplates: {
      resource: getResource(VmTemplateModel, {namespace: 'openshift', prop: 'commonTemplates', matchLabels: {[TEMPLATE_TYPE_LABEL]: TEMPLATE_TYPE_BASE}}),
    }`

then you need to specify `prop` value otherwise both would end up with `templates` which would break stuff.

in case of ProjectModel, lets say you want to load namespaces
`namespaces: {
      resource: getResource(NamespaceModel),
}`
still works - key is `namespaces` and prop will be `namespaces`

if you want to load Projects
`projects: {
      resource: getResource(ProjectModel),
}`
again key is `projects` and prop will be `projects`

You can of course break it :
`projects: {
      resource: getResource(ProjectModel), {prop: 'somethingelse'},
}`
key is `projects` and prop will be `somethingelse` but i think we can live with that for now

suomiy: oki, makes sense

suomiy:
created an issue on getResource: https://github.com/kubevirt/web-ui-components/issues/204

PR #206, opened 2019-02-04
Set template namespace to VM namespace

rawagner:


mareklibra:
I'm not sure if I understand this change. When is this needed?
No matter the template is from, the VM should be created within actual namespace.

PR #209, opened 2019-02-07
Use dataVolumeTemplates instead of creating PVC/DV objects.

rawagner:


package.json
```diff
@@ -13,9 +13,9 @@
     "/dist"
   ],
   "scripts": {
-    "lint": "yarn lint:js:browser && yarn lint:js:node && yarn lint:sass",
-    "lint:js:browser": "eslint src --config config/eslint.browser.js",
-    "lint:js:node": "eslint config tools --config config/eslint.node.js",
+    "lint": "yarn lint:js:browser && yarn lint:js:node && yarn lint:sass --fix",
```
mareklibra: This change definitely needs to go to a separate PR.

Anyway, I am not happy with this approach. What about building at release time? We should not change sources there.

I think, we should add new `lint:fix` script handling that.

rawagner: I actually forgot to remove it, it shouldn't be here :) 

But I agree to add new lint:fix script and create a separate PR for that

rawagner: https://github.com/kubevirt/web-ui-components/issues/211

rawagner:
@mareklibra this PR is ready for review :)

src/components/Wizard/CreateVmWizard/fixtures/ResultTab.fixture.js
```diff
@@ -14,15 +14,15 @@ export default [
     name: 'Error',
     props: {
       success: false,
-      result: 'Failed',
+      result: ['Failed'],
```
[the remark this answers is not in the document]
rawagner: fixed

PR #210, opened 2019-02-07
V2V

mareklibra:


lizsurette:
Awesome to see this! @mareklibra Anywhere we could start to see how it's looking?

mareklibra:
@lizsurette , I will need to create mock data along writing tests. I'll let you know in a later comment.
Anyway, there's still discussion about the general architecture.

lizsurette:
> @lizsurette , I will need to create mock data along writing tests. I'll let you know in a later comment.
> Anyway, there's still discussion about the general architecture (incl. backend).

Sounds great! Nice to see it's a WIP at least :)

mareklibra:
Rebased. There was a lot of conflicts but after brief checking it seems to be still working.
Continuing with tests and other remainers.

src/components/Wizard/CreateVmWizard/providers/vmware.js
```diff
@@ -0,0 +1,104 @@
...
+    },
+    required: true,
+    isVisible: isNewVmwareInstanceSelected,
+    // validate: settings => validateVmwareConnection(settingsValue(settings, PROVIDER_VMWARE_CONNECTION)),
+    help: 'User password to be used for connection to a vCenter instance.',
+  },
+  [PROVIDER_VMWARE_USER_PWD_REMEMBER_KEY]: {
+    id: 'vcenter-remember-credentials',
+    title: 'Remember vCenter credentials',
+    required: true,
```
rawagner: I does not make sense to have required checkbox :) If it really needs to be checked, then it should be just a label that informs user that the credentials will be remembered.

src/components/Wizard/CreateVmWizard/providers/vmware.js
```diff
@@ -0,0 +1,104 @@
...
+    type: 'checkbox',
+    isVisible: isNewVmwareInstanceSelected,
+    help: 'If checked, new secret keeping connection details will be created for later use.',
```
rawagner: help is not visible for checkboxes

src/components/Wizard/CreateVmWizard/providers/vmware.js
```diff
@@ -0,0 +1,104 @@
...
+    id: 'vcenter-username',
+    title: 'vCenter User Name',
+    required: true,
+    isVisible: isNewVmwareInstanceSelected,
+    help: 'User name to be used for connection to a vCenter instance.',
+  },
+  [PROVIDER_VMWARE_USER_PWD_AND_CHECK_KEY]: {
+    id: 'vcenter-userpwd',
+    title: 'vCenter Password',
+    type: 'custom',
```
rawagner: you can reuse form type constants from `Form/constants`

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
```diff
@@ -0,0 +1,182 @@
...
+  PROVIDER_VMWARE_USER_PWD_KEY,
+  PROVIDER_VMWARE_CONNECTION,
+} from '../constants';
+import { CONNECT_TO_NEW_INSTANCE } from '../strings';
+
+export const onVmwareCheckConnection = async (basicSettings, onChange, k8sCreate) => {
+  // Note: any changes to the dialog since issuing the Check-button action till it's finish will be lost due to tight binding of the onFormChange to basicSettings set at promise creation
+  onChange({ status: PROVIDER_STATUS_CONNECTING });
+
+  const url = get(basicSettings[PROVIDER_VMWARE_URL_KEY], 'value');
```
rawagner: `settingsValue` - in the future, we should rename it to `formValue` and probably move to `FormFactory`

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
```diff
@@ -0,0 +1,182 @@
...
+        isTemporary: true, // do not list this temporary secret in the dropdown box
+      })
+    );
+
+    const v2vVmware = await k8sCreate(
+      V2VVMwareModel,
+      getV2VVMwareObject({
+        name: `check-${getDefaultSecretName({ url, username })}-`,
+        namespace,
+        connectionSecretName: secret.metadata.name,
```
rawagner: getName(secret)

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
```diff
@@ -0,0 +1,182 @@
...
+        isTemporary: true, // remove this object automatically (by controller)
+      })
+    );
+
+    onChange({ V2VVmwareName: v2vVmware.metadata.name, status: PROVIDER_STATUS_CONNECTING }); // still "connecting" here, let content in the "status" of the CR decide otherwise (set by controller)
```
rawagner: getName

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
```diff
@@ -0,0 +1,182 @@
...
+          },
+        },
+        PROVIDER_VMWARE_USER_PWD_AND_CHECK_KEY,
+        formValid
+      );
+    }, 100);
+    return;
+  }
+
+  const namespace = get(prevBasicSettings, [NAMESPACE_KEY, 'value']);
```
rawagner: settingsValue

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
```diff
@@ -0,0 +1,182 @@
...
+  k8sGet,
+  k8sPatch,
+  valueValidationPair,
+  key,
+  formValid,
+  prevBasicSettings,
+  onFormChange
+) => {
+  const { value } = valueValidationPair; // name of VM to be imported
+  const namespace = get(prevBasicSettings, [NAMESPACE_KEY, 'value']);
```
rawagner: settingsValue

src/k8s/tests/request.test.js
```diff
@@ -516,6 +526,13 @@ describe('request.js - metadata', () => {
       );
       return results;
     }));
+  it('Import Secret is created', () =>
```
rawagner: could be converted to async

src/components/Wizard/CreateVmWizard/providers/index.js
```diff
@@ -0,0 +1,44 @@
...
+    required: true,
+    onChange: (...opts) => onProviderChanged(k8sCreate, k8sGet, ...opts),
+    isVisible: basicVmSettings => isImageSourceType(basicVmSettings, PROVISION_SOURCE_IMPORT),
+    help: getProviderHelp(basicSettings),
+  },
+  ...getVMWareSection(basicSettings, WithResources, k8sCreate, k8sGet, k8sPatch),
+});
+
+const onProviderChanged = (k8sCreate, k8sGet, valueValidationPair, key, formValid, prevBasicSettings) => {
+  startV2VVMWareController({ k8sCreate, k8sGet, namespace: get(prevBasicSettings[NAMESPACE_KEY], 'value') });
```
rawagner: if I do not select namespace before `startV2VVMWareController` is called, it will fail. Latest version of Create VM Wizard Design https://docs.google.com/document/d/1ks4H-ImD-r0s5eV4FsP_73kRdAP6sKJTRbj8nClpyp0/edit#heading=h.o4npgyz4fz5s guides user also by disabling fields we are not 'ready' to be selected. Ie when I open wizard, only name/description/namespace should be enabled. Once I fill those, it will enable Provision Source and so on. We should implement it soon.

src/components/Wizard/CreateVmWizard/providers/v2vvmwareController.js
```diff
@@ -0,0 +1,169 @@
...
+const rejectLogger = title => reason => console.warn(`Failed to create ${title}, reason`, reason);
+
+// The controller is namespace-scoped, especially due to security reasons
+// Let's make sure its started within the desired namespace (which is not by default).
+// The V2VVmware CRD is expected to be already created within the cluster (by Web UI installation)
+export const startV2VVMWareController = async ({ k8sCreate, k8sGet, namespace }) => {
+  try {
+    await k8sGet(DeploymentModel, V2VVMWARE_DEPLOYMENT_NAME, namespace);
+  } catch {
+    // Deployment does not exist
```
rawagner: it may also mean that user does not have permissions to get Deployments

src/components/Wizard/CreateVmWizard/providers/VMWarePasswordAndCheck.js
```diff
@@ -0,0 +1,212 @@
...
+          onClick={() => {
+            onCheckConnection(newValue =>
+              onChange({
+                [PROVIDER_VMWARE_USER_PWD_KEY]: pwdValue,
+                [PROVIDER_VMWARE_CONNECTION]: newValue,
+              })
+            );
+          }}
+        >
+          Check
```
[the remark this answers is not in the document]
mareklibra: This is per-design, the button is small for longer text.
What about keeping the layout as it is designed atm but changing the button text to `Login`?

WDYT, @lizsurette ? 
The question is about text/position of the `Check` button in the "Create New Vmware Connection" sub-part of the dialog.

Once the user hits the button, there's attempt to connect to the VMWare which is immediately followed by retrieving of the VMS list. The user watches progress/failure as designed (meens text/alert bellow the button).
In case of success, the next dropdown is populated with the names of VMs.

src/components/Wizard/CreateVmWizard/providers/VCenterVms.js
```diff
@@ -0,0 +1,76 @@
...
+
+  if (!v2vvmwareName) {
+    return <Dropdown id={id} value={PROVIDER_SELECT_VM} disabled />;
+  }
+
+  const resourceMap = {
+    v2vvmware: {
+      resource: getResource(V2VVMwareModel, {
+        name: v2vvmwareName,
+        namespace: get(basicSettings[NAMESPACE_KEY], 'value'),
```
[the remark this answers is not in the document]
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/VCenterInstances.js
```diff
@@ -0,0 +1,55 @@
...
+
+const areResourcesLoaded = resources => !!resources;
+
+const VCenterInstances = ({ onChange, id, value, extraProps }) => {
+  const { WithResources, basicSettings } = extraProps;
+
+  const resourceMap = {
+    vCenterSecrets: {
+      resource: getResource(SecretModel, {
+        namespace: basicSettings.namespace ? basicSettings.namespace.value : undefined,
```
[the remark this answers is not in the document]
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/VCenterInstances.js
```diff
@@ -0,0 +1,55 @@
...
+import { getResource } from '../../../../utils';
+import { SecretModel } from '../../../../models';
+import { VCENTER_TEMPORARY_LABEL, VCENTER_TYPE_LABEL } from '../../../../constants';
+import { Dropdown } from '../../../Form';
+
+import { CONNECT_TO_NEW_INSTANCE } from '../strings';
+
+const getVCenterInstanceSecrets = vCenterSecrets => {
+  vCenterSecrets = vCenterSecrets || [];
+  return [CONNECT_TO_NEW_INSTANCE, ...vCenterSecrets.map(secret => get(secret, 'metadata.name'))];
```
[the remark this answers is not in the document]
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/vmware.js
mareklibra: good catch :-)

src/components/Wizard/CreateVmWizard/providers/vmware.js
mareklibra: good point. Can I leave it as it is in case we add it? The behaviour of this checkbox might be confusing.

src/components/Wizard/CreateVmWizard/providers/vmware.js
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/vmware.js
rawagner: Sure, there is an issue https://github.com/kubevirt/web-ui-components/issues/238 for adding help to checkboxes

src/components/Wizard/CreateVmWizard/providers/index.js
rawagner: Follow up issue created https://github.com/kubevirt/web-ui-components/issues/239

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
mareklibra: done

src/k8s/tests/request.test.js
mareklibra: done on multiple other places as well

src/components/Wizard/CreateVmWizard/providers/v2vvmwareController.js
mareklibra: User will need to be informed i.e. using toast notification, recent logging to console is not enough.
Due to complexity of the flow, let's move that to a follow-up.

src/components/Wizard/CreateVmWizard/providers/index.js
mareklibra: Good catch, I'll fix that.

src/components/Wizard/CreateVmWizard/providers/VMWarePasswordAndCheck.js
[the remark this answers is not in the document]
lizsurette: @mareklibra I think "Login" works! I think @andybraren and I went back and forth on the right term here so I don't see any reason why that couldn't be our starting point and then we can get some feedback from users.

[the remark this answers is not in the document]
andybraren: I'd say our options are either: `Check`, `Verify`, `Validate`, `Log in`, or `Connect`.

I agree with @lizsurette that user feedback should determine what we end up using. My best guess is that they will be fine with `Check` because it is:

1) More common than `Verify`
2) More casual than `Validate`
3) More accurate than `Log in` (I feel like this is an unusual context to see "Log in")
4) Less technical than `Connect`

But I'm not passionate about this. I'm much more concerned about @rawagner's point that the button may not be obvious enough. I've mocked up two options.

option-1

Option 1 puts the Check button below the Password field as @rawagner suggested.

Behavior-wise, the button can stay enabled even while the fields are blank (one less dynamic thing to worry about) since any empty fields would be called out via error text. None of these fields or buttons should disappear after the check completes (the user may decide to change their credentials). The "Remember vCenter credentials" checkbox below the button becomes enabled if the credentials are valid, and unchecked/disabled if the user modifies any of the relevant fields after a successful check (same goes for the "VM to import" dropdown).

option-2

Option 2 is just making the Check button a primary blue color, with the same behavior described above.

I vote Option 1 (thoughts @lizsurette?)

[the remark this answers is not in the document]
lizsurette: > I vote Option 1 (thoughts @lizsurette?)

+1 to option 1. Thanks for thinking these options through @andybraren !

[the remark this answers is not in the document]
mareklibra: Thanks for the ideas. I've rearranged the password and Check-button based on option 1.
The Check button is disabled when the URL is not provided (username or pwd can be still empty).

I will implement the disabling of the checkbox in a later PR, probably new asynchronous flow will be needed for that. Please let me know if you think otherwise.

src/components/Wizard/CreateVmWizard/providers/v2vvmwareController.js
mareklibra: Follow-up issue: #240

mareklibra:
Recent screenshots:

01_connecting
--- 
02_failure

src/components/Wizard/CreateVmWizard/providers/vmwareProviderPod.js
```diff
@@ -0,0 +1,79 @@
...
+
+export const getDefaultSecretName = ({ username, url }) => {
+  if (!url) {
+    throw new Error('VMWare URL can not be empty.');
+  }
+
+  if (!url.startsWith('https://') && !url.startsWith('http://')) {
+    url = `https://${url}`;
+  }
+  const u = new URL(url || '');
```
suomiy: the url will never be falsy here, but nevertheless the `|| ''` would just throw an exception

src/components/Wizard/CreateVmWizard/providers/tests/vmwareProviderPod.test.js
```diff
@@ -0,0 +1,63 @@
...
+import { VCENTER_TEMPORARY_LABEL } from '../../../../../constants';
+
+describe('VMWareProviderPod()', () => {
+  it('returns default secret name correctly', () => {
+    expect(getDefaultSecretName({ username: 'a', url: 'url' })).toBe('url-a');
+    expect(getDefaultSecretName({ username: 'a', url: 'https://url' })).toBe('url-a');
+    expect(getDefaultSecretName({ username: 'a', url: 'http://url' })).toBe('url-a');
+    expect(getDefaultSecretName({ username: 'a', url: 'https://url.com/' })).toBe('url.com-a');
+    expect(getDefaultSecretName({ username: 'a', url: 'https://url.com/foo' })).toBe('url.com-a');
+    expect(getDefaultSecretName({ username: 'a@b', url: 'https://url.com/foo/bar' })).toBe('url.com-a-at-b');
```
suomiy: would be good to check the exceptions here as well

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
```diff
@@ -0,0 +1,187 @@
...
+  prevBasicSettings,
+  onFormChange
+) => {
+  // hint: onFormChange = (newValue, key, formValid) => {} to update basicSettings
+  const { value } = valueValidationPair;
+  const connectionSecretName = value;
+  if (!connectionSecretName || connectionSecretName === CONNECT_TO_NEW_INSTANCE) {
+    setTimeout(() => {
+      // let other events to be processed
+      onFormChange(
```
suomiy: why does this have to be postponed? Btw if by any chance the CreateVmDialog gets closed / unmounted, it will call setState on unmounted component.

src/components/Wizard/CreateVmWizard/providers/v2vvmwareController.js
```diff
@@ -0,0 +1,171 @@
...
+    apiVersion: ServiceAccountModel.apiVersion,
+    kind: ServiceAccountModel.kind,
+    metadata: {
+      name: V2VVMWARE_DEPLOYMENT_NAME,
+      namespace,
+    },
+  });
+
+const createRole = ({ k8sCreate, namespace }) =>
+  k8sCreate(RoleModel, {
```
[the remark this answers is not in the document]
mareklibra: I understand your point, it would be worth doing that in a more generic case.
But all these objects are for this very specific usecase only. And this `v2vvmwareController.js` handles that case only.

Honestly, I do not like having this deployment logic in UI.
But I did not find any other place where to put it.

From architecture perspective, good place can be:
- web-ui operator (but privileges issues)
- specific "namespace-preparation" code (does not exist ATM, would prepare a namespace for virtualization)

I've intentionally created the controller to be namespaced (not cluster-wide), mostly due to security reasons.

src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
```diff
@@ -368,6 +430,10 @@ BasicSettingsTab.defaultProps = {
 };
 
 BasicSettingsTab.propTypes = {
+  WithResources: PropTypes.func.isRequired,
```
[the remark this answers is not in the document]
mareklibra: Recurring issue, let's discuss that offline.
There is already plan to make such caller-context passing more friendly, but so far there was not time to implement that. Hopefully once the v2v effort is finished.

src/components/Wizard/CreateVmWizard/providers/VMWarePasswordAndCheck.js
```diff
@@ -0,0 +1,214 @@
...
+
+const LoadingData = () => (
+  <div className="kubevirt-create-vm-wizard__import-vmware-connection-status">
+    Connection successful. Loading data...
+    <Spinner loading size="sm" />
+  </div>
+);
+
+const ConnectionFailed = () => (
+  <div className="kubevirt-create-vm-wizard__import-vmware-connection-status">
```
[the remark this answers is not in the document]
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/VMWarePasswordAndCheck.js
```diff
@@ -0,0 +1,214 @@
...
+
+const CheckingCredentials = () => (
+  <div className="kubevirt-create-vm-wizard__import-vmware-connection-status">
+    Checking vCenter Credentials...
+    <Spinner loading size="sm" />
+  </div>
+);
+
+const LoadingData = () => (
+  <div className="kubevirt-create-vm-wizard__import-vmware-connection-status">
```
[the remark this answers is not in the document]
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/VMWarePasswordAndCheck.js
```diff
@@ -0,0 +1,214 @@
...
+};
+VMWareProviderStatusByPhase.defaultProps = {
+  phase: undefined,
+};
+VMWareProviderStatusByPhase.propTypes = {
+  phase: PropTypes.string,
+};
+
+// see onVmwareCheckConnection() for details
+const VMWareProviderStatus = ({ connValue, extraProps }) => {
```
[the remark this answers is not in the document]
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/v2vvmwareController.js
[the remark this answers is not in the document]
suomiy: Ok,I guess, if the meaning is to have them encapsulated in one file.

src/components/Wizard/CreateVmWizard/providers/tests/VMWarePasswordAndCheck.test.js
```diff
@@ -0,0 +1,66 @@
...
+import { basicSettingsImportVmwareNewConnection } from '../../../../../tests/forms_mocks/basicSettings.mock';
+
+const rendersProviderStatus = (providerStatus, phase) => {
+  const onChange = jest.fn();
+  const { value } = basicSettingsImportVmwareNewConnection[PROVIDER_VMWARE_USER_PWD_AND_CHECK_KEY];
+  value[PROVIDER_VMWARE_CONNECTION].providerStatus = providerStatus;
+
+  const extraProps = {
+    onCheckConnection: () => {},
+    WithResources: props => <WithResources {...props} tesOnlyPhase={phase} />,
```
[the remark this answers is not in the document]
mareklibra: The WithResources component is a mock. Mocks usually contain hacks to by-pass additional complex, not-tested, logic. IMO, this is the case.

src/components/Wizard/CreateVmWizard/providers/tests/v2vvmwareController.test.js
```diff
@@ -0,0 +1,20 @@
...
+describe('namespace-local kubevirt-vmware deployment', () => {
+  it('works', async () => {
+    const k8sCreate = jest.fn().mockReturnValue(Promise.resolve());
+    await startV2VVMWareController({
+      k8sCreate,
+      k8sGet: k8sGetFailing,
+      namespace: 'my-namespace',
+    });
+    expect(k8sCreate.mock.calls).toHaveLength(4);
+    expect(k8sCreate.mock.calls[0][0].kind).toBe(ServiceAccountModel.kind);
```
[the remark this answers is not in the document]
mareklibra: I think it's a matter of readability. The former syntax is more clear to me at the first glance.

src/components/Wizard/CreateVmWizard/providers/vmwareProviderPod.js
mareklibra: right, leftover, removing

src/components/Wizard/CreateVmWizard/providers/tests/vmwareProviderPod.test.js
mareklibra: good point, adding the test

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
mareklibra: It's about how the onChange/onFormeChange functions bubbles-through the event handling, the value set here is later replaced by something else. Rearranging with just `setTimeout(..., 0)` works fine as well.

src/components/Wizard/CreateVmWizard/providers/tests/VMWarePasswordAndCheck.test.js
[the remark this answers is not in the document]
suomiy: yes mocking is fine. I mean it would be easier to have components without WithResources so we wouldn't have to do this. WithResources should be just a slim wrapper which wouldn't need much testing.

src/components/Wizard/CreateVmWizard/providers/tests/v2vvmwareController.test.js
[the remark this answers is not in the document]
suomiy: ok

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
suomiy: I see, 0 looks much better now.

lizsurette:
> Recent screenshots:

Thanks for sharing these, @mareklibra ! It's looking great, well done. What do you think about using form inline error messages rather than the inline notification? Might look something like this instead:
screen shot 2019-02-25 at 8 52 17 pm

FYI @andybraren

mareklibra:
Thanks, @lizsurette , the inline notifications look better. I'll implement them in a folow-up, tracking: #242

PR #215, opened 2019-02-11
 add VmTemplateDetails + refactor components and tests

suomiy:


suomiy:
The design:
aa

Editing:
bb

src/components/Details/VmTemplateDetails/VmTemplateDetails.js
```diff
@@ -0,0 +1,231 @@
...
+
+  updateVmDetails = () => {
+    const { vmTemplate } = this.props;
+    const vm = selectVm(vmTemplate.objects);
+    const vmIndex = vmTemplate.objects.indexOf(vm);
+
+    this.setEditing(false);
+    const vmTemplatePatch = [];
+
+    const descriptionForm = get(this.state.form.description, 'value');
```
rawagner: you can use `settingsValue(this.state.form, DESCRIPTION_KEY)`

src/components/Details/VmTemplateDetails/VmTemplateDetails.js
```diff
@@ -0,0 +1,231 @@
...
+    const flavorForm = get(this.state.form.flavor, 'value');
```
rawagner: settingsValue

src/components/Details/VmTemplateDetails/VmTemplateDetails.js
```diff
@@ -0,0 +1,231 @@
...
+              <dt>Name</dt>
+              <dd>{vmTemplate.metadata.name}</dd>
+              <dt>Description</dt>
+              <dd>
+                <div className="kubevirt-vm-template-details__description">
+                  <Description
+                    editing={this.state.editing}
+                    updating={this.state.updating}
+                    LoadingComponent={LoadingComponent}
+                    formValues={get(this.state.form.description, 'value')}
```
rawagner: settingsValue

src/components/Details/VmTemplateDetails/VmTemplateDetails.js
```diff
@@ -0,0 +1,231 @@
...
+                  <dd>
+                    <Flavor
+                      flavor={getFlavor(vmTemplate) || CUSTOM_FLAVOR}
+                      vm={vm}
+                      editing={this.state.editing}
+                      updating={this.state.updating}
+                      LoadingComponent={LoadingComponent}
+                      onFormChange={(newValue, key, valid) => this.onFormChange('flavor', newValue, key, valid)}
+                      retrieveVmTemplate={() => retrieveVmTemplate(k8sGet, vmTemplate)}
+                      formValues={get(this.state.form.flavor, 'value')}
```
rawagner: settingsValue

rawagner:
@lizsurette WDYT about the design of the page ? I wonder about the `Source` part and showing URL inside. It seems a little messy to me.

src/components/Wizard/CreateVmWizard/fixtures/CreateVmWizard.fixture.js
```diff
@@ -1,7 +1,7 @@
 import { CreateVmWizard } from '../CreateVmWizard';
 import { ProcessedTemplatesModel } from '../../../../models';
 import { networkConfigs } from '../../../../tests/mocks/networkAttachmentDefinition';
-import { baseTemplates } from '../../../../tests/mocks/template';
```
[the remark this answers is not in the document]
suomiy: Base templates should not be semantically under the tests folder because they may be used in runtime.  It also breaks the build.

suomiy:
fixed

andybraren:
@suomiy Sorry I tend to see these late. 🙁

The overall design looks good to me - I think it includes all of the necessary fields and is consistent with other pages. I can't quite remember where the discussion landed, but "Virtual Machine Templates" might just become "Templates" in both the title and nav, making the blue badge just "T" instead of "VMT" as well.

Regarding the long URL, I can see that becoming a problem with even longer ones. If possible, I'd propose something like:

`https://www.mirror.domain.com/images/obscure-foldername/whatever/linux/centos/7.6/isos/CentOS-7-x86_64-Minimal-1810.iso`

Becomes:

`mirror.domain.com.../CentOS-7-x86_64-Minimal-1810.iso`

- Strip the protocol
- Strip any `www.` subdomains
- Ideally, strip all but the first subdomain (`mirror`), and domain name (I imagine this might be tricky to figure out, so I'm okay with not doing this)
- Replace the path with "..." but include the final part (probably the filename)

The beginning and very end of the URL with the filename seem like the most relevant/useful parts to show in this dashboard, and doing this will at least cut out most of the boring parts.

Maybe the "..." could be blue and clickable, and doing so would expand to show the full URL in case the user needs that?

rawagner:
> @suomiy Sorry I tend to see these late. 

no worries
> 
> The overall design looks good to me - I think it includes all of the necessary fields and is consistent with other pages. I can't quite remember where the discussion landed, but "Virtual Machine Templates" might just become "Templates" in both the title and nav, making the blue badge just "T" instead of "VMT" as well.

How would we distinguish between regular Templates and VM Templates ? Tectonic does not have a view for regular Templates yet but I believe they will add one.
> 
> Regarding the long URL, I can see that becoming a problem with even longer ones. If possible, I'd propose something like:
> 
> `https://www.mirror.domain.com/images/obscure-foldername/whatever/linux/centos/7.6/isos/CentOS-7-x86_64-Minimal-1810.iso`
> 
> Becomes:
> 
> `mirror.domain.com.../CentOS-7-x86_64-Minimal-1810.iso`
> 
> * Strip the protocol
> * Strip any `www.` subdomains
> * Ideally, strip all but the first subdomain (`mirror`), and domain name (I imagine this might be tricky to figure out, so I'm okay with not doing this)
> * Replace the path with "..." but include the final part (probably the filename)
> 
> The beginning and very end of the URL with the filename seem like the most relevant/useful parts to show in this dashboard, and doing this will at least cut out most of the boring parts.
> 
> Maybe the "..." could be blue and clickable, and doing so would expand to show the full URL in case the user needs that?

sounds good to me. Here is a followup issue https://github.com/kubevirt/web-ui-components/issues/217

PR #216, opened 2019-02-12
Improve VM Status - detect upload/clone and import datavolumes/pods

rawagner:
what is changed on VM list page:
1. instead of status `Importing` we show `Preparing Disks`
2. instead of `Import Error` we show `Disk Error`

in VM details:
1. We show `Preparing`/`Importing`/`Cloning`/`Uploading` `disk <diskName>`
2. In case of error we show `Importing`/`Cloning`/`Uploading` `disk <diskName> failed`

### DIFF
--- a/src/components/VmStatus/VmStatus.js
+++ b/src/components/VmStatus/VmStatus.js
@@ -7,21 +7,22 @@ import {
   VM_STATUS_STARTING,
   VM_STATUS_RUNNING,
   VM_STATUS_OFF,
-  VM_STATUS_IMPORTING,
   VM_STATUS_POD_ERROR,
   VM_STATUS_ERROR,
-  VM_STATUS_IMPORT_ERROR,
   VM_STATUS_MIGRATING,
-  CDI_KUBEVIRT_IO,
-  STORAGE_IMPORT_PVC_NAME,
+  VM_STATUS_DISKS_FAILED,
+  VM_STATUS_PREPARING_DISKS,
+  DATA_VOLUME_STATUS_PENDING,
+  DATA_VOLUME_STATUS_UPLOAD_IN_PROGRESS,
+  DATA_VOLUME_STATUS_CLONE_IN_PROGRESS,
+  DATA_VOLUME_STATUS_IMPORT_IN_PROGRESS,
 } from '../../constants';
 
 import { getSubPagePath } from '../../utils';
 import { PodModel, VirtualMachineModel } from '../../models';
 import { getVmStatusDetail } from './getVmStatus';
-import { getId } from '../../utils/selectors';
-
-const getAdditionalImportText = pod => ` (${pod.metadata.labels[`${CDI_KUBEVIRT_IO}/${STORAGE_IMPORT_PVC_NAME}`]})`;
+import { getNamespace, getName } from '../../utils/selectors';
+import { getVolumes } from '../../k8s/vmBuilder';
 
 const StateValue = ({ iconClass, children, linkTo, message }) => (
   <Fragment>
@@ -65,10 +66,52 @@ const StateStarting = ({ ...props }) => (
     Starting
   </StateValue>
 );
-const StateImporting = ({ additionalText, ...props }) => (
+
+const getDiskLinkTo = (pod, vm) =>
+  pod ? getSubPagePath(pod, PodModel, 'events') : getSubPagePath(vm, VirtualMachineModel, 'events');
+
+const StateDisksPreparing = ({ vm, statusDetail, verbose }) => {
+  const diskProps = {
+    linkTo: getDiskLinkTo(statusDetail.pod, vm),
+    diskName: statusDetail.diskName,
+    verbose,
+  };
+  switch (statusDetail.diskStatus) {
+    case DATA_VOLUME_STATUS_PENDING:
+      return <StatePreparingDisks action="Preparing" {...diskProps} />;
+    case DATA_VOLUME_STATUS_UPLOAD_IN_PROGRESS:
+      return <StatePreparingDisks action="Uploading" {...diskProps} />;
+    case DATA_VOLUME_STATUS_CLONE_IN_PROGRESS:
+      return <StatePreparingDisks action="Cloning" {...diskProps} />;
+    case DATA_VOLUME_STATUS_IMPORT_IN_PROGRESS:
+      return <StatePreparingDisks action="Importing" {...diskProps} />;
+    default:
+      return <StateUnknown />;
+  }
+};
+const StateDisksFailed = ({ vm, statusDetail, verbose }) => {
+  const message = verbose ? statusDetail.verboseMessage : statusDetail.message;
+  const child = verbose ? statusDetail.message : 'Disk Error';
+  return (
+    <StateError linkTo={getDiskLinkTo(statusDetail.pod, vm)} message={message}>
+      {child}
+    </StateError>
+  );
+};
+
+StateDisksFailed.defaultProps = {
+  verbose: false,
+};
+
+StateDisksFailed.propTypes = {
+  vm: PropTypes.object.isRequired,
+  statusDetail: PropTypes.object.isRequired,
+  verbose: PropTypes.bool,
+};
+
+const StatePreparingDisks = ({ diskName, verbose, action, ...props }) => (
   <StateValue iconClass="pficon pficon-import" {...props}>
-    Importing
-    {additionalText}
+    {verbose ? `${action} disk ${diskName}` : `Preparing Disks`}
   </StateValue>
 );
 const StateError = ({ children, ...props }) => (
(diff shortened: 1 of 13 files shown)


src/components/VmStatus/VmStatus.js
```diff
@@ -78,46 +131,47 @@ const StateError = ({ children, ...props }) => (
...
+      ));
     }
   }
-  return (
-    <div key={importerPods && importerPods.length > 0 ? getId(importerPods[0]) : '6d0c77-has-no-importer-pods'}>
-      <VmStatus {...props} />
-    </div>
-  );
+
+  return <VmStatus {...props} />;
```
suomiy: I think it would be a better practice to put a key here

src/components/VmStatus/VmStatus.js
```diff
@@ -78,46 +131,47 @@ const StateError = ({ children, ...props }) => (
...
   launcherPod: PropTypes.object,
-  importerPods: PropTypes.array,
+  cdiPods: PropTypes.array,
   migration: PropTypes.object,
+  dataVolumes: PropTypes.array,
 };
 
-export const VmStatus = ({ vm, launcherPod, importerPods, migration, verbose }) => {
-  const statusDetail = getVmStatusDetail(vm, launcherPod, importerPods, migration);
+const showVmStatus = (vm, launcherPod, statusDetail, verbose) => {
```
suomiy: this can be a component. it certainly behaves like one

src/components/VmStatus/VmStatus.js
```diff
@@ -127,13 +181,9 @@ export const VmStatus = ({ vm, launcherPod, importerPods, migration, verbose })
...
-    case VM_STATUS_IMPORTING:
-      return (
-        <StateImporting
-          linkTo={getSubPagePath(statusDetail.pod, PodModel, 'events')}
-          additionalText={verbose ? getAdditionalImportText(statusDetail.pod) : ''}
-        />
-      );
+    case VM_STATUS_PREPARING_DISKS:
+    case VM_STATUS_DISKS_FAILED:
+      return <StateDisks vm={vm} statusDetail={statusDetail} verbose={verbose} />;
```
suomiy: IMO, having two components would be better, it gets switched two times, here and inside the component (with the same meaning)

src/components/VmStatus/getVmStatus.js
```diff
@@ -156,17 +126,132 @@ const isWaitingForVmi = vm => {
...
+    if (failingContainer) {
+      return getContainerStatusReason(failingContainer);
+    }
+  }
+  return NOT_HANDLED;
+};
+
+export const isPreparingDisks = (vm, cdiPods, dataVolumes) => {
+  const diskStatuses = [];
+  if (dataVolumes) {
```
suomiy: I would prefer an early return here to stop the indentation

src/components/VmStatus/getVmStatus.js
```diff
@@ -156,17 +126,132 @@ const isWaitingForVmi = vm => {
...
+  cdiPods.find(pod => {
+    const podName = getName(pod);
+    const dataVolumeName = volume.dataVolume.name;
+    // upload dataVolume pod
+    if (podName === `cdi-upload-${dataVolumeName}`) {
+      return true;
+    }
+    // import dataVolume pod
+    const importerPodName = `importer-${dataVolumeName}-`;
+    if (podName.startsWith(importerPodName) && podName.length === importerPodName.length + 5) {
```
suomiy: why 5? Better to put some explanation and/or make a variable out of it

src/components/VmStatus/getVmStatus.js
```diff
@@ -156,17 +126,132 @@ const isWaitingForVmi = vm => {
   return NOT_HANDLED;
 };
 
-export const getVmStatusDetail = (vm, launcherPod, importerPods, migration) =>
-  isBeingMigrated(vm, migration) || // must be precceding isRunning() since vm.status.ready is true for a migrating VM
+const findCdiPod = (volume, cdiPods = []) =>
```
suomiy: wouldn't it be better to move this method to some util file?

src/components/VmStatus/getVmStatus.js
```diff
@@ -99,47 +109,7 @@ const isVmError = vm => {
...
-        };
-      }
-      return {
-        status: VM_STATUS_IMPORTING,
-        message: getNotRedyConditionMessage(pod),
-        pod,
-      };
-    });
-    const importErrorStatus = importerPodsStatuses.find(status => status.status === VM_STATUS_IMPORT_ERROR);
-    const message = importerPodsStatuses
```
suomiy: can we have this functionality preserved? I think it might be useful

src/components/VmStatus/VmStatus.js
rawagner: key added

src/components/VmStatus/VmStatus.js
rawagner: showVmStatus function is now VmStatusDetail component.

src/components/VmStatus/VmStatus.js
rawagner: is this resolved by making component out of showVmStatus ? or do you mean something else ?

src/components/VmStatus/getVmStatus.js
rawagner: added early return and also another one a few lines later

src/components/VmStatus/getVmStatus.js
rawagner: moved to `utils/selectors`

src/components/VmStatus/getVmStatus.js
rawagner: right, 5 is actually not really good. It was meant to be a suffix generated by kubernetes but It seems that length does not always have to be 5. Im looking at pod labels now, which is definitely better.

src/components/VmStatus/getVmStatus.js
rawagner: added messages back. See screenshots

screenshot from 2019-02-27 11-31-11
screenshot from 2019-02-27 11-33-15

matthewcarleton: I wonder if having "Anotherdisk import Failed" and "rootdisk import failed" Would be a better way of presenting these? A little more readable. 
Nitpick on design, but the tooltip is missing the arrow and the spacing seems odd for the first example.

src/components/VmStatus/getVmStatus.js
```diff
@@ -156,17 +131,138 @@ const isWaitingForVmi = vm => {
...
+    };
+    if (podError) {
+      diskStatuses.push({
+        ...diskStatus,
+        status: VM_STATUS_DISKS_FAILED,
+        message: getDiskErrorMessage(dVolume, volume.name),
+        verboseMessage: podError,
+      });
+    } else {
+      switch (get(dVolume, 'status.phase')) {
```
suomiy: can you please make this less verbose?

almost all the cases differ only by diskStatus string which could be resolved by function or object mappings

src/components/VmStatus/getVmStatus.js
```diff
@@ -156,17 +131,138 @@ const isWaitingForVmi = vm => {
...
+        default:
+          break;
+      }
+    }
+  });
+  if (diskStatuses.length === 1) {
+    return diskStatuses[0];
+  }
+  if (diskStatuses.length > 1) {
+    const failedDiskStatuses = diskStatuses.filter(ds => ds.status === VM_STATUS_DISKS_FAILED);
```
suomiy: shouldn't DATA_VOLUME_STATUS_FAILED statuses be also included?

src/components/VmStatus/getVmStatus.js
rawagner: > I wonder if having "Anotherdisk import Failed" and "rootdisk import failed" Would be a better way of presenting these? A little more readable.
+1, messages changed to \<disk_name\> import failed

> Nitpick on design, but the tooltip is missing the arrow and the spacing seems odd for the first example.

Spacing fixed, but the tooltip is actually link title and it is in the same style as other link titles

screenshot from 2019-03-01 14-26-26

suomiy: it was actually using pf-react tooltip in the past, but it was having problems with breaking long urls

matthewcarleton: Do you mean just using `title="text"` on the link? If so, we should really explore a different approach. The title attribute is not typically exposed in an accessible manner so it wouldn't be available to all users. I would think that a tooltip where a user could click to get further info would be most useful here.

rhrazdil:
@rawagner Any chance the vm status could have an id/class?

src/components/VmStatus/getVmStatus.js
rawagner: @matthewcarleton yes i meant just `title="text"`. Thanks for explanation,  I will explore the tooltip option.

src/components/VmStatus/VmStatus.js
rawagner: Ok, thanks, I get it. I split those two cases to two separate components

rawagner:
> @rawagner Any chance the vm status could have an id/class?

sure!

src/components/VmStatus/getVmStatus.js
rawagner: DATA_VOLUME_STATUS_FAILED is still picked up because im checking status field VM_STATUS_DISKS_FAILED which indicates that 'something' failed and that 'something' can be either data volume ( DATA_VOLUME_STATUS_FAILED ) or pod

src/components/VmStatus/getVmStatus.js
rawagner: I tried to make it less verbose. Please check again

rawagner:
> > @rawagner Any chance the vm status could have an id/class?
> 
> sure!

id added

src/components/VmStatus/getVmStatus.js
rawagner: Followup created https://github.com/kubevirt/web-ui-components/issues/254

matthewcarleton: great, thanks!

src/components/VmStatus/VmStatus.js
```diff
@@ -160,15 +200,38 @@ export const VmStatus = ({ vm, launcherPod, importerPods, migration, verbose })
...
+
+VmStatusDetail.defaultProps = {
+  verbose: false,
+  launcherPod: undefined,
+};
+
+export const VmStatus = ({ vm, launcherPod, cdiPods, migration, dataVolumes }) => {
+  const statusDetail = getVmStatusDetail(vm, launcherPod, cdiPods, migration, dataVolumes);
+  return (
+    <div id={`kubevirt-vm-status-${getNamespace(vm)}/${getName(vm)}`}>
```
suomiy: please use `utils/getId` function. Also I would rather avoid using `/` character inside an id (https://stackoverflow.com/questions/70579/what-are-valid-values-for-the-id-attribute-in-html).

Maybe you can refactor the `getId` function to use `_` instead of `-` to not have only hyphens in our ids. I guess you could also add a check for empty namespace while you are at it.

src/components/VmStatus/VmStatus.js
```diff
@@ -78,46 +121,53 @@ const StateError = ({ children, ...props }) => (
...
+    );
+
+    const statuses = vmDataVolumes.map(vmDv => getVmStatusDetail(vm, launcherPod, cdiPods, migration, [vmDv]));
+    const diskStatuses = statuses.filter(
+      status => status.status === VM_STATUS_PREPARING_DISKS || status.status === VM_STATUS_DISKS_FAILED
+    );
+
+    if (diskStatuses.length > 0) {
+      return (
+        <div id={`kubevirt-vm-statuses-${getNamespace(vm)}/${getName(vm)}`}>
```
suomiy: \+ here

src/components/VmStatus/VmStatus.js
```diff
@@ -78,46 +121,53 @@ const StateError = ({ children, ...props }) => (
...
+      );
     }
   }
-  return (
-    <div key={importerPods && importerPods.length > 0 ? getId(importerPods[0]) : '6d0c77-has-no-importer-pods'}>
-      <VmStatus {...props} />
-    </div>
-  );
+
+  return <VmStatus {...props} key={`${getName(vm)}-status`} />;
```
suomiy: can you please wrap this in a div with id to have a consistent layout?

PR #227, opened 2019-02-18
Form: enable password input-text components

mareklibra:


src/components/Form/Text.js
```diff
@@ -4,11 +4,11 @@ import { FormControl, noop } from 'patternfly-react';
 
 import { eventValueHandler } from './utils';
 
-export const Text = ({ id, value, disabled, defaultValue, onChange, onBlur }) => (
+export const Text = ({ id, value, disabled, defaultValue, onChange, onBlur, subtype }) => (
   <FormControl
     id={id}
     key={id}
-    type="text"
+    type={subtype}
```
suomiy: is there a reason to call it a subtype? I think passing just 'type' should be more clear.

mareklibra: I didn't want to mix with the web-ui-specific types (CUSTOM, TEXT, DROPDOWN, etc).
But you are right, it would hide the reality, so changing subtype to type.

mareklibra:
Done

mareklibra:
More tests added

PR #229, opened 2019-02-19
shorten URL in VmTemplateDetail and refactor TemplateSource

### LINKED ISSUE #217: Improve URL rendering in VM details page
@andybraren

> 
> Regarding the long URL, I can see that becoming a problem with even longer ones. If possible, I'd propose something like:
> 
> `https://www.mirror.domain.com/images/obscure-foldername/whatever/linux/centos/7.6/isos/CentOS-7-x86_64-Minimal-1810.iso`
> 
> Becomes:
> 
> `mirror.domain.com.../CentOS-7-x86_64-Minimal-1810.iso`
> 
> * Strip the protocol
> * Strip any `www.` subdomains
> * Ideally, strip all but the first subdomain (`mirror`), and domain name (I imagine this might be tricky to figure out, so I'm okay with not doing this)
> * Replace the path with "..." but include the final part (probably the filename)
> 
> The beginning and very end of the URL with the filename seem like the most relevant/useful parts to show in this dashboard, and doing this will at least cut out most of the boring parts.
> 
> Maybe the "..." could be blue and clickable, and doing so would expand to show the full URL in case the user needs that?

suomiy:
Fixes #217 

Shortens url. The href and tooltip contain the original url.

#### Transformations:

base url: http://merlin.fit.vutbr.cz/mirrors/centos/7.6.1810/isos/x86_64/CentOS-7-x86_64-Minimal-1810.iso

a

base url: http://fit.vutbr.cz/mirrors/centos/7.6.1810/isos/x86_64/CentOS-7-x86_64-Minimal-1810.iso

b

base url: http://merlin.fit.vutbr.cz/CentOS-7-x86_64-Minimal-1810.iso?a=5#here

e

One disadvantage is that the elipsis can appear 3 times:  [...sub.hostname.end/.../final-part...]() 
 The last bit for parameters. 

@andybraren What do you think?

### DIFF
--- a/src/components/Details/Url/Url.js
+++ b/src/components/Details/Url/Url.js
@@ -1,0 +1,66 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import { parseUrl } from '../../../utils';
+
+const ELIPSIS = '…';
+
+const elipsizeLeft = word => `${ELIPSIS}${word}`;
+
+const resolveOrigin = ({ hostname, origin, port }, maxHostnameParts) => {
+  const hostnameParts = hostname.split('.');
+  if (hostnameParts.length <= maxHostnameParts) {
+    return origin;
+  }
+
+  const resolvedHostname = hostnameParts.slice(hostnameParts.length - maxHostnameParts).join('.');
+  const resolvedPort = port ? `:${port}` : '';
+
+  return `${elipsizeLeft(resolvedHostname)}${resolvedPort}`;
+};
+
+const resolvePathname = ({ pathname }, maxPathnameParts) => {
+  const pathnameParts = pathname.split('/').filter(part => part);
+  if (pathnameParts.length <= maxPathnameParts) {
+    return pathname;
+  }
+
+  const resolvedPathname = pathnameParts.slice(pathnameParts.length - maxPathnameParts).join('/');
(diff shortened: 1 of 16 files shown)


andybraren:
Nice! I think the first "..." for the http/https:// part and the last "..." for the parameters can both be removed entirely. Compared to the base URL and filename they're unimportant and don't need to be hinted at.

suomiy:
I am ok with removing the parameters but IMO the first part is not so unimportant. I think it might be beneficial to user to hint that there are other subdomains and the protocol.

andybraren:
Ah you’re right, I agree that hinting at additional subdomains is important and the “...” should be kept in that case.

Are protocols other than http/https (ftp/sftp/smb) supported? If yes then the protocol is probably important enough to show at all times, despite the increased length. If no then I would still lean toward being able to hide http/https entirely without a “...”, but I defer to your judgment in this context.

So my proposal:

https://...sub.domain.com/filename.iso (additional subdomains stripped, but protocol is kept because we decide it’s important)

sub.domain.com/.../filename.iso (we decide the protocol is not important)

...sub.domain.com/.../filename.iso (we decide the protocol is not important, so the starting “...” only represents additional subdomains)

Another consideration - are there situations where port numbers would be included in the URL? If so, that number should probably be shown too.

PR #231, opened 2019-02-20
fix workload help message

suomiy:
https://bugzilla.redhat.com/show_bug.cgi?id=1669411


rawagner:
@jelkosz do we really want to change it to `workload` ?  Doc https://docs.google.com/document/d/1ks4H-ImD-r0s5eV4FsP_73kRdAP6sKJTRbj8nClpyp0/edit#heading=h.o4npgyz4fz5s still says `work`. Bugzilla says `the word "work" should be "workload"` . I wonder what is that based on ?

PR #232, opened 2019-02-20
add units for disk size in VM disk page

suomiy:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1667828

a

I also changed units to Gi in VmCreateDialog StorageTab for more consistent UI

b


andybraren:
I don't have an easy way of checking at the moment, but I believe the rest of OKD's UI includes the "B" in MiB/GiB/TiB, even though the YAML uses the shorthand Mi/Gi/Ti. Should we change to that for consistency? I think they also include a space between the number value and unit label: "4 GiB".

suomiy:
> I think they also include a space between the number value and unit label: "4 GiB".

there is `1Gi` in pvc detail for example.

We are also showing the `Gi` shorthand in the disks table as seen in the picture. 
The consistency is the way to go, but I don't know what is better solution. IMO shorthands are better, because they are the native k8s format.

rawagner:
`Gi` in PVC detail but `GiB` in new PVC form. So tectonic itself has inconsistencies. Also they do not add space between number and unit (based on PVC detail)

suomiy:
@andybraren maybe you can bring up the issue somewhere higher to get the whole UI consistent?

andybraren:
I was wrong - using `4Gi` without a space is okay for the reasons you both noted. I agree this is something we need to figure out at a higher level, but since shorthand is the native k8s format that's a strong reason to use this format (even though it will look even less familiar to users expecting to see GB/TB).

Weirdly, clicking the unit dropdown in the Add PVC form changes `GiB` to shorthand units. A bug?

pvc-unit-change

suomiy:
seems like it, nevertheless this should go to https://github.com/openshift/console

PR #235, opened 2019-02-21
Delete VM annotations and firmware fields when cloning

rawagner:


src/k8s/clone.js
```diff
@@ -58,17 +62,24 @@ const cleanVm = vm => {
...
+
   delete vm.status;
 };
 
-const updateVm = (vm, name, namespace, description, startVm) => {
+const updateVm = (vm, name, namespace, description, startVm, osName) => {
   vm.metadata.name = name;
   vm.metadata.namespace = namespace;
   vm.metadata.annotations = {
-    ...(vm.metadata.annotations || {}),
```
suomiy: the annotation could be used for many things. What is the main reason for deleting them? 

Wouldn't it be worth it to give the user the choice to keep some of them?

PR #241, opened 2019-02-25
add PVC provision source

suomiy:


rawagner:
I think Provision Source PVC may be confusing for users, lets take a look at example flow:
1. I will select PVC Provision source
2. help says 'Select existing PVC in Storage Tab'
3. I will go to Storage Tab but I have two options here - Create Disk and Attach Disk. How do I know what is PVC ?

@lizsurette any ideas ?

lizsurette:
> I think Provision Source PVC may be confusing for users, lets take a look at example flow:
> 
> 1. I will select PVC Provision source
> 2. help says 'Select existing PVC in Storage Tab'
> 3. I will go to Storage Tab but I have two options here - Create Disk and Attach Disk. How do I know what is PVC ?

+1 I've been wondering if it should actually be something like "ISO" that the user chooses from the provision source list. Just trying to think what the user would actually be adding as a PVC in these cases :) Do you guys know what it would typically be? Would it always be ISO? Or Image?

I think ultimately, we want them to choose maybe ISO, browse to/upload the ISO, and that should be added automatically for them in the Storage tab as a disk backed by PVC. They shouldn't have to know anything about PVC.

What do you think @matthewcarleton?

suomiy:
> Do you guys know what it would typically be? Would it always be ISO? Or Image?

Supported image formats are img, iso, qcow2 - anything can be stored there

I think it can be a bit confusing to call same things differently. Especially when you have a PVC tab in the same UI.

matthewcarleton:
Hey @suomiy, @lizsurette and I just talked this through. I think it makes sense to have image as the option rather than PVC. Then from that point the user can either choose from existing via a dropdown or provide a new one via the url. I totally agree that having PVC here can be a bit confusing since we are talking in terms of virtual machines and it's a new concept. 

The help text could be "Select an existing image or upload via url"

1-4-3-1-basic-provision-image

What do you think?

andybraren:
@matthewcarleton As-designed this looks more like I need to both select an image _**and**_ provide a URL to something. My knowledge is a little outdated, but I believe this would be the first time we would be asking the user to choose between two vertically-aligned options.

Could we do something like what we designed for the vCenter instance dropdown? Where the first option could be "Pull from URL" or something and selecting that would reveal the URL field below it?

Also, I'm probably out of the loop, but what would the Browse button next to the URL field do?

1-4-1-4-basic-provision-import-add

matthewcarleton:
@andybraren ya that's a good point, I think that would be a better option. The help text states either or but if we can enforce that with the dropdown that would be nice too. 
I could be wrong but an image could be uploading via url or I guess local path so it could be worded differently.

suomiy:
sounds reasonable (when having two options PVC x Upload). This PR doesn't address the upload so I just renamed `PVC` to `Image` and it will make more sense in the future

matthewcarleton:
so one follow up on this. Just talked to @aglitke and he suggests Disk rather than image to be more precise.

suomiy:
even when uploading the image? Because they are two different entities.

src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
```diff
@@ -38,6 +35,7 @@ import {
   PROVISION_SOURCE_CONTAINER,
   PROVISION_SOURCE_URL,
   PROVISION_SOURCE_IMPORT,
+  PROVISION_SOURCE_IMAGE,
```
aglitke: Image is a pretty vague term and I suggest we avoid it.  Do you mean we are cloning an existing vm disk?  Are we attaching it (taking ownership of a PVC)?  For clones I'd suggest PROVISION_SOURCE_CLONE_DISK.  For assuming ownership I'd recommend using PROVISION_SOURCE_ATTACH_DISK.

suomiy: We are cloning all PVS right now (attach disk & image). So we should also add clone keyword to attach disk as well - to be consistent.

suomiy: @aglitke @matthewcarleton 

What is the best solution?
1. use IMAGE 
2. put CLONE everywhere in the UI (e.g. button with `Attach Cloned Disk`)
3. use CLONE only in PROVISION_SOURCE_CLONE_DISK option. This would make the UI inconsistent

matthewcarleton: @aglitke maybe you can shed some more light here but if we are cloning regardless of attach or clone maybe we can just keep it as "provision source: disk" and allow them to upload or choose from existing? @suomiy in the designs for the storage step we've updated the action to just "add disk" rather than create/attach just to simply it, which I think aligns with what I'm saying.

matthewcarleton: @suomiy can you clarify what you mean here: 
>We are cloning all PVS right now (attach disk & image). So we should also add clone keyword >to attach disk as well - to be consistent.

suomiy: - Attaching disk by clicking `Attach Disk` in `Storage` tab clones the underlying PVC and then attaches it to the VM
- Choosing provision source `Image` (current impl) clones the underlying PVC and then the VM boots from it

I would prefer to have consistent UI and have the same terminology everywhere:

- either we decide it is important to pass the clone information to the user, then we should point this out everywhere in the UI
- or we decide it is not important and we omit that in the UI

I hope it makes sense

matthewcarleton: @suomiy so can you confirm that we are updating the storage section to show `add disk` rather  two separate actions of `create` and `attach`. I think this will help simplify the problem you are describing.

suomiy: I guess yes it will happen eventually (not that soon though). Can we merge this PR and resolve this once we start implementing the new flow?

aglitke: Attach and clone are two separate operations.  I'd expect attach to directly use an existing pvc without cloning it.  Clone should create a clone of an existing PVC.  There are use cases for both of these.

lizsurette: @aglitke Can you explain more why a user would Attach in one case and Clone in another? I was going to suggest we hide the term Clone from the user since it's really the underlying implementation and in the end it's just pointing to a PVC, but this actually makes a big difference :)

suomiy: I created an issue for this conversation as it is outside of the scope of this PR: https://github.com/kubevirt/web-ui-components/issues/283

rawagner:
LGTM, merging. Lets continue the discussion in https://github.com/kubevirt/web-ui-components/issues/283

PR #243, opened 2019-02-26
v2v: prefill Name and Description based on VMWare VM

mareklibra:
If the user did not change the VM name or description fields within the CreateVmWizard, they are automatically pre-filled with the values from the imported VMWare VM.

To highlight:
- `FormFactory.getFormElement()` passes `onFormChange()` to a custom component. This slightly breaks encapsulation given by strictly context-bound `onChange()` but enables to build more complex flows in a custom component and so keeping business logic better co-located instead of spreading it to different layers (i.e. poluting BasicsettingsTab and CreateVmWizard components with inner-flow details).
- the `onFormChange()` handles `BATCH_CHANGES_KEY` so an array of `{ target, value }` pairs can be passed within a single call. We can consider this as a sort of "transaction". With existing design, it is not possible to call `onFormChange()` multiple-times in a single event handling as it is bound to `basicSettings` at the render time, so multiple write attempts collide. Different possible solutions:
  - use sort of `setTimeout(onFormChange, 0)` - recently working but ugly and not-reliable in the long-term
  - refactor flows around `CreateVmWizard.setState()`, resp. `onChange()` to manipulate with most recent verision of `stepData`. Passing of just the diff instead of recent whole single-stepData would be needed.

### DIFF
--- a/src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
+++ b/src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
@@ -2,7 +2,15 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { get } from 'lodash';
 
-import { FormFactory, validateForm, DROPDOWN, CHECKBOX, TEXT_AREA, POSITIVE_NUMBER } from '../../Form';
+import {
+  FormFactory,
+  validateForm,
+  getFieldValidation,
+  DROPDOWN,
+  CHECKBOX,
+  TEXT_AREA,
+  POSITIVE_NUMBER,
+} from '../../Form';
 import { getName, getMemory, getCpu, getCloudInitUserData } from '../../../utils/selectors';
 import { getTemplate, getTemplateProvisionSource } from '../../../utils/templates';
 import { validateDNS1123SubdomainValue, validateURL, validateContainer } from '../../../utils/validations';
@@ -60,6 +68,7 @@ import {
   CLOUD_INIT_CUSTOM_SCRIPT_KEY,
   HOST_NAME_KEY,
   AUTHKEYS_KEY,
+  BATCH_CHANGES_KEY,
 } from './constants';
 
 import { importProviders } from './providers';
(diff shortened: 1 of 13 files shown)


src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
```diff
@@ -126,6 +128,14 @@ export const onVCenterInstanceSelected = async (
...
 
+const prefillVmName = (basicSettings, onFormChange, vmName, formValid) => {
+  const formName = settingsValue(basicSettings, NAME_KEY);
+  const prevSelectedVmName = settingsValue(basicSettings, PROVIDER_VMWARE_VM_KEY);
+  if (!formName || formName === prevSelectedVmName) {
+    onFormChange({ value: vmName }, NAME_KEY, formValid);
+  }
+};
+
 export const onVCenterVmSelectedConnected = async (
```
suomiy: just a suggestion: wouldn't it be better to move these callbacks from each field onChange to `publish()` in BasicSettingsTab and resolve everything there according to KEY?

mareklibra:
Please do not merge yet. While working on additional functionality, I'll refactor this part as well.

src/components/Wizard/CreateVmWizard/providers/vmwareActions.js
mareklibra: I would like to avoid that as it is splitting business logic across multiple component and especially layers. Good for now, but harder to maintain later. I prefer collocating related things.

src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
```diff
@@ -265,24 +266,40 @@ const asValueObject = (value, validation) => ({
...
 const publish = ({ basicSettings, templates, onChange, dataVolumes }, value, target, formValid, formFields) => {
-  const newBasicSettings = {
-    ...basicSettings,
-    [target]: value,
-  };
+  let newBasicSettings;
+  if (target === BATCH_CHANGES_KEY) {
+    // the "value" is an array of pairs { value, target }
+    const difference = value.value.reduce((map, obj) => {
+      map[obj.target] = { value: obj.value, validation: undefined };
```
rawagner: you should not set validation to undefined. See my comment below

src/components/Wizard/CreateVmWizard/providers/VCenterVmsWithPrefill.js
```diff
@@ -0,0 +1,99 @@
...
+      result.push(namePair);
+    }
+
+    const descrPair = this.prefillVmDescription(basicSettings, onFormChange, vmVmware);
+    if (descrPair) {
+      result.push(descrPair);
+    }
+
+    if (result.length > 0) {
+      onFormChange({ value: result }, BATCH_CHANGES_KEY);
```
rawagner: since you are calling onFormChange, not onChange, you need to also validate field.  See https://github.com/kubevirt/web-ui-components/blob/master/src/components/Form/FormFactory.js#L149 149-161. Maybe you could extract it to separate function and reuse it here

src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
mareklibra: done

src/components/Wizard/CreateVmWizard/providers/VCenterVmsWithPrefill.js
mareklibra: done

src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
```diff
@@ -265,24 +274,48 @@ const asValueObject = (value, validation) => ({
...
+      if (value.value === NO_TEMPLATE) {
+        newBasicSettings[target] = asValueObject(undefined);
+      } else {
+        const allTemplates = getTemplate(templates, TEMPLATE_TYPE_VM);
+        if (allTemplates.length > 0) {
+          const userTemplate = allTemplates.find(template => template.metadata.name === value.value);
+          updateTemplateData(userTemplate, newBasicSettings, dataVolumes);
+        }
       }
+      formValid = validateForm(formFields, newBasicSettings);
```
suomiy: we could move this statement to the end of the method for both branches

mareklibra: To many `if`s :-) 
This time, the validation is just for the `target === USER_TEMPLATE_KEY` branch.
There's still flow where the validation is not executed.

PR #244, opened 2019-02-26
v2v: Delete V2VVMWare object on CreateVmWizard close

mareklibra:
It is just a friendly help to keep things clean.

The garbage collector would remove these left-overs anyway but after a delay.
Thanks to this PR, the V2VVMWare are removed sooner.

### DIFF
--- a/src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
+++ b/src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
@@ -38,6 +38,7 @@ import {
   getTemplateWorkloadProfiles,
   getTemplateOperatingSystems,
   settingsValue,
+  getV2VVmwareName,
 } from '../../../k8s/selectors';
 
 import {
@@ -72,6 +73,7 @@ import {
 } from './constants';
 
 import { importProviders } from './providers';
+import { V2VVMwareModel } from '../../../models';
 
 const getProvisionSourceHelp = basicSettings => {
   const provisionSource = settingsValue(basicSettings, PROVISION_SOURCE_TYPE_KEY);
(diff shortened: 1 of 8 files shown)


src/components/Wizard/CreateVmWizard/CreateVmWizard.js
```diff
@@ -283,10 +283,23 @@ export class CreateVmWizard extends React.Component {
     }
   };
 
+  onHideWrapper = (...eventArgs) => {
+    this.wizardStepsNewVM.forEach(step => {
+      if (step.onCloseWizard) {
+        step.onCloseWizard(this.state.stepData[step.key].value, this.props.k8sKill);
```
suomiy: can you pls send k8sKill wrapped in an object so this can be easier extended for other props in future?

src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
```diff
@@ -346,6 +348,35 @@ const updateTemplateData = (userTemplate, newBasicSettings, dataVolumes) => {
...
+  const v2vvmwareName = getV2VVmwareName(basicSettings);
+  if (v2vvmwareName) {
+    // This is a friendly help to keep things clean.
+    // If missed here (e.g. when the browser window is closed), the kubevirt-vmware controller's garbage
+    // collector will do following automatically after a delay.
+    const resource = {
+      metadata: {
+        name: v2vvmwareName,
+        // TODO: potential issue if the user changed the namespace after creation of the v2vvmware object
+        // to fix: either store the namespace along the v2vvmwareName or empty v2vvmwareName on namespace change
```
suomiy: yes, probably better to delete on namespace change

src/components/Wizard/CreateVmWizard/CreateVmWizard.js
mareklibra: done, I'm calling it `callerContext` to be aligned with my plan to wrap similar web-ui references in such an object and so pass them consistently through-out various web-ui-components calls.
Such a refactoring is planed for a follow-up.

PR #245, opened 2019-02-26
Improve VM details styling for overview view

rawagner:


src/components/Details/VmDetails/VmDetails.js
```diff
@@ -166,99 +169,92 @@ export class VmDetails extends React.Component {
...
-        </h1>
-        {this.state.k8sError && <Alert onDismiss={this.onErrorDismiss}>{this.state.k8sError}</Alert>}
-        <Row>
-          <Col lg={4} md={4} sm={4} xs={4} id="name-description-column">
-            <dl>
-              <dt>Name</dt>
-              <dd>{vm.metadata.name}</dd>
-              <dt>Description</dt>
+      <React.Fragment>
+        {!overview && (
```
suomiy: please can you make variables out of these? at least the first one?

rawagner: variable `header` created from both

suomiy:
Can we have a cosmos a fixture of this and at least a snapshot test?

rawagner:
added cosmos and test

@matthewcarleton 
please take a look at latest design implementation and let me know what you think
vm_overview

Lets take a look at VM item 
 - design shows status, metrics and kebab menu
 - i implemented only status as we do not have metrics yet and kebab menu is not present on other items (pods, deployment configs..) so we will need to implement it in upstream first

VM Overview:
 - metrics are missing in implementation again because we do not have them yet
 - description list items are in different order - VM Overview is the same page as VM Details (with less items) https://docs.google.com/document/d/1Fh2eb9TD0aq9JulVbLlss2ucdqYzx_uqTaThse_fcgE/edit?ts=5c191346 and IMHO we should have the same ordering of description items for both Overview and Details otherwise I would find it confusing

matthewcarleton:
@rawagner thanks! yes I think having those metrics and the actions in the kebab menu would be very useful. I am assuming the same is true for the pod chart? 
Can you provide a screen shot of what the page would look like if grouped by label?
In the design we also offer further info of type above the name of the resource, which might make more sense in the label view. But it could be good to have it for consistency. 
I would suggest removing the Description completely if it's not available. It would be a cleaner approach.
Is there a technical reason why we can't have "View full details" in the flyout section? I think that adds quite a bit of value.
In the design we do have a resource option, but it wasn't fully explored at the time.  We could possibly provide Snapshots, NICs and disks there, although that might be overkill for this view.

src/components/Details/VmDetails/fixtures/VmDetails.fixture.js
```diff
@@ -113,20 +116,34 @@ const vmiFixture = {
   },
 };
 
+const k8sGet = () =>
+  new Promise(resolve => {
+    resolve(fedora28);
+  });
+
+const overviewStyle = {
```
suomiy: Is there a specific reason to have a set width for the component? I think it would leave more room for the experimentation/exploration of component if we omit that. The user can always use the browser to change viewport size.

rawagner: ok, i removed additional styling from fixture

rawagner:
> @rawagner thanks! yes I think having those metrics and the actions in the kebab menu would be very useful. I am assuming the same is true for the pod chart?
Yes, same for pod charts. 
> Can you provide a screen shot of what the page would look like if grouped by label?
> In the design we also offer further info of type above the name of the resource, which might make more sense in the label view. But it could be good to have it for consistency.

Ah, I didnt notice it at first :) Anyways we should add that to upstream tectonic. I can work on that and create a PR as a followup

> I would suggest removing the Description completely if it's not available. It would be a cleaner approach.

Ok, removing description when there is none

> Is there a technical reason why we can't have "View full details" in the flyout section? I think that adds quite a bit of value.

No technical reason. But we should also have that consistent with the rest of the resources shown in overview - for pods/ deployment configs etc and we should implement it in upstream too. I can work on that as a followup.

> In the design we do have a resource option, but it wasn't fully explored at the time. We could possibly provide Snapshots, NICs and disks there, although that might be overkill for this view.

Yeah, i didnt know what to put under resources :) Once we find out what to show there I will be happy to implement it

Adding screenshots for VM with no description and VM with description grouped by Application (means resource has label app=\<something\>, if the label is missing its put under `other resources`

vm_no_desc


vm_desc

matthewcarleton:
@rawagner this looks great! 
+1 on following up on and continuing to make things more consistent!
Ya we should push further on the resources. I'll investigate this further and update the doc.

rawagner:
> @rawagner this looks great!
> +1 on following up on and continuing to make things more consistent!
> Ya we should push further on the resources. I'll investigate this further and update the doc.
> Are we able to pull in Utilization to the overview?

I could add utilization to overview and details but It requires openshift 4 (to be able to pull metrics from prometheus) with kubevirt installed (to get VM metrics into prometheus). But currently we do not have such environment :/ Once we will have something I will add them - issue is here https://github.com/kubevirt/web-ui-components/issues/251

rawagner:
@suomiy rebased, can you take a look ? thanks!

src/components/Details/VmDetails/fixtures/VmDetails.fixture.js
```diff
@@ -113,20 +114,19 @@ const vmiFixture = {
   },
 };
 
+const k8sGet = () =>
```
suomiy: we could add this to tests/k8sGet function 
note: k8sGet should be converted to return promise

rawagner: fixed

PR #252, opened 2019-03-07
v2v: prefill cpu and memory

mareklibra:
Contains:
- refactoring of the VCenterVmsWithPrefill to simplify unit testing and share code
- prefil of CPU and Memory using `Custom` flavor


mareklibra:
Code-coverage will be improved within #253

PR #256, opened 2019-03-08
v2v: prefill networks

mareklibra:
The `Networks` tab of the Create VM Wizard is filled by networks of the imported VMWare VM.

The user is expected to pair source networks with the target network configurations.

### DIFF
--- a/src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
+++ b/src/components/Wizard/CreateVmWizard/BasicSettingsTab.js
@@ -70,6 +70,7 @@ import {
   HOST_NAME_KEY,
   AUTHKEYS_KEY,
   BATCH_CHANGES_KEY,
+  INTERMEDIARY_NETWORKS_TAB_KEY,
 } from './constants';
 
 import { importProviders } from './providers';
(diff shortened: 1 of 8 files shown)


mareklibra:
Once the user passes the `Basic Settings` tab, the networks are prefilled based on source VM configuration.
The tab is intentionally in an error state to make the user perform the selection of network configurations.

Please note:
- the `Create` button is disabled
- the `Actions` kebab is missing (as it would contain just a single `Remove` action)

01_entry

---

02_select

---

03_end

lizsurette:
> The tab is intentionally in an error state to make the user perform the selection of network configurations.

How about rather than intentionally dropping the user into a saved error state, we drop them into the edit state where they need to choose the network configuration and save? We would save them one step of having to click "edit".

mareklibra:
> > The tab is intentionally in an error state to make the user perform the selection of network configurations.
> 
> How about rather than intentionally dropping the user into a saved error state, we drop them into the edit state where they need to choose the network configuration and save? We would save them one step of having to click "edit".

It makes sense, the user is expected to provide the data.
I am filing new issue to track this - recent table component will need to be extended to support that.

PR #257, opened 2019-03-08
Host status selector

jtomasek:
This pull request adds BaremetalHostModel definition to models, introduces basic getHostStatus selector to be used with hosts filtering and determining host status.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,5 +20,6 @@ export * from './components/ConfigurationSummary';
 // helpers
 export * from './constants';
 export * from './utils';
+export * from './models';
 export * from './k8s/request';
 export * from './k8s/selectors';
--- a/src/models/index.js
+++ b/src/models/index.js
@@ -19,6 +19,19 @@ export const VirtualMachineModel = {
   id: 'virtualmachine',
 };
 
+export const BaremetalHostModel = {
+  label: 'Bare Metal Host',
+  labelPlural: 'Bare Metal Hosts',
+  apiVersion: 'v1alpha1',
+  path: 'baremetalhosts',
+  apiGroup: 'metalkube.org',
+  plural: 'baremetalhosts',
+  abbr: 'BMH',
+  namespaced: true,
+  kind: 'BareMetalHost',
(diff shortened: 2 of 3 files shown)


src/components/HostStatus/getHostStatus.js
```diff
@@ -0,0 +1,3 @@
+import * as _ from 'lodash';
+
+export const getHostStatus = host => _.get(host, ['metadata', 'labels', 'metalkube.org/operational-status']);
```
suomiy: this should probably be in `utils/selectors` as it is not a component. Or make a file  `utils/hostSelectors`,  if you think there will be more of them

jtomasek: I have taken an inspiration for this in VmStatus/getVmStatus.js selector which is colocated to the component. I expect this selector to grow in a similar manner and a statu component to be added. But yeah, I am happy to move it under utils as you're saying.

suomiy: `getVmStatus.js` does a bit more than just data selection and it belongs to `VmStatus` component. Nevertheless some of the functionality from `getVmStatus.js` could be refactored to some selectors file.

jtomasek: Done

src/index.js
```diff
@@ -20,5 +20,6 @@ export * from './components/ConfigurationSummary';
 // helpers
 export * from './constants';
 export * from './utils';
+export * from './models';
```
rawagner: Why do you want to export models ? We usually have all models in `web-ui` and `web-ui-components` contains only copies of necessary models from `web-ui`

jtomasek: The TODO comment in `kubevirt/modes/vm.ts` [1] suggest reusing the models from web-ui-components. Exporting models here allows doing that. If I am testing it correctly, then I just need to re-export the model in `metalkube/models/host.ts` which I find better than maintaining 2 copies in 2 repositories.

rawagner: makes sense :)

PR #264, opened 2019-03-13
move vmStatus logic to utils/status/vm

suomiy:
web-ui PR: https://github.com/kubevirt/web-ui/pull/224


mareklibra:
@suomiy , are there any additional changes than just
- moving code around
- rename
   - `getVmStatus()` to `getSimpleVmStatus()`
   -  `getVmStatusDetail()` to `getVmStatus()`,

please?

suomiy:
@mareklibra 

no additional changes besides that

@jtomasek 

I guess they could live under `src`, but nevertheless they are utility functions in their essence

jtomasek:
> I guess they could live under `src`, but nevertheless they are utility functions in their essence

The question here is rather whether we need to have a separate `status` directory or if this should live in `selectors/` as these functions are selectors too.

suomiy:
They are quite special kind of selectors and I would rather extract the selection logic of them into the selectors. On the other hand it is not good to pollute the src directory. Not sure what is the best location for the statuses. 

\+ ofc there will be statuses for other entitites in the future as well

suomiy:
let's merge #265 first

suomiy:
rebased and moved selection logic to selectors.

I decided to leave the status in utils as there are just two exported functions and constants. We can move it to top `src` folder once status functionality gets more complicated.

I think it should be enough for now. thoughts?

PR #265, opened 2019-03-13
refactor selectors

suomiy:


jtomasek:
Since we're doing the refactoring, aren't we overloading `utils/` here? Why not putting the selectors directly under `src/` ?

suomiy:
I guess they could be under `src` because they are quite important part of the application

suomiy:
rebased and moved to `src`

suomiy:
rebased, @rawagner @mareklibra can we go forward with this patch?

PR #266, opened 2019-03-14
v2v: use datavolume instead of PVC

mareklibra:
Hot fix only.
The PVC should be enough but recent validation within UI expects the PVC to be already
created if used in the StorageTab of the Create VM Wizard.

The v2v flow will create the PVC during final step of the wizard, right before
creating the Conversion POD.


suomiy:
> The v2v flow will create the PVC during final step of the wizard, right before
creating the Conversion POD.

aren't the PVCs created with the vm's TemplateModel ? How does the Conversion pod get started? link to the documentation (if there is one) should be enough..

Issue #263, opened 2019-03-13
ClusterOverview: Improve responsiveness of Cards


rawagner:

rawagner:
fixed by https://github.com/kubevirt/web-ui-components/pull/268

suomiy:
responsiveness of contents of cards is still lacking

PR #268, opened 2019-03-14
refactor ClusterOverview - introduce Dashboard and DashboardCard component

suomiy:
fixes https://github.com/kubevirt/web-ui-components/issues/261 and solves responsivness od dashboard #263 (not the cards by itself)

Dashboard and DashboardCard are mostly wrappers and are good entrypoints for introducing our own css & the provider

Two new dependecies were introduced:
- `@patternfly/react-core`: this would happen inevitably in the future
- `react-responsive`: helps with the responsive layout ([link](https://github.com/contra/react-responsive))

The preview: https://clever-allen-639701.netlify.com/?component=ClusterOverview%2FJ&fixture=ClusterOverview

@andybraren @lizsurette Can you please check the responsiveness of the dashboard and tell me what you think about the location of the cards?


mareklibra:
@suomiy , the `react-responsive` looks/behaves much better than former layout https://kubevirt.io/web-ui-components/?component=ClusterOverview%2FFe&fixture=ClusterOverview .

This effort represents something more generic and I think we should be aligned with patternfly for that.
Maybe I am wrong, but I can't find suitable component for our needs in recet pf-react.
So it might be a nice PR adding it there. Wdyt @jelkosz ?

src/components/ClusterOverview/Utilization/Utilization.js
```diff
@@ -43,8 +50,30 @@ export class Utilization extends React.PureComponent {
...
+UtilizationBody.title = 'Cluster utilization';
+
+UtilizationBody.propTypes = {
+  stats: PropTypes.object.isRequired,
+};
+
+const Utilization = ({ stats, loaded }) => (
+  <DashboardCard>
+    <DashboardCardHeader>
+      <DashboardCardTitle>Details</DashboardCardTitle>
```
rawagner: Should be Cluster Utilization

jelkosz:
> So it might be a nice PR adding it there. Wdyt @jelkosz ?

eh, in general I agree. But based on previous experience, contributing anything into pf-react is a terrible time sink and I would do it with very little priority. Maybe first opening an issue just if it would be even accepted?

> Edit: I would not block this PR on the potential pf-react effort. I like the approach implemented here.

+1 -> we need to be careful. This dashboard might be shown on differently sized screens and we want to be careful to not to fall apart in the worst possible moment ;)

package.json
```diff
@@ -41,11 +41,13 @@
   },
   "dependencies": {
     "@patternfly/react-console": "1.x",
+    "@patternfly/react-core": "2.3.x",
```
mareklibra: This might slightly complicate web-ui build which uses
```
"@patternfly/react-core": "^1.43.5"
```

Do we need 2.x now?
What about declaring that as `pperDependency`?

suomiy:
> > So it might be a nice PR adding it there. Wdyt @jelkosz ?

> eh, in general I agree. But based on previous experience, contributing anything into pf-react is a terrible time sink and I would do it with very little priority. Maybe first opening an issue just if it would be even accepted?

Ok, I will create an issue and we'll see how it goes from there

> Can you please fix the outdated snapshot ? Otherwise LGTM

fixed

package.json
suomiy: not sure, but we'll certainly need it once we start using new table and inline editing

ok, I will declare it as peerDependency for now

suomiy: declared as a peerDependency: `"@patternfly/react-core": "1.43.x"`

but it will come to us again in the future

Issue #262, opened 2019-03-13
ClusterOverview: implement Provider


rawagner:

mareklibra:
Not sure how much time will https://github.com/kubevirt/web-ui-components/pull/268 take to merge. I will stack this change on top of its wip. It looks to me like a good starting point.

rawagner:
@mareklibra i agree

PR #269, opened 2019-03-15
Introduce ClusterOverviewContext component

mareklibra:
Used to manage large heterogenous data set for inner components (cards).

Fixes: #262


src/components/ClusterOverview/ClusterOverview.js
```diff
@@ -15,70 +17,80 @@ import Capacity from './Capacity/Capacity';
...
-  capacityStats: PropTypes.object.isRequired,
-  utilizationStats: PropTypes.object.isRequired,
-  eventsData: PropTypes.object.isRequired,
-  consumersData: PropTypes.object.isRequired,
-};
+const MainCards = () => (
+  <GridItem lg={6} md={12} sm={12}>
+    <Grid>
+      <GridItem span={6}>
+        <ClusterOverviewContextConsumer Component={Health} dataPath="healthData" />
```
suomiy: Wouldn't it be better to consume the data inside the component itself? There, the component can decide what data it needs (if any). This helps with sharing the resources (e.g. vms, pods)

suomiy: meaning: each card decides what it consumes

rawagner: +1 i would like to have such flexibility

mareklibra: Good question.

Such approach would limit reusability of the cards (not sure if needed). It would state requirement on
- existence of the Context (means Provider)
- specific structure of the Context

Moving manipulation with the consumer one step up leads to centralization of manipulation with the data structure and leaves the components (means cards) independent.

There's no requirement to use the `ClusterOverviewCOntextConsumer` component for all children. It's just recent case for the mock data. I expect that some of the children will be created by using the `ClusterOverviewContext.Consumer` directly.

I can imagine that other dashboards will implement the Provider "on their own" so this will help - they will not need to meet certain structure of the context data.

mareklibra: What about having both :-) : 
- keeping the card components using props
- and wrapping them with on-liners to provide them with the context-data

suomiy: I think the specific structure is already prescribed by the props and it is still needed to specify the dataPath. We can still export two components per card (one connected, one with props) if reusability issue arise.

suomiy: +1 you were a bit faster (:

mareklibra: done

src/components/ClusterOverview/ClusterOverviewContext/ClusterOverviewContext.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { get } from 'lodash';
+import PropTypes from 'prop-types';
+
+export const ClusterOverviewContext = React.createContext({});
+
+export const ClusterOverviewContextGenericConsumer = ({ Component, dataPath }) => (
```
suomiy: I get that for now, it is enough, but this logic will move to the cards once they need more than one property.

mareklibra: sure, this is as expected.
The wrapping (`*Connected`) is separate layer with its simplicity/complexity

PR #270, opened 2019-03-15
Upgrade yarn.lock

mareklibra:
Dependencies are upgraded within limits set in the package.json (untouched).

Needs to be verified manually.


mareklibra:
Will be merged _after_ release today as this might be breaking change

PR #271, opened 2019-03-18
Create Cluster Details Card

rawagner:
Create Cluster Details Card
    Add loading state to Details item

### DIFF
--- a/src/components/ClusterOverview/Details/ClusterDetails.js
+++ b/src/components/ClusterOverview/Details/ClusterDetails.js
@@ -1,0 +1,74 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { get } from 'lodash';
+
+import { ClusterOverviewContextGenericConsumer } from '../ClusterOverviewContext';
+import { InlineLoading } from '../../Loading';
+import { Details } from './Details';
+
+export const ClusterDetails = ({ infrastructure, openshiftVersionResponse, LoadingComponent }) => {
+  const items = {
+    name: {
+      title: 'Name',
+      value: getClusterName(infrastructure),
+      isLoading: !infrastructure,
+    },
+    provider: {
+      title: 'Provider',
+      value: get(infrastructure, 'status.platform'),
+      isLoading: !infrastructure,
+    },
+    rhhi: {
+      title: 'RHHI version',
+      value: 'hardcoded-version', // this will be hardcoded for the demo
+      isLoading: false,
+    },
+    openshift: {
+      title: 'Openshift version',
+      value: getOpenshiftVersion(openshiftVersionResponse),
+      isLoading: !openshiftVersionResponse,
+    },
+  };
+
+  return <Details items={items} LoadingComponent={LoadingComponent} />;
+};
+
+ClusterDetails.defaultProps = {
+  infrastructure: null,
+  openshiftVersionResponse: null,
+  LoadingComponent: InlineLoading,
+};
+
+ClusterDetails.propTypes = {
+  infrastructure: PropTypes.object,
+  openshiftVersionResponse: PropTypes.object,
+  LoadingComponent: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
+};
+
+export const ClusterDetailsConnected = () => (
+  <ClusterOverviewContextGenericConsumer Component={ClusterDetails} dataPath="detailsData" />
+);
+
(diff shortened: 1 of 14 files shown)


rawagner:
@lizsurette @andybraren

if any item in details card is too long, we want to set text-overflow to ellipsis and show a popup tooltip with full value, right ?
details_overflow

src/components/ClusterOverview/Details/Details.js
```diff
@@ -8,51 +9,57 @@ import {
   DashboardCardTitle,
 } from '../../Dashboard/DashboardCard';
 import { ClusterOverviewContextGenericConsumer } from '../ClusterOverviewContext';
+import { Loading } from '../../Loading';
```
suomiy: The loading component still has `blank-slate-pf` style inside. I would suggest to make two component `<Loading>` and `<InlineLoading>`

rawagner:
@suomiy @mareklibra I have issues with cosmos fixtures as it seems Provider for some reason does not work correctly, it does not pass props and always returns undefined. Any ideas ? It also does not work on master, every component just renders loading which is due to defaultProps. https://kubevirt.io/web-ui-components/?component=p&fixture=ClusterOverview

suomiy:
> @suomiy @mareklibra I have issues with cosmos fixtures as it seems Provider for some reason does not work correctly, it does not pass props and always returns undefined. Any ideas ?

You have to wrap the fixture props in value: 

```
const ClusterOverview = props => (
  <ClusterOverviewContext.Provider value={props}>
    <ClusterOverviewComponent />
  </ClusterOverviewContext.Provider>
);
```

src/components/Loading/Loading.js
```diff
@@ -3,6 +3,23 @@ import PropTypes from 'prop-types';
 
 import { prefixedId } from '../../utils';
 
+export const InlineLoading = ({ text, id }) => (
+  <div id={id} key={prefixedId(id, 'progress') || 'progress'}>
+    <div className="spinner spinner-lg blank-slate-pf-icon" />
+    {text && <h3 className="blank-slate-pf-main-action">{text}</h3>}
```
suomiy: we probably do not want h3 text inside the InlineLoading. I do not think the text is necessary at all.

also, not sure if it would make sense to have a span instead of a div.

suomiy: or 2nd option: having the text inline

rawagner: text is now removed and I also made spinner size configurable

rawagner:
> > @suomiy @mareklibra I have issues with cosmos fixtures as it seems Provider for some reason does not work correctly, it does not pass props and always returns undefined. Any ideas ?
> 
> You have to wrap the fixture props in value:
> 
> ```
> const ClusterOverview = props => (
>   <ClusterOverviewContext.Provider value={props}>
>     <ClusterOverviewComponent />
>   </ClusterOverviewContext.Provider>
> );
> ```

Thanks! I was expecting that it will be something small and fishy :)

src/components/ClusterOverview/Details/Details.js
```diff
@@ -8,51 +11,77 @@ import {
...
-      <DetailsBody data={data} />
+    <DashboardCardBody>
+      <DetailsBody items={items} LoadingComponent={LoadingComponent} />
     </DashboardCardBody>
   </DashboardCard>
 );
 
 Details.defaultProps = {
-  loaded: false,
+  LoadingComponent: Loading,
```
suomiy: not sure if it is not published yet, but please use the `InlineLoading` component here

src/components/ClusterOverview/Details/ClusterDetails.js
```diff
@@ -0,0 +1,76 @@
...
+  return clusterName;
+};
+
+const getOpenshiftVersion = openshiftVersionResponse => {
+  const result = get(openshiftVersionResponse, 'data.result');
+  let version;
+  if (result) {
+    let item = result;
+    if (Array.isArray(result)) {
+      [item] = result;
```
mareklibra: can we use more obvious syntax here, please?
Sort of
```
item = result[0]
```

mareklibra:
```
const item = Array.isArray(result) ? result[0] : result
```

rawagner: @mareklibra updated

PR #273, opened 2019-03-18
implement Inventory Card (statuses)

suomiy:
- refactored vmStatus and moved some logic to selectors/other statuses.
- changed API of vmStatus for easier usage accross web-ui+components
- refactored passing data to ClusterDetails
- introduced pod/node/pvc status

sibling: https://github.com/kubevirt/web-ui/pull/242


suomiy:
This PR now depends on and includes refactoring of vmStatus and other statuses.

@pcbailey can you please take a look?

src/utils/status/vm/vmStatus.js
```diff
@@ -137,16 +127,22 @@ const isWaitingForVmi = vm => {
...
 
-export const getVmStatus = (vm, launcherPod, importerPods, migration) =>
-  isBeingMigrated(vm, migration) || // must be precceding isRunning() since vm.status.ready is true for a migrating VM
-  isRunning(vm) ||
-  isReady(vm) ||
-  isVmError(vm) ||
-  isCreated(vm, launcherPod) ||
-  isBeingImported(vm, importerPods) ||
-  isWaitingForVmi(vm) || { status: VM_STATUS_UNKNOWN };
+export const getVmStatus = (vm, pods, migrations, importerPods = null) => {
```
suomiy: note: it is now possible to pass pods without any filtering to get the correct vm status. ImporterPods can be passed as an optimization which is up to the consumer

suomiy:
the design:

inventory

src/components/ClusterOverview/Inventory/Inventory.js
```diff
@@ -1,80 +1,66 @@
...
-    </div>
-  ) : (
-    <div>
-      <span className="glyphicon glyphicon-remove-circle" />
-    </div>
-  );
-};
+const InventoryBody = ({ nodes, pods, vms, vmis, pvcs, migrations }) => (
+  <React.Fragment>
+    <InventoryRow title="Hosts" {...mapNodesToProps(nodes)} />
```
honza: Nodes and hosts are two different things, AFAIK

suomiy:
fixed

src/components/ClusterOverview/Inventory/Inventory.js
suomiy: Yes, they are. I changed it to `Nodes`.

Nevertheless, I don't what is the resolution about this issue. If both nodes and hosts will be present in this card.

src/components/ClusterOverview/Inventory/InventoryItemStatus.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Icon } from 'patternfly-react';
+
+const StatusPropTypes = {
+  count: PropTypes.number.isRequired,
+};
+
+const Ok = ({ count }) => (
```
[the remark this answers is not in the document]
suomiy: Renamed. I guess yes, they would have to be renamed once they are exported from this file.

suomiy:
note: relevant changes https://github.com/kubevirt/web-ui/pull/237

src/components/ClusterOverview/Inventory/Inventory.js
```diff
@@ -1,80 +1,66 @@
...
-    </div>
-  ) : (
-    <div>
-      <span className="glyphicon glyphicon-remove-circle" />
-    </div>
-  );
-};
+const InventoryBody = ({ nodes, pods, vms, vmis, pvcs, migrations }) => (
+  <React.Fragment>
+    <InventoryRow title="Nodes" {...mapNodesToProps(nodes)} />
```
rawagner: this is not generic enough. Other overviews will want to show different resources here. Inventory card should accept something like `items` and every item would define title and its statuses.

suomiy: this component is not meant to be generic, but an implementation of a card in clusterOverview dashboard.

If someone wishes to have inventory card in their own dashboard, they can create their card with their own InventoryRows. This should be simple and generic enough and can also be enhanced if something is missing.

rawagner: In that case, we should also remove `ClusterDetails` and use `Details` only. If anybody will want to use `Details` card they will implement their own with their `DetailItems`. I do not mind this approach as long as its consistent among all cards.

suomiy: Ok, I will refactor it. I would rather do it in the follow up as this PR is already packed.

rawagner: +1

PR #274, opened 2019-03-18
Create labels for elements in vertical forms

honza:
When the form is vertical, we simply use the `<ControlLabel>` component to display the label.  When it's horizontal, we wrap the simple `<ControlLabel>` component in a `<Col>` --- just like it was done before.

This retains the previous functionality of horizontal forms.

### DIFF
--- a/src/components/Form/FormFactory.js
+++ b/src/components/Form/FormFactory.js
@@ -176,6 +176,16 @@ const onChange = (formFields, formValues, value, key, onFormChange) => {
   onFormChange({ value, validation }, key, formValid); // to update state data
 };
 
+const getLabel = field =>
+  field.type !== 'checkbox' && (
+    <React.Fragment>
+      <ControlLabel className={field.required ? 'required-pf' : null}>{field.title}</ControlLabel>
+      {field.help && (
+        <FieldLevelHelp className="kubevirt-form-group__field-help" placement="right" content={field.help} />
+      )}
+    </React.Fragment>
+  );
+
 const getFormGroups = ({ fields, fieldsValues, onFormChange, textPosition, labelSize, controlSize, horizontal }) =>
   Object.keys(fields)
     .filter(key => fields[key] && (!fields[key].isVisible || fields[key].isVisible(fieldsValues)))
@@ -187,6 +197,9 @@ const getFormGroups = ({ fields, fieldsValues, onFormChange, textPosition, label
       const validationMessage = get(validation, 'message');
       const hasValidationMessage = !!validationMessage;
       const hasAddendum = !!field.addendum;
+      let colProps = {
+        sm: controlSize,
+      };
 
       const child = getFormElement({
         ...field,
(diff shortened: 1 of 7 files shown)


rawagner:
This would break existing horizontal forms as we do not expect that label will be displayed. See

before
horizontal_form_ok


after
horizontal_form

Maybe you could add another prop for form ie `showLabels` ?

honza:
Something like this? If you think that makes sense, I can work on fixing the tests. With this change, my form looks the way I'd like. :smile:

src/components/Form/FormFactory.js
```diff
@@ -225,8 +253,9 @@ const getFormGroups = ({ fields, fieldsValues, onFormChange, textPosition, label
             'kubevirt-form-group--no-help': !horizontal && !hasValidationMessage,
           })}
         >
-          {label}
+          {!showLabels && label}
```
suomiy: I think the naming is a bit unfortunate here, because we show the label even if we say `!showLabels`

src/components/Form/FormFactory.js
```diff
@@ -200,19 +209,38 @@ const getFormGroups = ({ fields, fieldsValues, onFormChange, textPosition, label
...
-                {field.help && (
-                  <FieldLevelHelp className="kubevirt-form-group__field-help" placement="right" content={field.help} />
-                )}
-              </React.Fragment>
-            )}
+            {labelFragment}
           </Col>
         );
+
+        if (horizontal) {
```
suomiy: it seems this is always true, can you simplify the ifs which use `horizontal`?

src/components/Form/FormFactory.js
```diff
@@ -200,19 +209,38 @@ const getFormGroups = ({ fields, fieldsValues, onFormChange, textPosition, label
       });
 
       let label;
-      if (horizontal && field.title) {
+
+      const labelFragment = (
```
suomiy: could you extract this to a function?

src/components/Form/FormFactory.js
honza: Lol, I must have been half asleep when I wrote this.

src/components/Form/FormFactory.js
honza: Done, good idea

src/components/Form/FormFactory.js
honza: I didn't end up needing it.

src/components/Form/FormFactory.js
```diff
@@ -176,6 +176,19 @@ const onChange = (formFields, formValues, value, key, onFormChange) => {
   onFormChange({ value, validation }, key, formValid); // to update state data
 };
 
+const getLabel = field => (
+  <>
```
suomiy: I do not see a benefit of wrapping this in a Fragment

honza: It won't compile without it. How do you suggest it could be improved?

suomiy: seems to work fine here
```jsx
const getLabel = field =>
  field.type !== 'checkbox' && (
    <React.Fragment>
      <ControlLabel className={field.required ? 'required-pf' : null}>{field.title}</ControlLabel>
      {field.help && (
        <FieldLevelHelp className="kubevirt-form-group__field-help" placement="right" content={field.help} />
      )}
    </React.Fragment>
  );
```

honza: Done

PR #281, opened 2019-03-20
Capacity: refactor for real data

mareklibra:


src/components/ClusterOverview/Capacity/Capacity.js
```diff
@@ -15,56 +15,72 @@ const pfGetUtilizationDonutTooltipContents = d =>
...
+  title: PropTypes.string.isRequired,
+  used: PropTypes.number.isRequired,
+  total: PropTypes.number.isRequired,
+  unit: PropTypes.string.isRequired,
+};
 
+const CapacityBody = ({ stats }) => (
+  <div className="kubevirt-capacity__items">
+    {Object.keys(stats)
+      .filter(key => stats[key].data && !Number.isNaN(stats[key].data.used) && !Number.isNaN(stats[key].data.total))
```
rawagner: So now you do not show CapacityItem if it does not have data yet. I would suggest to show title and under it show loading. IMO it will look better - user knows what will be shown right away..CPU/MEMORY/STORAGE/NETWORK and we will also prevent these items jumping around.

mareklibra: done, for screenshots see https://github.com/kubevirt/web-ui/pull/235#issuecomment-475223232

src/components/ClusterOverview/Capacity/Capacity.js
```diff
@@ -15,70 +17,98 @@ const pfGetUtilizationDonutTooltipContents = d =>
...
+  title: PropTypes.string.isRequired,
+  used: PropTypes.number.isRequired,
+  total: PropTypes.number.isRequired,
+  unit: PropTypes.string.isRequired,
+};
 
+const CapacityItemLoading = ({ title }) => (
+  <div className="kubevirt-capacity__item">
+    <h2 className="kubevirt-capacity__item-title">{title}</h2>
+    <InlineLoading />
```
rawagner: Can you make loading component configurable ? so we could pass the one from tectonic

mareklibra: done

src/components/ClusterOverview/Capacity/Capacity.js
```diff
@@ -15,70 +17,103 @@ const pfGetUtilizationDonutTooltipContents = d =>
...
-              </div>
-            </div>
-          );
-        })}
+const CapacityItem = ({ idSuffix, title, used, total, unit }) => {
+  const available = total - used;
+  const columns = [['Used', used], ['Available', available]];
+
+  const sum = columns.reduce((acc, x) => acc + x[1], 0);
+  const percentageUsed = Math.round((100 * columns[0][1]) / sum);
```
suomiy: why not use the `used` variable here instead?

suomiy: please add check for division by zero, so we don't show NaN in the UI

src/components/ClusterOverview/Capacity/Capacity.js
```diff
@@ -15,70 +17,103 @@ const pfGetUtilizationDonutTooltipContents = d =>
...
-                />
-              </div>
-            </div>
-          );
-        })}
+const CapacityItem = ({ idSuffix, title, used, total, unit }) => {
+  const available = total - used;
+  const columns = [['Used', used], ['Available', available]];
+
+  const sum = columns.reduce((acc, x) => acc + x[1], 0);
```
suomiy: isn't the sum equal to total?

mareklibra: good point. Moving code here and there ...

src/components/ClusterOverview/Capacity/Capacity.js
mareklibra: done

src/components/ClusterOverview/Capacity/Capacity.js
```diff
@@ -1,7 +1,10 @@
 import React from 'react';
+import * as _ from 'lodash';
```
rawagner: import just the needed function (isNaN) plase

src/utils/utils.js
```diff
@@ -68,3 +68,34 @@ export const getResource = (
...
+  TB: 1024 ** 4,
+  PB: 1024 ** 5,
+};
+
+export const formatBytes = (bytes, preferredUnit, fixed = 2) => {
+  if (preferredUnit) {
+    return { value: Number((bytes / BYTE_UNITS[preferredUnit]).toFixed(fixed)), unit: preferredUnit };
+  }
+
+  if (bytes < 1024) return { value: bytes, unit: 'B' };
```
rawagner: we could reuse `BYTE_UNITS` here

src/utils/utils.js
```diff
@@ -68,3 +68,34 @@ export const getResource = (
...
+const BYTE_UNITS = {
+  B: 1,
+  KB: 1024 ** 1,
+  MB: 1024 ** 2,
+  GB: 1024 ** 3,
+  TB: 1024 ** 4,
+  PB: 1024 ** 5,
+};
+
+export const formatBytes = (bytes, preferredUnit, fixed = 2) => {
```
suomiy: this could have a better implementation in the future

src/utils/utils.js
mareklibra: done

src/components/ClusterOverview/Capacity/Capacity.js
mareklibra: done - it's even better to use `Number.isNaN()`

src/utils/utils.js
```diff
@@ -68,3 +68,25 @@ export const getResource = (
 
   return res;
 };
+
+const BYTE_UNITS = {
+  B: 0,
```
mareklibra: We can even use an array and its index for that, but ...

src/utils/utils.js
mareklibra: done

cloudbehl:
@mareklibra I remember we had similar kind of conversion implementation in older openshift console(https://github.com/openshift/origin-web-console).
 
https://github.com/openshift/origin-web-console/blob/a6055df09882c490536e5fb50e8214cc2cfa6680/app/scripts/filters/util.js

Just an FYI, If this can benefit us in anyway for understanding the validation and mapping of the value.

PR #285, opened 2019-03-22
[WIP]Validations

flofuchs:
This adds a general BaremetalHostStatus component to be used to show that status in the host listing, as well as a validations-specific selector and component.


dhellmann:
Do we really want to be doing these checks in the UI code?

flofuchs:
> Do we really want to be doing these checks in the UI code?

Not necessarily. The UI needs a list of validation errors, but the checks themselves could also be performed somewhere else. Somewhere next to the hardware profile detection maybe? (https://github.com/metalkube/baremetal-operator/blob/master/pkg/controller/baremetalhost/baremetalhost_controller.go#L306)

PR #287, opened 2019-03-24
Render VmDetails with deleted template curreclty.

yaacov:
**Description**
When we want to render VmDetails with missing VmTemplate we get a warning that we can not find the template. But It is OK to delete a template of an existing vm, the view should silently fall back to vm's current resources.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1685393

**Screenshots**

When template is missing, show warning triangle instead of an error:

Peek 2019-03-31 14-35

Replace drop-down with a label when only one flavor option is available:

Peek 2019-03-31 16-34

Add an optional template resource link method to vmDetails:

Peek 2019-04-01 11-43

Show labels for CPU and Memory in flavor form:

Peek 2019-04-01 14-46

### DIFF
--- a/src/components/Details/Flavor/Flavor.js
+++ b/src/components/Details/Flavor/Flavor.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 
 import { prefixedId } from '../../../utils';
@@ -7,15 +7,11 @@ import { InlineEdit } from '../../InlineEdit/InlineEdit';
 import { CUSTOM_FLAVOR } from '../../../constants';
 import { getTemplateFlavors, settingsValue } from '../../../k8s/selectors';
 import { Loading } from '../../Loading/Loading';
-import { validateForm, DROPDOWN, POSITIVE_NUMBER } from '../../Form';
+import { DROPDOWN, POSITIVE_NUMBER } from '../../Form';
 
 export class Flavor extends React.Component {
   constructor(props) {
     super(props);
-    this.state = {
-      loadingTemplate: false,
-      template: null,
-    };
     this.resolveInitialValues();
   }
 
@@ -31,46 +27,30 @@ export class Flavor extends React.Component {
     }
   };
 
-  componentDidMount() {
-    this.setState({
-      loadingTemplate: true,
-    });
-    this.props
-      .retrieveVmTemplate()
-      .then(result => {
-        this.props.onFormChange(result, 'template', validateForm(this.flavorFormFields(), this.props.formValues));
-        return this.setState({
-          loadingTemplate: false,
-          template: result,
-        });
-      })
-      .catch(error => {
-        this.props.onLoadError(error.message || 'An error occurred while loading vm flavors. Please try again.');
-        return this.setState({
-          loadingTemplate: false,
-          template: null,
-        });
-      });
-  }
-
   getFlavorChoices = () => {
     const flavors = [];
-    if (this.state.template) {
-      flavors.push(...getTemplateFlavors([this.state.template]));
+    if (this.props.template) {
+      flavors.push(...getTemplateFlavors([this.props.template]));
     }
     if (!flavors.some(flavor => flavor === CUSTOM_FLAVOR)) {
       flavors.push(CUSTOM_FLAVOR);
+    }
+    if (!flavors.some(flavor => flavor === this.props.flavor)) {
+      flavors.push(this.props.flavor);
     }
     return flavors;
   };
 
   flavorFormFields = () => {
-    const { id } = this.props;
+    const { id, flavor } = this.props;
+    const choices = this.getFlavorChoices();
+
     return {
       flavor: {
         id: prefixedId(id, 'flavor-dropdown'),
         type: DROPDOWN,
-        choices: this.getFlavorChoices(),
+        choices,
+        isVisible: () => choices.length > 1 || flavor !== CUSTOM_FLAVOR,
       },
       cpu: {
         id: prefixedId(id, 'flavor-cpu'),
@@ -90,21 +70,30 @@ export class Flavor extends React.Component {
   };
 
   render() {
-    const { id, editing, updating, LoadingComponent, onFormChange, flavor } = this.props;
+    const { id, editing, updating, LoadingComponent, onFormChange, flavor, formValues, vm } = this.props;
     const formFields = this.flavorFormFields();
+    const choices = this.getFlavorChoices();
 
     return (
-      <InlineEdit
-        formFields={formFields}
-        editing={editing}
-        updating={updating || (editing && this.state.loadingTemplate)}
-        LoadingComponent={LoadingComponent}
-        onFormChange={onFormChange}
-        fieldsValues={this.props.formValues}
-      >
-        <div id={prefixedId(id, 'flavor')}>{flavor}</div>
-        <div id={prefixedId(id, 'flavor-description')}>{getFlavorDescription(this.props.vm)}</div>
-      </InlineEdit>
+      <Fragment>
+        {choices.length === 1 && flavor === CUSTOM_FLAVOR && editing ? (
+          <div className="kubevirt-flavor__label">{CUSTOM_FLAVOR}</div>
+        ) : (
+          ''
+        )}
+        <InlineEdit
+          formFields={formFields}
+          editing={editing}
+          updating={updating}
+          LoadingComponent={LoadingComponent}
+          onFormChange={onFormChange}
+          fieldsValues={formValues}
+          showLabels
+        >
+          <div id={prefixedId(id, 'flavor')}>{flavor}</div>
+          <div id={prefixedId(id, 'flavor-description')}>{getFlavorDescription(vm)}</div>
+        </InlineEdit>
+      </Fragment>
     );
   }
 }
@@ -116,10 +105,9 @@ Flavor.propTypes = {
   onFormChange: PropTypes.func.isRequired,
   updating: PropTypes.bool,
   editing: PropTypes.bool,
-  retrieveVmTemplate: PropTypes.func.isRequired,
   LoadingComponent: PropTypes.func,
   formValues: PropTypes.object,
-  onLoadError: PropTypes.func,
+  template: PropTypes.object,
 };
 
 Flavor.defaultProps = {
@@ -128,5 +116,5 @@ Flavor.defaultProps = {
   editing: false,
   LoadingComponent: Loading,
   formValues: undefined,
-  onLoadError: () => {},
+  template: null,
 };
--- a/src/components/Details/Flavor/Flavor.js
+++ b/src/components/Details/Flavor/Flavor.js
@@ -42,7 +42,7 @@ export class Flavor extends React.Component {
   };
 
   flavorFormFields = () => {
-    const { id, flavor } = this.props;
+    const { id } = this.props;
     const choices = this.getFlavorChoices();
 
     return {
@@ -50,7 +50,7 @@ export class Flavor extends React.Component {
         id: prefixedId(id, 'flavor-dropdown'),
         type: DROPDOWN,
         choices,
-        isVisible: () => choices.length > 1 || flavor !== CUSTOM_FLAVOR,
+        isVisible: () => choices.length > 1,
       },
       cpu: {
         id: prefixedId(id, 'flavor-cpu'),
@@ -76,11 +76,7 @@ export class Flavor extends React.Component {
 
     return (
       <Fragment>
-        {choices.length === 1 && flavor === CUSTOM_FLAVOR && editing ? (
-          <div className="kubevirt-flavor__label">{CUSTOM_FLAVOR}</div>
-        ) : (
-          ''
-        )}
+        {choices.length === 1 && editing ? <div className="kubevirt-flavor__label">{flavor}</div> : ''}
         <InlineEdit
           formFields={formFields}
           editing={editing}
--- a/src/components/Details/Flavor/fixtures/Flavor.fixture.js
+++ b/src/components/Details/Flavor/fixtures/Flavor.fixture.js
@@ -9,19 +9,18 @@ export default [
     component: Flavor,
     name: 'Flavor',
     props: {
+      id: 'Flavor',
       vm: cloudInitTestVm,
       flavor: getFlavor(cloudInitTestVm),
       onFormChange: () => {},
-      retrieveVmTemplate: () =>
-        new Promise(resolve => {
-          resolve(fedora28);
-        }),
+      template: null,
     },
   },
   {
     component: Flavor,
     name: 'Flavor editing',
     props: {
+      id: 'Flavor editing',
       vm: cloudInitTestVm,
       flavor: getFlavor(cloudInitTestVm),
       formValues: {
@@ -37,24 +36,19 @@ export default [
       },
       editing: true,
       onFormChange: () => {},
-      retrieveVmTemplate: () =>
-        new Promise(resolve => {
-          resolve(fedora28);
-        }),
+      template: fedora28,
     },
   },
   {
     component: Flavor,
     name: 'Flavor updating',
     props: {
+      id: 'Flavor updating',
       vm: cloudInitTestVm,
       flavor: getFlavor(cloudInitTestVm),
       updating: true,
       onFormChange: () => {},
-      retrieveVmTemplate: () =>
-        new Promise(resolve => {
-          resolve(fedora28);
-        }),
+      template: null,
     },
   },
 ];
--- a/src/components/Details/VmDetails/VmDetails.js
+++ b/src/components/Details/VmDetails/VmDetails.js
@@ -1,7 +1,7 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { get } from 'lodash';
-import { Button, Alert, FieldLevelHelp } from 'patternfly-react';
+import { Button, Alert, FieldLevelHelp, Icon, Tooltip, OverlayTrigger } from 'patternfly-react';
 import classNames from 'classnames';
 
 import { VmStatuses } from '../../VmStatus';
@@ -12,6 +12,7 @@ import {
   getNodeName,
   getOperatingSystem,
   getVmTemplate,
+  getTemplateDisplayName,
   getWorkloadProfile,
   getVmiIpAddresses,
   getOperatingSystemName,
@@ -46,7 +47,9 @@ export class VmDetails extends React.Component {
       editing: false,
       updating: false,
       k8sError: null,
+      templateError: null,
       form: {},
+      template: null,
     };
   }
 
@@ -68,11 +71,6 @@ export class VmDetails extends React.Component {
         },
       },
     }));
-
-  onLoadError = error =>
-    this.setState({
-      k8sError: error,
-    });
 
   updateVmDetails = () => {
     this.setEditing(false);
@@ -131,6 +129,28 @@ export class VmDetails extends React.Component {
     }
   }
 
+  componentDidMount() {
+    const { k8sGet, vm } = this.props;
+    this.setState({
+      updating: true,
+    });
+    retrieveVmTemplate(k8sGet, vm)
+      .then(result => {
+        this.onFormChange('flavor', result, 'template', true);
+        return this.setState({
+          updating: false,
+          template: result,
+        });
+      })
+      .catch(error =>
+        this.setState({
+          updating: false,
+          template: null,
+          templateError: error.message || 'An error occurred. Please try again.',
+        })
+      );
+  }
+
   render() {
     const {
       pods,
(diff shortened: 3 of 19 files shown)


yaacov:
wip until fixing coverage ...

src/components/Details/VmDetails/VmDetails.js
```diff
@@ -123,10 +129,17 @@ export class VmDetails extends React.Component {
...
   componentDidUpdate() {
-    const { launcherPod, importerPods, migration, vm } = this.props;
+    const { launcherPod, importerPods, migration, k8sGet, vm } = this.props;
     if (this.state.editing && !this.isVmOff(vm, launcherPod, importerPods, migration)) {
       this.setEditing(false);
     }
+
+    // Check template, if template is missing, update view.
+    if (getVmTemplate(vm) != null && this.state.templateError === PENDING_STR) {
+      retrieveVmTemplate(k8sGet, vm)
```
suomiy: retrieveVmTemplate is already called for the flavor. Please call this method only once.

yaacov: :+1:

src/components/Details/VmDetails/VmDetails.js
```diff
@@ -123,10 +129,17 @@ export class VmDetails extends React.Component {
...
 
   componentDidUpdate() {
-    const { launcherPod, importerPods, migration, vm } = this.props;
+    const { launcherPod, importerPods, migration, k8sGet, vm } = this.props;
     if (this.state.editing && !this.isVmOff(vm, launcherPod, importerPods, migration)) {
       this.setEditing(false);
     }
+
+    // Check template, if template is missing, update view.
+    if (getVmTemplate(vm) != null && this.state.templateError === PENDING_STR) {
```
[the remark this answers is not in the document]
yaacov: Thanks ! will fix that.

src/components/Details/VmDetails/tests/VmDetails.test.js
```diff
@@ -15,6 +15,8 @@ import { getId } from '../../../../selectors';
 
 const testVmDetails = (vm, otherProps) => <VmDetails {...VmDetailsFixture[0].props} vm={vm} {...otherProps} />;
 
+const flushAllPromises = () => new Promise(resolve => setImmediate(resolve));
```
rawagner: you can reuse https://github.com/kubevirt/web-ui-components/blob/master/src/tests/enzyme.js#L3

yaacov: Nice, thanks :exclamation: :+1:

yaacov:
@rawagner @suomiy please re-review.

I'v removed the part that indicate missing template in the vm details view, because it is not part of the BZ, and may be implemented better in a separate PR.

rawagner:
@yaacov I wonder if its okay to not show anything in UI if we cannot load the template ? We use the loaded template to get possible flavors when editing. Lets consider this flow:
1. I will create VM from template with Small flavor
2. I will delete the template (or somebody else)
3. I will go to VM details - I can see that flavor is Small
4. I will go to Edit mode - flavor dropdown now shows only Custom Flavor and I might get confused why is that so.

@matthewcarleton WDYT ?

src/components/Details/Flavor/Flavor.js
```diff
@@ -44,13 +44,14 @@ export class Flavor extends React.Component {
           template: result,
         });
       })
-      .catch(error => {
-        this.props.onLoadError(error.message || 'An error occurred while loading vm flavors. Please try again.');
-        return this.setState({
+      .catch(error =>
+        // An error occurred while loading template, it's ok, template may be deleted.
+        // Ref: BZ1685393
```
suomiy: is it necessary to include the BZ number?

yaacov: @rawagner , 

> is it necessary to include the BZ number?

No :-) , but it is strange not to do anything on error ... so I needed some excuse.
looking again at @rawagner comment https://github.com/kubevirt/web-ui-components/pull/287#issuecomment-476177652 it may be better to find some way to indicate that the template is missing ... so this will be replaced with actually doing something, and no need for the BZ ref.

yaacov: @suomiy sorry ^^

yaacov:
@rawagner 
> I wonder if its okay to not show anything in UI if we cannot load the template ?

I agree, will look for a way to indicate missing template to the user.

yaacov:
@rawagner @suomiy please re-review

Changes:
moved  `retrieveVmTemplate` one computnent higher from falvor to vm - 
pros:
- knowledge about actual template in etcd is now in the vm sate, and can be used in rendering vm details.
 - vmTemplate that get the template (with it's flavors) as a prop do need to re-read it from etcd when editing it's flavor.
- added tests for vm details with deleted template, and flavor in edit state.

src/components/Details/VmTemplateDetails/VmTemplateDetails.js
```diff
@@ -218,9 +212,8 @@ export class VmTemplateDetails extends React.Component {
                       updating={this.state.updating}
                       LoadingComponent={LoadingComponent}
                       onFormChange={(newValue, key, valid) => this.onFormChange('flavor', newValue, key, valid)}
-                      retrieveVmTemplate={() => retrieveVmTemplate(k8sGet, vmTemplate)}
                       formValues={settingsValue(this.state.form, FLAVOR_KEY)}
-                      onLoadError={this.onLoadError}
+                      template={vmTemplate}
```
suomiy: this component should still retrieve the base template for flavor options

suomiy:
> vmTemplate that get the template (with it's flavors) as a prop do need to re-read it from etcd when editing it's flavor.

not sure this is actually happening

yaacov:
> not sure this is actually happening

Sorry, typo, I meant to write "prop ~do~ **do not** need to re-read it from etcd when editing it's flavo"

src/components/Details/VmDetails/VmDetails.js
```diff
@@ -122,7 +124,33 @@ export class VmDetails extends React.Component {
...
+    retrieveVmTemplate(k8sGet, vm)
+      .then(result =>
+        this.setState({
+          updating: false,
+          template: result,
+        })
+      )
+      .catch(error =>
+        this.setState({
+          updating: false,
```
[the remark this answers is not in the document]
yaacov: make sense to me :+1:

src/components/Details/VmDetails/VmDetails.js
```diff
@@ -122,7 +124,33 @@ export class VmDetails extends React.Component {
 
   isFormValid = () => Object.keys(this.state.form).every(key => this.state.form[key].valid);
 
-  componentDidUpdate() {
+  updateVmTemplate = () => {
+    const { k8sGet, vm } = this.props;
+
+    // if already loading or have template, just return.
+    if (this.state.template || this.state.updating) {
```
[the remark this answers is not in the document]
yaacov: Right :clap: removing

src/components/Details/Flavor/fixtures/Flavor.fixture.js
```diff
@@ -37,24 +39,19 @@ export default [
...
+      id: 'Flavor updating',
       vm: cloudInitTestVm,
       flavor: getFlavor(cloudInitTestVm),
       updating: true,
       onFormChange: () => {},
-      retrieveVmTemplate: () =>
-        new Promise(resolve => {
-          resolve(fedora28);
-        }),
+      template: cloudInitTestVm,
```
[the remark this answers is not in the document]
yaacov: @suomiy 
I think it should be `fedora28` ?
like it was on old line 56 ?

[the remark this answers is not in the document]
suomiy: I mean, it doesn't need a template when in updating state? Nevertheless both should be fine.

src/components/Details/Flavor/Flavor.js
```diff
@@ -31,32 +27,20 @@ export class Flavor extends React.Component {
...
-        this.props.onLoadError(error.message || 'An error occurred while loading vm flavors. Please try again.');
-        return this.setState({
-          loadingTemplate: false,
-          template: null,
-        });
-      });
+  componentDidUpdate(prevProps, prevState) {
+    if (this.props.template && prevProps.template !== this.props.template) {
+      this.props.onFormChange(
+        this.props.template,
```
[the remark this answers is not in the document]
yaacov: Oh... this should be done on the vm side ... will remove `componentDidUpdate` from flavor.

src/components/Details/VmTemplateDetails/VmTemplateDetails.js
yaacov: @suomiy , can we have a recursion of template depending on another template ?

will re-add loading of base template, just to be sure, but not sure if we need to, WDYT ?

src/components/Details/Flavor/fixtures/Flavor.fixture.js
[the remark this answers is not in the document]
yaacov: :+1:

src/components/Details/VmTemplateDetails/VmTemplateDetails.js
suomiy: not real recursion (just the recursion of types (: ) But yes, user vm templates are depending on pre-instaled templates (that's where we get our flavors from).

yaacov: :+1:  Thanks !

matthewcarleton:
> @yaacov I wonder if its okay to not show anything in UI if we cannot load the template ? We use the loaded template to get possible flavors when editing. Lets consider this flow:
> 
> 1. I will create VM from template with Small flavor
> 2. I will delete the template (or somebody else)
> 3. I will go to VM details - I can see that flavor is Small
> 4. I will go to Edit mode - flavor dropdown now shows only Custom Flavor and I might get confused why is that so.
> 
> @matthewcarleton WDYT ?

If I'm understanding this correctly, I agree that we shouldn't show an error if the template is deleted. I think we should show under templates that the template no longer exists. We might need to explore this in the design a bit further. I would say if all we can offer is a custom flavour at that point then we should not offer a dropdown with just one option. It could just show as a heading of "custom" with the fields for CPUs and Memory. Is the idea that the values of the small flavour will now be mapped to a custom set of values?

yaacov:
> think we should show under templates that the template no longer exists. We might need to explore this in the design a bit further.

:+1: 

> I would say if all we can offer is a custom flavour at that point then we should not offer a dropdown with just one option.

Make sense to me, this may be a more global issue, how should we render any dropdown with only one option, IMHO this is also an issue for more exploring.

> Is the idea that the values of the small flavour will now be mapped to a custom set of values?

IFAIK no, IIUC the kubevirt operator running on the server will continue to use values from the now missing template, until we actively  push a change.

suomiy:
> Make sense to me, this may be a more global issue, how should we render any dropdown with only one option, IMHO this is also an issue for more exploring.

we can start with the flavor and then resolve other use cases in the future

yaacov:
@suomiy the testing looks stuck :-( 

> should show under templates that the template no longer exists

a "pficon-warning-triangle-o" with tooltip , WDYT ?

>  how should we render any dropdown with only one option,

make it disabled, WDYT ?

can you re-review current state of code ?

src/components/Details/VmDetails/fixtures/VmDetails.fixture.js
```diff
@@ -87,6 +87,36 @@ export const vmFixtures = {
...
+              requests: {
+                memory: '2G',
+              },
+            },
+          },
+        },
+      },
+      running: false,
+    },
+    editing: true,
```
suomiy: should this line be included in the vm fixture?

src/components/Details/common/tests/details.js
```diff
@@ -58,6 +59,7 @@ export const disablesEditOnCancel = component =>
   awaitVmDetails(() => {
     expect(component.find(InlineFormFactory).exists()).toBeFalsy();
     clickButton(component, 'Edit');
+    component.setState({ updating: false });
```
suomiy: is this because of the waiting on retrieveVmTemplate? Can we rather mock it and wait for the promise to finish?

suomiy:
> @suomiy the testing looks stuck :-(

not sure what is going on, maybe the next push will fix it :)

> > should show under templates that the template no longer exists

> a "pficon-warning-triangle-o" with tooltip , WDYT ?

@matthewcarleton ?
sounds good to me, we could also implement it in this PR 

> > how should we render any dropdown with only one option,

> make it disabled, WDYT ?

> It could just show as a heading of "custom" with the fields for CPUs and Memory.

sounds good, it doesn't have to be a dropdown,

src/components/Details/common/tests/details.js
yaacov: > Can we rather mock it and wait for the promise to finish?

I can, It's already mocked, I tried to wait for the promise and then removed the `async` and `await flushPromises` because I couldn't make it look nice, I can try again, maybe re-factor it this time to look nicer :-)

matthewcarleton:
@suomiy, @yaacov 
This is what I'm thinking. Would we want to show the action related to the VM no longer being available? "This template has been deleted" ? My only question is, if the user is looking to recover that template is that possible? 
Screen Shot 2019-03-27 at 2 14 21 PM
I like the idea of having just a heading for flavour rather than a disabled dropdown. A disabled dropdown doesn't indicate why it's disabled so it could be confusing.

yaacov:
> My only question is, if the user is looking to recover that template is that possible?

AFAIK no, it is not possible to recover deleted user defined templates that user deleted.

matthewcarleton:
@yaacov ok, if that's the case then we wouldn't need any action in the popover. If it were, I was thinking we could offer a recover option.

yaacov:
@mareklibra @suomiy does it make sense to make the template a link, if it exist ?

matthewcarleton:
@yaacov I don't think it's been shown in the design, but I would assume if the template is available it would be a link.

suomiy:
> does it make sense to make the template a link, if it exist ?

I guess yes, for the vmDetail. The template detail not so much, because even though the template exists, the representation of that object  in oc console does not (no generic templates page).

yaacov:
Added a warning triangle when template is missing:

Peek 2019-03-31 14-35

yaacov:
Replace drop-down with a label when only one option is available:

Peek 2019-03-31 16-34

yaacov:
Add an optional template resource link method to vmDetails:

Peek 2019-04-01 11-43

src/components/Details/VmTemplateDetails/VmTemplateDetails.js
```diff
@@ -124,10 +142,10 @@ export class VmTemplateDetails extends React.Component {
   isFormValid = () => Object.keys(this.state.form).every(key => this.state.form[key].valid);
 
   render() {
-    const { vmTemplate, dataVolumes, NamespaceResourceLink, LoadingComponent, k8sGet } = this.props;
+    const { vmTemplate, dataVolumes, NamespaceResourceLink, LoadingComponent } = this.props;
     const id = getId(vmTemplate);
     const vm = selectVm(vmTemplate.objects);
-    const baseTemplate = getVmTemplate(vmTemplate);
+    const baseTemplateLabels = getVmTemplate(vmTemplate);
```
suomiy: please can you rename this method to something like getVmTemplateLabels/getVmTemplateInfo. Or even better rename it to getTemplateDisplayName and reimplement it so that it returns string?

yaacov: sure :smile_cat:

suomiy:
can we add some margin to the custom label so it appears similar to the dropdown ? Not sure if that is necessary, just my opinion. 

mar

src/components/Details/VmTemplateDetails/VmTemplateDetails.js
```diff
@@ -145,7 +145,10 @@ export class VmTemplateDetails extends React.Component {
     const { vmTemplate, dataVolumes, NamespaceResourceLink, LoadingComponent } = this.props;
     const id = getId(vmTemplate);
     const vm = selectVm(vmTemplate.objects);
-    const baseTemplateLabels = getVmTemplate(vmTemplate);
+    const getTemplateDisplayName = () => {
```
suomiy: I meant, can we replace the getVmTemplate method? So it can be also reused in vmDetails.

yaacov: :-)

yaacov:
> can we add some margin to the custom label so it appears similar to the dropdown ?

```
.kubevirt-flavor__label {
   margin-bottom ...
}
```
@suomiy  WDYT ?

yaacov:
@suomiy @matthewcarleton do we need some labels to the inputs in edit mode ?

Peek 2019-04-01 11-43

we curently have:
```
[ flavor drop down ]
[ number input     ]
[ number input     ]
```
should it have labels in edit mode, e.g.
```
[ flavor drop down        ]
Memory [ number input     ]
CPU    [ number input     ]
```
WDYT ?

b - Does `<InlineEdit formFields` has some props for that ?

src/components/Details/VmTemplateDetails/VmTemplateDetails.js
yaacov: I think it's ok now :white_flower:

suomiy:
> ```
> .kubevirt-flavor__label {
>   margin-bottom ...
> }
> ```
> @suomiy WDYT ?

yup, I would make the margin similar to the dropdown

> [ flavor drop down        ]
> Memory [ number input     ]
> CPU    [ number input     ]
>
> WDYT ?

> b - Does <InlineEdit formFields has some props for that ?

+1, that is a good remark. I guess it has to be some regression with InlineFormFactory, because It already has a title. Can you look at it?

yaacov:
> I guess it has to be some regression with InlineFormFactory, because It already has a title. Can you look at it?

Sure :+1:

src/components/Details/VmTemplateDetails/VmTemplateDetails.js
```diff
@@ -124,10 +143,10 @@ export class VmTemplateDetails extends React.Component {
   isFormValid = () => Object.keys(this.state.form).every(key => this.state.form[key].valid);
 
   render() {
-    const { vmTemplate, dataVolumes, NamespaceResourceLink, LoadingComponent, k8sGet } = this.props;
+    const { vmTemplate, dataVolumes, NamespaceResourceLink, LoadingComponent } = this.props;
     const id = getId(vmTemplate);
     const vm = selectVm(vmTemplate.objects);
-    const baseTemplate = getVmTemplate(vmTemplate);
+    const baseTemplateLabels = getTemplateDisplayName(getVmTemplate(vmTemplate));
```
suomiy: nit: variable name

yaacov: sorry ... `baseTemplateName` ?

suomiy: np, yup could be or baseTemplateDisplayName

yaacov:
show labels in inline form:

Peek 2019-04-01 14-46

yaacov:
@suomiy this is nearing the end :tada:  :balloon: , should I squash, or leave separate commits for easier review ?

matthewcarleton:
@yaacov this looks great, thanks! One question, what happens when the available CPU/Memory is exceeded?

yaacov:
> what happens when the available CPU/Memory is exceeded?

@matthewcarleton 

TL;DR - I do not know.

IFAIK we do not have the `quota/total` values to validate the form ... @suomiy do you know something about this ? 

p.s.
What I think happen when values are bad is that openshift will try to execute and fail after some tries ... @suomiy - do you know something about that ?

yaacov:
@suomiy @rawagner  please review ... I think it's ready :beers:

suomiy:
> > > what happens when the available CPU/Memory is exceeded?
> 
> @matthewcarleton
> 
> TL;DR - I do not know.
> 
> IFAIK we do not have the quota/total values to validate the form ... @suomiy do you know > something about this ?
> 
> p.s.
> What I think happen when values are bad is that openshift will try to execute and fail after >some tries ... @suomiy - do you know something about that ?

Currently, we do not have a way to know the quota AFAIK. The request will go ok, just the VM will fail with an error the next time you try to start it.

suomiy:
> @suomiy this is nearing the end tada  balloon , should I squash, or leave separate commits for easier review ?

don't worry about that. It will be squashed on merge

suomiy:
> .kubevirt-flavor__label {
>   margin-bottom ...
> } 

is this present in this PR?

yaacov:
>    .kubevirt-flavor__label {
>    margin-bottom ...
>    }
>
> is this present in this PR?

Aaa... forgot to do `git add` :100:

matthewcarleton:
@suomiy ah ok. Ya it would be a nicer experience for sure to know what the limit is at the time. Would it be possible to explore adding that feature (in another PR :) )

suomiy:
> Would it be possible to explore adding that feature (in another PR :) )

I think this should be explored on the backend. Doesn't seem like we can do much about it here in the UI.

@jelkosz Do you have some additional information on this issue?

suomiy:
looks good, merging

PR #289, opened 2019-03-25
Refactor TopConsumers and add Workload and Infrastructure metrics

rawagner:


sass/components/ClusterOverview/consumers.scss
```diff
@@ -1,10 +1,52 @@
...
+.kubevirt-consumers__bar-column-progress {
+  padding: 0;
+}
+
+.kubevirt-consumers__bar-description {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  color: #030303;
+  font-size: 12px;
```
suomiy: can we use em units instead?

src/components/ClusterOverview/TopConsumers/ClusterTopConsumers.js
```diff
@@ -0,0 +1,132 @@
...
+import { formatBytes as utilFormatBytes } from '../../../utils';
+
+const formatCpu = (cpuTime, fixed = 0) => Number(cpuTime).toFixed(fixed);
+
+const formatBytes = memory => {
+  const bytes = utilFormatBytes(memory);
+  return `${bytes.value} ${bytes.unit}`;
+};
+
+const getConsumers = (kind, results, nameLabel, formatLabel) => {
```
suomiy: maybe this would better fit to selectors

src/components/ClusterOverview/TopConsumers/TopConsumers.js
```diff
@@ -14,83 +12,141 @@ import {
...
-  const max = Math.max(...filteredConsumers.map(c => c.usage));
-  return filteredConsumers
+ConsumersBar.propTypes = {
+  now: PropTypes.number.isRequired,
+  description: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+};
+
+const getConsumers = metric => {
+  const max = Math.max(...metric.consumers.map(c => c.usage));
```
suomiy: now value will be always relative to the highest consumers, which I don't think is very useful.

rawagner: after discussion with @jelkosz this is what we want. The first consumer bar is 100% and others are relative to the first one.

sass/components/ClusterOverview/consumers.scss
rawagner: using em now

src/components/ClusterOverview/TopConsumers/ClusterTopConsumers.js
rawagner: moved to `selectors/prometheus/consumers`

sass/components/ClusterOverview/consumers.scss
```diff
@@ -1,10 +1,51 @@
-.kubevirt-consumers__bar .progress {
+.kubevirt-consumers__bar {
+  height: 20px;
```
suomiy: the em units are preferable in other places as well. We should probably convert the whole dashboard to be consistent with this.

src/components/ClusterOverview/TopConsumers/TopConsumers.js
suomiy: I talked to @jelkosz and we concluded that it would be good to include some icon with a popover that would explain what is happening. E.g. `something something, The value xy is relative to the current higest value (100% bar)`. It could also include something about the query.

rawagner: created a followup https://github.com/kubevirt/web-ui-components/issues/313 - I will ask UXD for help

sass/components/ClusterOverview/consumers.scss
rawagner: fixed

rawagner:
@suomiy @mareklibra can you please review ?

src/components/ClusterOverview/TopConsumers/TopConsumers.js
```diff
@@ -1,100 +1,188 @@
...
-        ))}
-      </div>
+      <DashboardCard>
+        <DashboardCardHeader>
+          <DashboardCardTitle>Top Consumers</DashboardCardTitle>
+          <DashboardCardTitleHelp>help for top consumers</DashboardCardTitleHelp>
+        </DashboardCardHeader>
+        <DashboardCardBody>
+          <ConsumersFilter>
+            {getFormElement(typeDropdown)}
```
suomiy: would be maybe good to add some margin between them

src/components/ClusterOverview/TopConsumers/TopConsumers.js
```diff
@@ -1,100 +1,188 @@
...
-          {getFormElement(sortByDropdown)}
-        </div>
-        {sortConsumers(this.props.metrics, this.state.filter, this.state.sortBy).map((c, index) => (
-          <UtilizationBar key={index} {...c} className="kubevirt-consumers__bar" />
-        ))}
-      </div>
+      <DashboardCard>
+        <DashboardCardHeader>
+          <DashboardCardTitle>Top Consumers</DashboardCardTitle>
+          <DashboardCardTitleHelp>help for top consumers</DashboardCardTitleHelp>
```
suomiy: re: https://github.com/kubevirt/web-ui-components/issues/313. For now all the metrics behave the same, so we could add the explanation text here until #313 is resolved.

src/components/ClusterOverview/TopConsumers/TopConsumers.js
```diff
@@ -1,100 +1,188 @@
...
+} from './strings';
+import { ConsumersFilter } from '../../Dashboard/TopConsumers/ConsumersFilter';
 
-const sortConsumers = (metrics, filter, sortBy) => {
-  const metricKey = Object.keys(metrics).find(key => metrics[key].title === sortBy);
-  const filteredConsumers =
-    filter !== 'All' ? metrics[metricKey].consumers.filter(c => c.kind === filter) : metrics[metricKey].consumers;
-  const max = Math.max(...filteredConsumers.map(c => c.usage));
-  return filteredConsumers
+const getMetricConsumers = ({ consumers }) => {
```
suomiy: this should probably go to selectors or some util class (Dashboard/TopConsumers/utils or some other utils?)

src/components/ClusterOverview/TopConsumers/TopConsumers.js
rawagner: added

space

src/components/ClusterOverview/TopConsumers/TopConsumers.js
rawagner: `getMetricConsumers` moved to Dashboard/TopConsumers/utils. The other two methods were moved to ClusterOverview/TopConsumers/utils

src/components/ClusterOverview/TopConsumers/TopConsumers.js
rawagner: Added `Consumption bar is relative to biggest consumer which is always 100%`  Not sure if that is good enough

sass/components/ClusterOverview/consumers.scss
```diff
@@ -1,10 +1,3 @@
...
-  height: 10px;
-  margin-top: 15px;
-}
-
-.kubevirt-consumers__filters {
-  display: flex;
-  padding-bottom: 10px;
-  border-bottom: 1px solid #d1d1d1;
+.kubevirt-consumers__dropdown-filter {
+  width: 49%;
```
suomiy: I think margin should be more descriptive/generic than width

src/components/ClusterOverview/TopConsumers/strings.js
```diff
@@ -0,0 +1,17 @@
+export const BY_CPU = 'By CPU';
+export const CPU_DESC = 'CPU Time';
```
suomiy: I think it would make more sense to make a mapping of that

src/components/ClusterOverview/TopConsumers/utils.js
```diff
@@ -0,0 +1,14 @@
+import { formatBytes as utilFormatBytes } from '../../../utils';
+import { getConsumers } from '../../../selectors';
+
+export const formatCpu = (cpuTime, fixed = 0) => Number(cpuTime).toFixed(fixed);
+
+export const formatBytes = memory => {
```
suomiy: probably better to think of a different name to not clash with a `formatBytes` from utils

sass/components/ClusterOverview/consumers.scss
rawagner: Im using Columns + padding as I had some issues on firefox when not using width.

src/components/ClusterOverview/TopConsumers/utils.js
rawagner: named updated

src/components/ClusterOverview/TopConsumers/strings.js
rawagner: mapping is done in https://github.com/kubevirt/web-ui-components/pull/289/files#diff-2d8caae240c01140ce53da1f24c6cdb4R37  and https://github.com/kubevirt/web-ui-components/pull/289/files#diff-2d8caae240c01140ce53da1f24c6cdb4R39 or do you mean some other mapping ? Would it bring any additional value ?

PR #293, opened 2019-03-27
Add Host Status selector and components

flofuchs:


src/selectors/host/selectors.js
```diff
@@ -1,3 +1,44 @@
...
+    // - Deleting
+    
+    let hostStatus = '-';
+
+    // A series of checks to determine the status.
+    // Each check must either return a string or false.
+    // Added in ascending order of importance.
+    const statusChecks = [
+      // Check provisioning status
+      host => !!get(host, ['status', 'provisioning', 'ID'], false) ? 'Provisioning' : false
```
honza: Maybe use a `_.has()`?

```js
has(host, ['status', 'provisioning', 'ID']) ? 'Provisioning' : false
```

Also a small irritation about returning two different types (string vs bool). :smirk:

honza:
I think it needs a rebase on top of master.

src/components/BareMetalHosts/BaremetalHostStatus.js
```diff
@@ -0,0 +1,32 @@
...
+  ready: GenericSuccess,
+  provisioned: GenericSuccess,
+  provisioning: GenericProgress,
+  'preparing to provision': GenericProgress,
+  deprovisioning: GenericProgress,
+  'making host available': GenericProgress,
+};
+
+export const BaremetalHostStatus = ({ host }) => {
+  const hostStatus = getHostStatus(host);
```
suomiy: we found out that for better reusability it is better to decouple status functionality from the UI (e.g. vmStatus is used at many places). Please see utils/status. Could you use this approach here?

Basically it is a function which returns
```js
{
   status, // constant
   ...additionalUsefulData // e.g. message
}
```

src/selectors/host/selectors.js
```diff
@@ -1,3 +1,49 @@
 import { get } from 'lodash';
 
-export const getHostStatus = host => get(host, ['metadata', 'labels', 'metalkube.org/operational-status']);
+// Status titles. Could be i18n'ed at some point.
+const statusTextMap = {
```
suomiy: + the string mapping could be done there

src/components/BareMetalHosts/StatusComponents.js
```diff
@@ -0,0 +1,69 @@
...
+  ) : (
+    <IconAndText icon="pficon-error-circle-o" text={getHostStatusText(host)} />
+  );
+
+GenericError.propTypes = {
+  host: PropTypes.object.isRequired,
+};
+
+// Generic status component as a fallback
+export const GenericStatus = ({ host }) => <React.Fragment>{getHostStatusText(host)}</React.Fragment>;
```
suomiy: why the `React.Fragment` ?

src/components/BareMetalHosts/BaremetalHostStatus.js
jtomasek: I agree we could make the textMap and componentMap driven by constants. We can also enhance the getHostStatus selector to return an object as you're describing, which would allow us to include error message etc. and the UI components would not have to reach for it separately.

Would it be ok to do this in a subsequent PR? Quite a lot of work depends on this one so we'd like to get this one in.

suomiy: yes. Also this should go to `utils/status`. Selectors should be kept simplest as possible.

I would rather do it in this PR. Especially because other work depends on it. 
It is not a time consuming task and will lead to (hopefully) less bugs.

src/components/BareMetalHosts/StatusComponents.js
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Popover } from '@patternfly/react-core';
+
+import { getHostStatusText } from '../../selectors/host/selectors';
```
jtomasek: can be shortened to
`import { getHostStatusText } from '../../selectors';`

src/components/BareMetalHosts/StatusComponents.js
```diff
@@ -0,0 +1,69 @@
...
+
+export const IconAndText = ({ icon, text }) => {
+  const className = `pficon ${icon} co-icon-and-text__icon`;
+  return (
+    <span className="co-icon-and-text">
+      <span aria-hidden="true" className={className} />
```
jtomasek: Icon component is available from patternfly
`<Icon type="pf" name="error-circle-o" />`

src/utils/status/host/constants.js
```diff
@@ -1,5 +1,74 @@
 export const HOST_STATUS_READY = 'HOST_STATUS_READY';
+export const HOST_STATUS_DISCOVERED = 'HOST_STATUS_DISCOVERED';
```
jtomasek: I was thinking more about defining the states as constants this way:
`export const HOST_STATUS_VALIDATION_ERROR = 'validation error'`...

And then
```
export const statusTextMap = {
 [HOST_STATUS_VALIDATION_FAILED]: 'Validation Error(s)',
 ...
}
```

And getStatusText doing:
```
return statusTextMap[hostStatus] || hostStatus
```

src/selectors/host/selectors.js
```diff
@@ -1,3 +1,35 @@
...
+   * - Starting maintenance
+   * - Maintenance
+   * - Stopping maintenance
+   * - Deleting
+   */
+
+  const operationalStatus = get(host, 'status.operationalStatus');
+  const provisioningState = get(host, 'status.provisioning.state');
+
+  return RESPONSE_TO_HOST_STATUS[provisioningState || operationalStatus];
```
jtomasek: see below, this can be `return provisioningState || operationalStatus;` with below changes

src/utils/status/host/constants.js
```diff
@@ -1,5 +1,74 @@
...
+export const HOST_STATUS_STOPPING_MAINTENANCE = 'HOST_STATUS_STOPPING_MAINTENANCE';
+export const HOST_STATUS_MAINTENANCE = 'HOST_STATUS_MAINTENANCE';
+
+export const HOST_STATUS_ALL_ERROR = [HOST_STATUS_REGISTRATION_ERROR, HOST_STATUS_VALIDATION_ERROR];
 
-// TODO(honza): fill out these values
-export const HOST_STATUS_ALL_ERROR = [];
 export const HOST_STATUS_ALL_WARN = [];
+
+export const HOST_STATUS_ALL_PROCESS = [
```
honza: I think this should be `HOST_STATUS_ALL_IN_PROGRESS`.

src/utils/status/host/constants.js
```diff
@@ -1,5 +1,74 @@
...
+  registering: HOST_STATUS_REGISTERING,
+  'registration error': HOST_STATUS_REGISTRATION_ERROR,
+  'making host available': HOST_STATUS_MAKING_HOST_AVAILABLE,
+  provisioning: HOST_STATUS_PROVISIONING,
+  deprovisioning: HOST_STATUS_DEPROVISIONING,
+  provisioned: HOST_STATUS_PROVISIONED,
+};
+
+export const HOST_STATUS_TO_TEXT = {
+  HOST_STATUS_READY: 'Ready',
```
suomiy: you should do this if you want to use the constant value: 

```js
{
  [HOST_STATUS_READY]: 'Ready',
}
```

src/components/BareMetalHosts/StatusComponents.js
```diff
@@ -0,0 +1,67 @@
...
+  </span>
+);
+
+IconAndText.propTypes = {
+  icon: PropTypes.string.isRequired,
+  text: PropTypes.string.isRequired,
+};
+
+// Generic success status component
+export const GenericSuccess = ({ host }) => <IconAndText icon="pficon-ok" text={getHostStatusText(host)} />;
```
jtomasek: Icon component takes the name of the icon, which is the class without the pficon/fa prefix, so you'll have to update the icon prop:
`<IconAndText icon="ok" text={getHostStatusText(host)} />;`

Same in other IconAndText usages

src/components/BareMetalHosts/StatusComponents.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Popover } from '@patternfly/react-core';
+import { Icon } from 'patternfly-react';
+
+import { getHostStatusText } from '../../selectors';
+
+export const IconAndText = ({ icon, text }) => (
+  <span className="co-icon-and-text">
```
rawagner: can you create a new style which conforms to BEM ?

flofuchs: I've re-used the `.co-icon-and-text__icon` class to add the margin on the right.

src/components/BareMetalHosts/StatusComponents.js
flofuchs: Done

src/utils/status/host/constants.js
flofuchs: Done

src/utils/status/host/constants.js
flofuchs: Thanks!

src/selectors/host/selectors.js
flofuchs: Done

src/utils/status/host/constants.js
flofuchs: Done

src/components/BareMetalHosts/StatusComponents.js
flofuchs: Done

src/components/BareMetalHosts/StatusComponents.js
flofuchs: Done

src/components/BareMetalHosts/StatusComponents.js
flofuchs: Thanks! Wasn't necessary. Removed it.

src/components/BareMetalHosts/StatusComponents.js
jtomasek: I have checked and using wrapping span with `co-icon-and-text` class has no impact on styling so it can be removed completely.

For the icon itself, you can keep the className which adds the margin or remove the className and just add space between Icon and text which does the same.

`<Icon type="pf" name={icon} /> {text}`

src/selectors/host/selectors.js
```diff
@@ -1,3 +1,16 @@
...
 
-export const getHostStatus = host => get(host, ['metadata', 'labels', 'metalkube.org/operational-status']);
+import { HOST_STATUS_TO_TEXT } from '../../utils/status/host/constants';
+
+export const getHostStatus = host => {
+  // Returns a status string based on the available host information.
+  const operationalStatus = get(host, 'status.operationalStatus');
+  const provisioningState = get(host, 'status.provisioning.state');
+
+  return [provisioningState || operationalStatus];
```
jtomasek: I think the brackets are not needed here.

src/selectors/host/selectors.js
```diff
@@ -1,3 +1,16 @@
...
+};
+
+export const getHostStatusText = host => {
+  const hostStatus = getHostStatus(host);
+  return HOST_STATUS_TO_TEXT[hostStatus] ? HOST_STATUS_TO_TEXT[hostStatus] : hostStatus;
```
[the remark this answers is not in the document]
flofuchs: Done

src/selectors/host/selectors.js
flofuchs: Done, thanks

src/components/BareMetalHosts/StatusComponents.js
flofuchs: The class creates a margin that is a tiny bit wider than a space, so I kept it to keep it consistent with other tables in the console.

src/components/BareMetalHosts/StatusComponents.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Popover } from '@patternfly/react-core';
+import { Icon } from 'patternfly-react';
+
+import { getHostStatusText } from '../../selectors';
+
+export const IconAndText = ({ icon, text }) => (
+  <React.Fragment>
+    <Icon type="pf" name={icon} className="co-icon-and-text__icon" />
```
rawagner: please rename this class to `kubevirt-icon-and-text` and define required margin there

src/components/BareMetalHosts/StatusComponents.js
```diff
@@ -0,0 +1,67 @@
...
+export const GenericSuccess = ({ host }) => <IconAndText icon="ok" text={getHostStatus(host).text} />;
+
+GenericSuccess.propTypes = {
+  host: PropTypes.object.isRequired,
+};
+
+// Generic error status component
+// If the errorMessage property isn't empty its contents are
+// shown in a Popover.
+export const GenericError = ({ host }) =>
```
jtomasek: To avoid multiple calls of the selector, the StatusComponents can take just host status object as prop which you can spread as `({ status, text, errorMessage }) => ...`

flofuchs: Done

src/components/BareMetalHosts/BaremetalHostStatus.js
```diff
@@ -0,0 +1,46 @@
...
+  HOST_STATUS_PREPARING_TO_PROVISION,
+  HOST_STATUS_PROVISIONING,
+  HOST_STATUS_DEPROVISIONING,
+  HOST_STATUS_MAKING_HOST_AVAILABLE,
+  HOST_STATUS_VALIDATION_ERROR,
+  HOST_STATUS_REGISTRATION_ERROR,
+} from '../../utils/status/host/constants';
+
+// Map status strings to components
+const componentMap = {
```
suomiy: why is this not the same as `HOST_STATUS_ALL_SUCCESS, HOST_STATUS_ALL_PROGRESS` ? Can we use these? 

If not, can you implement one of these options?

1. move this logic to `host/hostStatus.js` ? and return something like 

```
{
 ...start,
 genericStatus: HOST_GENERIC_STATUS_SUCCESS,
}
```
from there?

\+ create constants for progress/success/validationError, etc. ?  And then do the mapping of these constants here? Ik we would still neeed to do the mapping here but it would be less verbose and more extensible.

2. option is to create another array something like `HOST_GENERIC_STATUS_ALL_SUCCESS`

PR #296, opened 2019-03-28
Fix vmStatus ordering

pcbailey:
This PR changes the ordering of VM statuses in vmStatus to check importing status before running status to avoid the VM reflecting a status of "Off" when it is actually in "Importing" status.

### DIFF
--- a/src/utils/status/vm/vmStatus.js
+++ b/src/utils/status/vm/vmStatus.js
@@ -132,7 +132,8 @@ export const getVmStatus = (vm, pods, migrations, importerPods = null) => {
   const migration = findVMIMigration(migrations, vm);
   const vmImporterPods = getVmImporterPods(importerPods || pods, vm);
   return (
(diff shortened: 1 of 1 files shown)


src/utils/status/vm/vmStatus.js
```diff
@@ -138,12 +138,12 @@ const isWaitingForVmi = vm => {
 };
 
 export const getVmStatus = (vm, launcherPod, importerPods, migration) =>
-  isBeingMigrated(vm, migration) || // must be precceding isRunning() since vm.status.ready is true for a migrating VM
+  isBeingMigrated(vm, migration) || // must precede isRunning() since vm.status.ready is true for a migrating VM
+  isBeingImported(vm, importerPods) || // must precede isRunning() since importing status doesn't rely on running status
```
suomiy: @mareklibra can you please double check the order here?

PR #301, opened 2019-04-02
Refactor Capacity Card and extract CapacityBody and CapacityItem to D…

rawagner:
…ashboard folder

### DIFF
--- a/src/components/Dashboard/Capacity/CapacityItem.js
+++ b/src/components/Dashboard/Capacity/CapacityItem.js
@@ -1,0 +1,90 @@
+import React from 'react';
+import { DonutChart } from 'patternfly-react';
+import PropTypes from 'prop-types';
+
+import { InlineLoading } from '../../Loading';
+import { prefixedId } from '../../../utils';
+import { USED, AVAILABLE, LOADING, CAPACITY, UNKNOWN, NOT_AVAILABLE } from './strings';
+
+const pfGetUtilizationDonutTooltipContents = d =>
+  `<span class="donut-tooltip-pf" style="white-space: nowrap;">${Math.round(d[0].ratio * 1000) / 10} ${
+    d[0].name
+  }</span>`;
+
+export const CapacityItem = ({ id, title, used, total, unit, formatValue, LoadingComponent, isLoading }) => {
+  let description;
+  let columns = [[USED, 0], [AVAILABLE, 100]];
+  let tooltip;
+  let primaryTitle;
+  let secondaryTitle = CAPACITY;
+  if (isLoading) {
+    description = <LoadingComponent />;
+    primaryTitle = LOADING;
+  } else if (used == null || total == null) {
+    description = NOT_AVAILABLE;
+    primaryTitle = UNKNOWN;
+  } else {
+    const totalFormatted = formatValue(total, unit);
+    const usedFormatted = formatValue(used, totalFormatted.unit);
+
+    unit = totalFormatted.unit; // eslint-disable-line prefer-destructuring
+    total = totalFormatted.value;
+    used = usedFormatted.value;
+
+    const available = total - used;
+    columns = [[USED, used], [AVAILABLE, available]];
+
+    const percentageUsed = total > 0 ? Math.round((100 * used) / total) : 0;
+    const availableNice = Number.isInteger(available) ? available : available.toFixed(1);
+    const totalNice = Number.isInteger(total) ? total : total.toFixed(1);
+
+    description = `${availableNice} ${unit} available out of ${totalNice} ${unit}`;
+    tooltip = { contents: pfGetUtilizationDonutTooltipContents };
+    primaryTitle = `${percentageUsed.toString()}%`;
+    secondaryTitle = USED;
+  }
+
+  return (
+    <div className="kubevirt-capacity__item">
+      <h2 className="kubevirt-capacity__item-title">{title}</h2>
+      <h6>{description}</h6>
+      <div>
+        <DonutChart
+          id={prefixedId('donut-chart', id)}
+          size={{ width: 150, height: 150 }}
+          data={{
+            unload: true,
+            columns,
+            order: null,
+          }}
+          unloadBeforeLoad
+          tooltip={tooltip}
+          title={{
+            primary: primaryTitle,
+            secondary: secondaryTitle,
+          }}
+        />
+      </div>
+    </div>
+  );
+};
+
+CapacityItem.defaultProps = {
+  formatValue: (value, unit) => ({ value, unit }),
+  unit: undefined,
+  LoadingComponent: InlineLoading,
+  used: null,
+  total: null,
+  isLoading: false,
+};
+
+CapacityItem.propTypes = {
+  id: PropTypes.string.isRequired,
+  title: PropTypes.string.isRequired,
+  used: PropTypes.number,
+  total: PropTypes.number,
+  unit: PropTypes.string,
+  formatValue: PropTypes.func,
+  LoadingComponent: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
+  isLoading: PropTypes.bool,
+};
--- a/src/selectors/prometheus/selectors.js
+++ b/src/selectors/prometheus/selectors.js
@@ -1,0 +1,21 @@
+import { get } from 'lodash';
+
+import { parseNumber } from '../../utils';
+
+export const getConsumers = (kind, results, nameLabel, formatLabel) => {
+  if (!results) {
+    return null;
+  }
+  const result = get(results, 'data.result');
+  return result.map(r => ({
+    kind,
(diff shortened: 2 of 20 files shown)


rawagner:
@gnehapk @cloudbehl FYI

src/components/Dashboard/Capacity/CapacityItem.js
```diff
@@ -0,0 +1,102 @@
...
+);
+
+LoadingCapacityItem.propTypes = {
+  idSuffix: PropTypes.string.isRequired,
+  title: PropTypes.string.isRequired,
+  LoadingComponent: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+};
+
+export const CapacityItem = ({ idSuffix, title, used, total, unit, formatValue, LoadingComponent }) => {
+  if (!Number.isNaN(parseFloat(used)) && !Number.isNaN(parseFloat(total))) {
```
suomiy: this component probably shouldn't do the parsing and expect numbers instead. We could maybe do this in `getCapacityStats`. Then we can check for null here

src/components/Dashboard/Capacity/CapacityItem.js
```diff
@@ -0,0 +1,102 @@
...
+
+import { InlineLoading } from '../../Loading';
+
+const pfGetUtilizationDonutTooltipContents = d =>
+  `<span class="donut-tooltip-pf" style="white-space: nowrap;">${Math.round(d[0].ratio * 1000) / 10} ${
+    d[0].name
+  }</span>`;
+
+const LoadingCapacityItem = ({ idSuffix, title, LoadingComponent }) => (
+  <div className="kubevirt-capacity__item">
```
suomiy: this layout is already present in `CapacityItem`. Can you please put some ifs there and do the loading design there? This way the layout would be duplicated and harder to modify in the future.

src/components/Dashboard/Capacity/CapacityItem.js
```diff
@@ -0,0 +1,102 @@
...
+    const availableNice = Number.isInteger(available) ? available : available.toFixed(1);
+    const totalNice = Number.isInteger(total) ? total : total.toFixed(1);
+
+    return (
+      <div className="kubevirt-capacity__item">
+        <h2 className="kubevirt-capacity__item-title">{title}</h2>
+        <h6>{`${availableNice} ${unit} available out of ${totalNice} ${unit}`}</h6>
+        <div>
+          <DonutChart
+            id={`donut-chart-${idSuffix}`}
```
suomiy: I get that the idSuffix is useful (lazy approach). But I would prefer the id part was configurable as well. Can you either create idPrefix prop with `donut-chart` value as a default or add just an id prop which is more conventional?

suomiy: also there is a function called `prefixedId`

src/components/Dashboard/Capacity/CapacityItem.js
rawagner: parsing moved to `getCapacityStats` and `CapacityItem` is checking for null

src/components/Dashboard/Capacity/CapacityItem.js
rawagner: done

src/components/Dashboard/Capacity/CapacityItem.js
rawagner: prop renamed to `id` and Im using `prefixedId` now

src/selectors/prometheus/selectors.js
```diff
@@ -0,0 +1,22 @@
...
+    label: formatLabel ? formatLabel(r.value[1]) : r.value[1],
+  }));
+};
+
+export const getCapacityStats = response => {
+  const value = get(response, 'data.result[0].value[1]');
+  if (Number.isNaN(parseFloat(value))) {
+    return null;
+  }
+  return Number(value);
```
suomiy: Do we expect whitespaces in the value? Number is more stronger so it could still return `NaN`. We can just use the parseFloat value if we want to be more lenient.

src/components/Dashboard/Capacity/CapacityItem.js
```diff
@@ -0,0 +1,82 @@
...
+
+    const percentageUsed = total > 0 ? Math.round((100 * used) / total) : 0;
+    const availableNice = Number.isInteger(available) ? available : available.toFixed(1);
+    const totalNice = Number.isInteger(total) ? total : total.toFixed(1);
+
+    description = `${availableNice} ${unit} available out of ${totalNice} ${unit}`;
+    tooltip = { contents: pfGetUtilizationDonutTooltipContents };
+    primaryTitle = `${percentageUsed.toString()}%`;
+    secondaryTitle = USED;
+  }
```
suomiy: please move at least the ` description = <LoadingComponent />` to else branch

src/components/Dashboard/Capacity/CapacityItem.js
```diff
@@ -0,0 +1,82 @@
...
+    d[0].name
+  }</span>`;
+
+export const CapacityItem = ({ id, title, used, total, unit, formatValue, LoadingComponent }) => {
+  let description = <LoadingComponent />;
+  let columns = [[USED, 0], [AVAILABLE, 100]];
+  let tooltip;
+  let primaryTitle = LOADING;
+  let secondaryTitle = CAPACITY;
+  if (used != null && total != null) {
```
suomiy: hmm, on the secod look, I would rather pass isLoading from above and check  that we already got the response (presence of response.data? in ClusterOverview/Capacity). This way the component can distinguish between three cases.

- loading (no response - pass isLoading boolean)
- loaded with wrong response (used/total is null) - we could show `Not Available` label for example
- loaded correctly (show the data)

suomiy: not sure about the prometheus format when there is no data yet or if prometheus can return invalid response. 

@rawagner thoughts?

src/components/Dashboard/Capacity/CapacityItem.js
```diff
@@ -0,0 +1,82 @@
...
+    secondaryTitle = USED;
+  }
+
+  return (
+    <div className="kubevirt-capacity__item">
+      <h2 className="kubevirt-capacity__item-title">{title}</h2>
+      <h6>{description}</h6>
+      <div>
+        <DonutChart
+          id={prefixedId('donut-chart', id)}
```
suomiy: I just didn't want to force the `donut-chart` part on the consumer. Nevertheless it probably doesn't matter that much.

src/components/Dashboard/Capacity/CapacityItem.js
rawagner: @suomiy I think that `no data` can happen if prometheus didnt scrape the endpoint yet. Not sure about the invalid response. Anyways I will add `Not Available` state too.

src/selectors/prometheus/selectors.js
rawagner: Not sure about whitespaces...i would say no and stick with the Number for now

src/components/Dashboard/Capacity/CapacityItem.js
rawagner: added isLoading prop and Not Available state.
capacity

src/components/Dashboard/Capacity/CapacityItem.js
rawagner: moved

src/selectors/prometheus/selectors.js
suomiy: I just saw https://stackoverflow.com/questions/12227594/which-is-better-numberx-or-parsefloatx. Each one of them has it's quirks. Can we check for NaN of both `parseFloat` and `Number` ?

maybe it would be worth to also make a generic parsing function out of it. It would return number or null only.

mareklibra: I think a number-like value is returned, not a string

suomiy: ok, still better to check against `Number('') === 0`

rawagner: new function in `utils` - `parseNumber` is using combination of `parseFloat`/`Number`/`Number.isNaN` . WDYT ?

suomiy: +1

src/utils/utils.js
```diff
@@ -20,6 +20,8 @@ export const parseUrl = url => {
   }
 };
 
+export const parseNumber = value => (!Number.isNaN(parseFloat(value)) && !Number.isNaN(Number(value)) ? Number(value) : null);
```
suomiy: we can reuse the result of Number(value)

rawagner: fixed

Issue #276, opened 2019-03-19
ClusterOverview - Implement fallback logic when prometheus is not available


rawagner:
@lizsurette

Any idea what to show when prometheus is not available ? 

Basically two cases can happen:
Overview card requires prometheus only for some information
ie Details tab - openshift version is detected via promeheus
![prom_na](https://user-images.githubusercontent.com/2078045/54603766-73f2c680-4a45-11e9-9a46-a4ed639c1484.png)
should we not show the field at all or show something like `prometheus not available` as field value?

Second case would be that all informations are detected via prometheus - Cluster Capacity
![prom_na2](https://user-images.githubusercontent.com/2078045/54603881-bfa57000-4a45-11e9-9ff0-e61c97932956.png)

should we hide the card or keep it visible and again show something like `prometheus not available` ?

lizsurette:
> Any idea what to show when prometheus is not available ?

How about an informational icon with "Data Not Available" as an initial implementation?

Here is a very basic suggestion from PatternFly 3 cards view:
https://www.patternfly.org/pattern-library/cards/base-card/#design

And how oVirt is making use of something similar in their Dashboard when a VM is suspended:
https://github.com/lizsurette/ovirt-design/blob/aa17c7ba2b46a3cb1440f86337992404680072e4/user-portal/img/vm-details-suspended.png

Any thoughts @andybraren?

andybraren:
In the future when longer Details cards are implemented (see examples below) I'd worry that including the icon could make it harder to see other (more important) icons. I'd suggest just "Not available" text without an icon, mocked up below.

To prevent the Capacity card from jumping around too much, I think all parts of the card should stay visible but we replace the values with "Not available" or "Unknown". Thoughts on this @lizsurette @mceledonia?

![concept-no-prometheus](https://user-images.githubusercontent.com/9122899/54645546-bbdc1280-4a72-11e9-88e8-389927fb0cd1.png)

![details-longer](https://user-images.githubusercontent.com/9122899/54629967-267b5700-4a4f-11e9-9705-e8b29d157c73.png)

lizsurette:
> In the future when longer Details cards are implemented (see examples below) I'd worry that including the icon could make it harder to see other (more important) icons. I'd suggest just "Not available" text without an icon, mocked up below.
> 
[...]

PR #307, opened 2019-04-04
ClusterOverview: refactor Utilization card and connect

mareklibra:
Utilization card is refactored to split concern between
ClusterOverview and Dashboard components.

Connected to real data.

So far CPU only, other items will follow.

### DIFF
--- a/src/components/Dashboard/Utilization/UtilizationItem.js
+++ b/src/components/Dashboard/Utilization/UtilizationItem.js
@@ -1,0 +1,74 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Col, Row, SparklineChart } from 'patternfly-react';
+
+import { InlineLoading } from '../../Loading';
+import { NOT_AVAILABLE } from './strings';
+
+export const UtilizationItem = ({ id, title, data, maxY, unit, isLoading, LoadingComponent }) => {
+  const axis = {
+    y: {
+      show: true,
+      padding: {
+        top: 0,
+        bottom: 0,
+      },
+      tick: {
+        count: 3,
+      },
+    },
+    x: {
+      show: false,
+    },
+  };
+  if (maxY) {
+    axis.y.max = maxY;
+  }
+
+  let actual;
+  let chartData;
+  if (!isLoading && data) {
+    actual = Math.round(data[data.length - 1]);
+    chartData = {
+      columns: [[unit, ...data.map(val => val.toFixed(1))]],
+      unload: true,
+      type: 'area',
+    };
+  }
+
+  return (
+    <div className="kubevirt-utilization__item">
+      <Row>
+        <Col lg={2} md={2} sm={2} xs={2}>
+          {title}
+        </Col>
+        <Col className="kubevirt-utilization__item-actual" lg={3} md={3} sm={3} xs={3}>
+          {isLoading || !data || `${actual} ${unit}`}
+        </Col>
+        <Col className="kubevirt-utilization__item-chart" lg={7} md={7} sm={7} xs={7}>
+          {isLoading && <LoadingComponent />}
+          {!isLoading && data && <SparklineChart id={id} data={chartData} axis={axis} unloadBeforeLoad />}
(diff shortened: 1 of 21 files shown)


mareklibra:
Loading state:

01_loading
---
02_data

ohadlevy:
just wondering, whats the timeout on fetching the data? (in case Prometheus is down for example), do we also show error to load the card?

mareklibra:
@ohadlevy , good point. It will be implemented within #276 consistently for all cards.

src/selectors/prometheus/selectors.js
```diff
@@ -19,3 +19,8 @@ export const getCapacityStats = response => {
   const value = get(response, 'data.result[0].value[1]');
   return parseNumber(value);
 };
+
+export const getUtilizationCpuStats = response => {
+  const values = get(response, 'data.result[0].values', []);
+  return values.map(timeValuePair => parseNumber(timeValuePair[1]));
```
rawagner: I think it would be better to return null if we have not values, instead of empty array

src/components/Dashboard/Utilization/UtilizationItem.js
```diff
@@ -0,0 +1,67 @@
...
+    <div className="kubevirt-utilization__item">
+      <Row>
+        <Col lg={2} md={2} sm={2} xs={2}>
+          {title}
+        </Col>
+        <Col className="kubevirt-utilization__item-actual" lg={3} md={3} sm={3} xs={3}>
+          {isLoading || `${actual} ${unit}`}
+        </Col>
+        <Col className="kubevirt-utilization__item-chart" lg={7} md={7} sm={7} xs={7}>
+          {isLoading ? <LoadingComponent /> : <SparklineChart id={id} data={chartData} axis={axis} unloadBeforeLoad />}
```
rawagner: can you add 'Not available' state ? when isLoading is false and data are null. We already have that in `ClusterCapacity` https://github.com/kubevirt/web-ui-components/blob/master/src/components/Dashboard/Capacity/CapacityItem.js#L24

src/components/Dashboard/Utilization/UtilizationItem.js
```diff
@@ -0,0 +1,67 @@
...
+  const chartData = {
+    columns: [[unit, ...data.map(val => val.toFixed(1))]],
+    unload: true,
+    type: 'area',
+  };
+
+  return (
+    <div className="kubevirt-utilization__item">
+      <Row>
+        <Col lg={2} md={2} sm={2} xs={2}>
```
rawagner: looking at https://docs.google.com/presentation/d/1kfzD1jyEAIejWFKoKSTJLCetwvmIriLI2UlFbyB-xbg/edit#slide=id.g54d7130601_45_38 - we need to improve behaviour of UtilizationItem when there is not enough space

src/selectors/prometheus/selectors.js
mareklibra: done

src/components/Dashboard/Utilization/UtilizationItem.js
mareklibra: done

src/components/Dashboard/Utilization/UtilizationItem.js
mareklibra: Good catch. Can I deal with that in a separate PR?

There are basically two options
- introduce sort of `isNarrow` property (preferred)
- refactor the component structure and handle it via CSS (too complicated considering the case)

mareklibra:
If metric can not be accessed:

03_na

src/components/Dashboard/Utilization/UtilizationItem.js
rawagner: Sure, lets fix this in followup

rawagner:
@mareklibra just FYI I have asked Andy to provide us designs for Not Available/ Loading state https://github.com/kubevirt/web-ui-components/issues/310 so once we will have those we may need to change this a bit, however Im happy with the current state

PR #308, opened 2019-04-04
ClusterOverview: Improve prometheus query for CPU Capacity

mareklibra:
and fix Capacity tests.


rawagner:
@mareklibra is there any PR for `web-ui` which changes CPU query to get CPU usage in % ? Or maybe you want to use `cpuUtilization` instead of `cpuUsed` ?

mareklibra:
@rawagner , I forgot to push :-) https://github.com/kubevirt/web-ui/pull/263

PR #312, opened 2019-04-04
Add hosts to the inventory card

honza:


src/utils/status/host/hostStatus.js
```diff
@@ -0,0 +1,6 @@
+import { HOST_STATUS_READY } from './constants';
+
+// TODO(honza): Fill this out
+export const getHostStatus = host => ({
+  status: HOST_STATUS_READY,
```
honza: @flofuchs How could we use your work in #293 here?

suomiy: yup, it would make sense to introduce this method according to our conventions so we can  use it in inventory and other places

suomiy:
merging, there are TODOs, but they are not blocking since the inventory will just show number of hosts and ok statuses for all of them. Other statuses like Error/Warn/InProgress will be implemented in the future. Maybe in https://github.com/kubevirt/web-ui-components/pull/293?

PR #314, opened 2019-04-05
Dashboard, Utilization: Render in narrow container

mareklibra:
The Utilization component renders different sub-component structure
for a narrow container.

### DIFF
--- a/src/components/Dashboard/Utilization/UtilizationItem.js
+++ b/src/components/Dashboard/Utilization/UtilizationItem.js
@@ -1,60 +1,109 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
+import Measure from 'react-measure';
 
 import { Col, Row, SparklineChart } from 'patternfly-react';
+
+import { MEDIA_QUERY_SM } from '../../../utils';
 
 import { InlineLoading } from '../../Loading';
 import { NOT_AVAILABLE } from './strings';
 
-export const UtilizationItem = ({ id, title, data, maxY, unit, isLoading, LoadingComponent }) => {
-  const axis = {
-    y: {
-      show: true,
-      padding: {
-        top: 0,
-        bottom: 0,
-      },
-      tick: {
-        count: 3,
-      },
-    },
-    x: {
-      show: false,
+export class UtilizationItem extends React.PureComponent {
+  state = {
+    dimensions: {
+      width: -1,
     },
   };
-  if (maxY) {
-    axis.y.max = maxY;
-  }
 
-  let actual;
-  let chartData;
-  if (!isLoading && data) {
-    actual = Math.round(data[data.length - 1]);
-    chartData = {
-      columns: [[unit, ...data.map(val => val.toFixed(1))]],
-      unload: true,
-      type: 'area',
+  onResize = contentRect => {
+    this.setState({ dimensions: contentRect.bounds });
+  };
+
+  render() {
+    const { id, title, data, maxY, unit, isLoading, LoadingComponent } = this.props;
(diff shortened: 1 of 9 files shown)


mareklibra:
narrow

rawagner:
I would expect that the component behaviour would be more dynamic, controlled by actual size of the card, not by `isNarrow` prop. Maybe we could use some library like `react-bounds`. @suomiy @mareklibra do you guys have any experience with that ?

mareklibra:
@rawagner , not yet but I can give it a try. Based on documentation, it looks great.

mareklibra:
The use of `react-bounds` seems to be straightforward, just unit testing is not so simple as without it (requires the component to be mounted).

Let's see it in action for a moment, it might make sense to replace the `MediaQuery` by this library.

rawagner:
@mareklibra looks great! Can you center Loading and Not Available texts ? 
utilization_loading

src/components/Dashboard/Utilization/UtilizationItem.js
```diff
@@ -1,69 +1,107 @@
...
+      const chartData = {
+        columns: [[unit, ...data.map(val => val.toFixed(1))]],
+        unload: true,
+        type: 'area',
+      };
+      actual = `${Math.round(data[data.length - 1])} ${unit}`;
+      chart = <SparklineChart id={id} data={chartData} axis={axis} unloadBeforeLoad />;
+    }
+
+    if (isBound('narrow')) {
```
rawagner: maybe it would be easier to maintain if we keep only one return statement and parametrize Column sizes and Chart part. For the first div with different className with can use `classNames` function

mareklibra: the component structure is different, I use two `<Row>` components in the `narrow` case

suomiy:
I know `react-responsive` is not perfect but I would advise against using `react-bounds`. Mainly because this project has only one maintainer and is not actively maintained.

I think the desired functionality can be achieved with `react-responsive`.

mareklibra:
> I think the desired functionality can be achieved with react-responsive.

Good point with the maintenance ...
IIUC, `react-responsive` can not be used for checking container element boundary, just for the window size.

rawagner:
how about[ react-measure](https://www.npmjs.com/package/react-measure) ? It seems to be a lot more active than react-bounds

mareklibra:
The `react-measure` is slightly harder to work with, but its much better unit-testable.
The project is active atm and serves our needs, so this PR is reimplemented to use it.

rawagner:
LGTM. @suomiy any objections agaist react-measure ?

suomiy:
> `react-responsive` can not be used for checking container element boundary, just for the window size.

I think it could make more sense to give the graph more space and use only the narrow version only?

Also the x axis + units are missing. 

Although `react-measure` is a bit better (still 1-2 maintainers). Is this really needed? Can't we rather design alongside patternfly breakpoints?  This way the page will morph quite often.

mareklibra:
> I think it could make more sense to give the graph more space and use only the narrow version only?

I like the wider (recent) design more if space is available.

> Also the x axis + units are missing.

I will play with the X-axis independently together with the time-period dropdown selector. I can not use X-axis features of the C3chart for that to meet recent design.

The units will be implemented within this PR, thanks for spotting it.

> Can't we rather design alongside patternfly breakpoints?

This gives us wider flexibility. Using media breakpoints will not solve resizing of an inner component based on available space. Recently, the same component will be used either in wide central column or narrow right-column. Without checking boundary on the component-level, this is hard to achieve.

rawagner:
X-axis will be solved in a followup issue.

PR #316, opened 2019-04-07
Do not filter by namespace if vm is missing a namespace

yaacov:
Currently when vm is missing a namespace we do not show any nic, we want to show all namespaces when not namespace is defined.

Ref: https://github.com/kubevirt/web-ui/pull/265

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1687917

### DIFF
--- a/src/components/CreateNicRow/CreateNicRow.js
+++ b/src/components/CreateNicRow/CreateNicRow.js
@@ -53,12 +53,12 @@ const getUsedNetworks = vm => {
   return usedNetworks;
 };
 
-const getNetworkChoices = (vm, networks) => {
+const getNetworkChoices = (vm, networks, namespace) => {
   const usedNetworks = getUsedNetworks(vm);
   const networkChoices = networks
     .filter(
       network =>
-        network.metadata.namespace === vm.metadata.namespace &&
+        (!namespace || network.metadata.namespace === namespace) &&
         !usedNetworks
           .filter(usedNetwork => usedNetwork.networkType === NETWORK_TYPE_MULTUS)
           .find(usedNetwork => usedNetwork.name === getName(network))
@@ -78,9 +78,10 @@ const getNetworkChoices = (vm, networks) => {
 
 const getNicColumns = (nic, networks, LoadingComponent) => {
   const bindingChoices = getNetworkBindings(get(settingsValue(nic, 'network'), 'networkType'));
+  const namespace = (nic.vm && nic.vm.metadata.namespace) || (nic.vmTemplate && nic.vmTemplate.metadata.namespace);
   let network;
   if (networks) {
-    const networkChoices = getNetworkChoices(nic.vm, networks);
+    const networkChoices = getNetworkChoices(nic.vm, networks, namespace);
     network = {
       id: 'network-type',
       type: DROPDOWN,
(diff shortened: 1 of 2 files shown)


yaacov:
@rawagner Thanks for finding the issue that caused the bug 100 ,

@rawagner @suomiy please review.

I think this should be wip until https://github.com/kubevirt/web-ui/pull/265 is mered, does that make sense ?

rawagner:
IMHO its not necessary to do the filtering here again, when we agree that we are expecting networks which are already filtered

yaacov:
> IMHO its not necessary to do the filtering here again, when we agree that we are expecting networks which are already filtered

@rawagner  :+1: I also think so, will remove the check, and leave as WIP until 265 is merged.

p.s.
I didn't know what is going to happen with https://github.com/kubevirt/web-ui/pull/265 so didn't want to remove ...

src/components/CreateNicRow/CreateNicRow.js
```diff
@@ -53,12 +53,12 @@ const getUsedNetworks = vm => {
...
 };
 
-const getNetworkChoices = (vm, networks) => {
+const getNetworkChoices = (vm, networks, namespace) => {
   const usedNetworks = getUsedNetworks(vm);
   const networkChoices = networks
     .filter(
       network =>
-        network.metadata.namespace === vm.metadata.namespace &&
+        (!namespace || network.metadata.namespace === namespace) &&
```
suomiy: please use `getNamespace` method here as well

src/components/CreateNicRow/CreateNicRow.js
```diff
@@ -78,9 +78,10 @@ const getNetworkChoices = (vm, networks) => {
 
 const getNicColumns = (nic, networks, LoadingComponent) => {
   const bindingChoices = getNetworkBindings(get(settingsValue(nic, 'network'), 'networkType'));
+  const namespace = (nic.vm && nic.vm.metadata.namespace) || (nic.vmTemplate && nic.vmTemplate.metadata.namespace);
```
suomiy: \+ here

suomiy:
> Currently when vm is missing a namespace we do not show any nic, we want to show all namespaces when not namespace is defined.

what do you mean? how could a vm possibly miss a namespace?

src/components/CreateNicRow/CreateNicRow.js
yaacov: @suomiy , per @rawagner comment https://github.com/kubevirt/web-ui-components/pull/316#issuecomment-480741538 I'm going to remove this code, it will move to web-ui repo via https://github.com/kubevirt/web-ui/pull/265

yaacov:
> what do you mean? how could a vm possibly miss a namespace?

@suomiy we do not have a vm at that stage. The bug happen when we edit `vmTemplate` , when `createNicRow` is called from edit `vmTemplate` `vm.namespace` is not defined.

yaacov:
@rawagner @suomiy I removed the filtering by namespace per comment https://github.com/kubevirt/web-ui-components/pull/316#issuecomment-480741538 this should also adress comment https://github.com/kubevirt/web-ui-components/pull/316#pullrequestreview-223754545 because filtering is moved upstream.

yaacov:
Removed wip because https://github.com/kubevirt/web-ui/pull/265 is merged.

yaacov:
@suomiy any thoughts about coveralls error, I do not know what I did to upset it :-( ?

rawagner:
@yaacov coverage decreased so coveralls wont approve this PR - but I will!

PR #318, opened 2019-04-08
Adding Capacity Card to Storage Overview

AyushAmbastha:
Added Capacity card to show 'Total capacity' in the form of a donut chart.

Screenshot from 2019-04-08 16-33-15

Tasks yet to be done as suggested by @julienlim@redhat.com -
1. Change the font of 'Total Capacity' title to a smaller one. The text should be smaller than the card title 'Capacity'.
2. Change GB to GiB.
How do I go about it?

These changes will be sent in a following patch.
Also, I'll send a PR in kubevirt/web-ui for the same and reference this there.

### DIFF
--- a/src/components/StorageOverview/Capacity/Capacity.js
+++ b/src/components/StorageOverview/Capacity/Capacity.js
@@ -1,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { InlineLoading } from '../../Loading';
+
+import {
+  DashboardCard,
+  DashboardCardBody,
+  DashboardCardHeader,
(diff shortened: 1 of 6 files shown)


rawagner:
@AyushAmbastha you can change units to *iB format here https://github.com/kubevirt/web-ui-components/blob/master/src/utils/utils.js#L85 but it will affect all units we have including network traffic so for example GBps will became GiBps. @andybraren I think that is OK, but to make sure..

AyushAmbastha:
Update: Integration in kubevirt/web-ui also done https://github.com/kubevirt/web-ui/pull/273

PR #321, opened 2019-04-08
Improve dashboard performance

rawagner:
- Convert cards which are using data from Prometheus to PureComponent

### DIFF
--- a/src/components/ClusterOverview/Capacity/Capacity.js
+++ b/src/components/ClusterOverview/Capacity/Capacity.js
@@ -15,62 +15,67 @@ import { formatBytes, formatPercents, formatNetTraffic } from '../../../utils';
 import { getCapacityStats, getLastUtilizationStat } from '../../../selectors';
 import { CapacityBody } from '../../Dashboard/Capacity/CapacityBody';
 
-export const Capacity = ({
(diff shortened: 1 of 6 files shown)


rawagner:
@mareklibra as @suomiy already knows, we have a performance issue with dashboard - all cards are re-rendered too often - it is caused by Firehose (props are changing everytime anything changes in watched resources). This PR limits re-rendering of components which are using only data from prometheus. We will have to explore how to fix Firehose issue later too, but its a bit more complicated

mareklibra:
Possible root-cause: https://github.com/kubevirt/web-ui/pull/244

The value was chosen without proper evaluation, just based on assumption that the bottle-neck is prometheus. Intensive refresh-rate was required as a short-term solution for one specific use-case.

I see multiple options how to move forward, let's discuss them offline tomorrow.
The workaround suggested in this PR is not a way I would go. I am even wondering, that linter is fine with that :-)

rawagner:
@mareklibra prometheus refresh rate is actually fine, the problem is with Firehose. Lets discuss it tomorrow :)

rawagner:
replaced by https://github.com/kubevirt/web-ui-components/pull/321

PR #322, opened 2019-04-08
Utilization: Add Disk Usage chart

mareklibra:
and refactor prometheus queries for disk capacity so they
can be reused for utilization.

### DIFF
--- a/src/components/ClusterOverview/Capacity/Capacity.js
+++ b/src/components/ClusterOverview/Capacity/Capacity.js
@@ -52,7 +52,7 @@ export const Capacity = ({
         <CapacityItem
           id="storage"
           title="Storage"
-          used={getCapacityStats(storageUsed)}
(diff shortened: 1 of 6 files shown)


mareklibra:
In the memory/storage utilization charts, the `total` is used as maximum for Y-axis.

If `total` is much higher than actual consumption, the chart does not show "progress" vividly as the consumption is nearly to be a constant.

What about using
- maximal historic value (from the given range) as Y-axis maximum
- but units (KiB, TiB, etc) from the `total`?

What others think?

rawagner:
I would vote for using value AND units from max historic value

mareklibra:
And yet another (better) possibility handling the case with "almost constant" values within the range:
For max-y, use max historic value multiplied by 2 or cluster's capacity total, whatever is lower.

This approach handles:
- near-constant values
- values `<<` total

suomiy:
+1

PR #325, opened 2019-04-08
[Cluster overview] Add disks to inventory

honza:
Ref kubevirt/web-ui#294

### DIFF
--- a/src/components/ClusterOverview/Inventory/Inventory.js
+++ b/src/components/ClusterOverview/Inventory/Inventory.js
@@ -15,17 +15,19 @@ import {
   mapPvcsToProps,
   mapVmsToProps,
   mapHostsToProps,
+  mapDiskStatsToProps,
 } from '../../Dashboard/Inventory/utils';
 import { InventoryRow } from '../../Dashboard/Inventory/InventoryRow';
 import { InlineLoading } from '../../Loading';
 
-const InventoryBody = ({ nodes, pods, vms, vmis, pvcs, migrations, hosts, LoadingComponent }) => (
+const InventoryBody = ({ nodes, pods, vms, vmis, pvcs, migrations, hosts, diskStats, LoadingComponent }) => (
   <React.Fragment>
     <InventoryRow title="Nodes" {...mapNodesToProps(nodes)} LoadingComponent={LoadingComponent} />
     <InventoryRow title="Hosts" {...mapHostsToProps(hosts)} LoadingComponent={LoadingComponent} />
     <InventoryRow title="PVCs" {...mapPvcsToProps(pvcs)} LoadingComponent={LoadingComponent} />
     <InventoryRow title="Pods" {...mapPodsToProps(pods)} LoadingComponent={LoadingComponent} />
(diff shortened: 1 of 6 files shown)


rawagner:
@honza do you plan to add any checking for disk status or is it going to be hardcoded to OK ?

honza:
The baremetal operator doesn't expose any kind of disk status information. It only shows the description of the hardware. Would it be appropriate to query Prometheus in this instance?

src/components/ClusterOverview/Inventory/Inventory.js
```diff
@@ -15,16 +15,18 @@ import {
   mapPvcsToProps,
   mapVmsToProps,
   mapHostsToProps,
+  mapDiskStatsToProps,
 } from '../../Dashboard/Inventory/utils';
 import { InventoryRow } from '../../Dashboard/Inventory/InventoryRow';
 
-const InventoryBody = ({ nodes, pods, vms, vmis, pvcs, migrations, hosts }) => (
+const InventoryBody = ({ nodes, pods, vms, vmis, pvcs, migrations, hosts, diskStats }) => (
```
rawagner: where are these `diskStats` coming from ? Is there any PR for `web-ui` ? I also wonder how is it that the build is passing but the prop is missing in `propTypes` :(

honza: It's coming from prometheus, the web-ui PR is coming soon. It's very similar to the code in the storage dashboard.

honza: PR is up https://github.com/kubevirt/web-ui/pull/294

src/constants/storage.js
```diff
@@ -0,0 +1,14 @@
+export const STORAGE_PROMETHEUS_QUERIES = {
+  CEPH_STATUS_QUERY: 'ceph_health_status',
+
+  CEPH_OSD_UP_QUERY: 'sum(ceph_osd_up)',
+  CEPH_OSD_DOWN_QUERY: 'count(ceph_osd_up == 0.0) OR vector(0)',
+
+  CAPACITY_STORAGE_TOTAL_BASE_CEPH_METRIC: 'ceph_cluster_total_bytes', // available with Ceph only
```
suomiy: I think it makes more sense to make these queries more generic. I.e. without CAPACITY/UTILIZATION prefixes.

src/constants/storage.js
```diff
@@ -0,0 +1,14 @@
...
+  CAPACITY_STORAGE_TOTAL_QUERY: 'ceph_cluster_total_bytes',
+  CAPACITY_STORAGE_TOTAL_DEFAULT_QUERY: 'sum(node_filesystem_avail_bytes)',
+
+  UTILIZATION_STORAGE_USED_QUERY: 'ceph_cluster_total_used_bytes',
```
suomiy: maybe better to rename this to something  like `CEPH_CLUSTER_TOTAL_USED` ? etc.

src/constants/storage.js
honza: Since the names are very convoluted, I'd suggest merging the two PRs that extract those values into constants before renaming them. This should simplify code review, and prevent bugs.

suomiy: ok then

PR #328, opened 2019-04-09
Change Font size and Units in Capacity card

AyushAmbastha:
Changes done-
1. Font size of 'Total capacity' reduced as it shouldn't be larger than the card title. This also maintains consistency across cards in terms of aesthetics.
2. Units changed from GB to GiB and so on.

Before-
Screenshot from 2019-04-09 16-17-01

After-
Screenshot from 2019-04-09 16-36-47

### DIFF
--- a/src/components/ClusterOverview/Capacity/tests/__snapshots__/Capacity.test.js.snap
+++ b/src/components/ClusterOverview/Capacity/tests/__snapshots__/Capacity.test.js.snap
@@ -22,11 +22,11 @@ exports[`<Capacity /> renders correctly 1`] = `
       <div
         class="kubevirt-capacity__item"
       >
-        <h2
+        <h3
           class="kubevirt-capacity__item-title"
         >
           CPU
-        </h2>
+        </h3>
         <h6>
           50 % available out of 100 %
         </h6>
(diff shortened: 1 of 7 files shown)


rawagner:
please change https://github.com/kubevirt/web-ui-components/blob/0fba1d4fc046de0b150524386eed4f55d1f2b9da/src/utils/utils.js#L95 to PiB too. Or maybe better would be `Object.keys(BYTE_UNITS)[Object.keys(BYTE_UNITS).length-1]` to avoid this in the future

AyushAmbastha:
> please change
> 
> [web-ui-components/src/utils/utils.js](https://github.com/kubevirt/web-ui-components/blob/0fba1d4fc046de0b150524386eed4f55d1f2b9da/src/utils/utils.js#L95)
> 
> Line 95 in [0fba1d4](/kubevirt/web-ui-components/commit/0fba1d4fc046de0b150524386eed4f55d1f2b9da)
> 
>  unit = unit || Object.keys(BYTE_UNITS).find(key => bytes < 1024 ** (BYTE_UNITS[key] + 1)) || 'PB'; 
> to PiB too. Or maybe better would be `Object.keys(BYTE_UNITS)[Object.keys(BYTE_UNITS).length-1]` to avoid this in the future

Should I change 1024 to 1000 everywhere as well? As Gibibytes is by 1000 and Gigabytes is by 1024, for an example.

rawagner:
hm..i think 1 KiB = 1024 Bytes and 1 KB = 1000 Bytes so if we keep the calculations and change units to *iB we should be fine.

AyushAmbastha:
@rawagner Sorry for the confusion regarding the units, I've made the requested changes. Please review.

PR #333, opened 2019-04-09
Add selector to check if host can add machine

jtomasek:


src/components/BareMetalHosts/BaremetalHostStatus.js
```diff
@@ -0,0 +1,32 @@
...
+  ready: GenericSuccess,
+  provisioned: GenericSuccess,
+  provisioning: GenericProgress,
+  'preparing to provision': GenericProgress,
+  deprovisioning: GenericProgress,
+  'making host available': GenericProgress,
+};
+
+export const BaremetalHostStatus = ({ host }) => {
+  const hostStatus = getHostStatus(host);
```
suomiy: we found out that for better reusability it is better to decouple status functionality from the UI (e.g. vmStatus is used at many places). Please see  `utils/status`. Could you use this approach here?

 Basically it is a function which returns

```js
{
   status, // constant
   ...additionalUsefulData // e.g. message
}
```

src/components/BareMetalHosts/StatusComponents.js
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Popover } from '@patternfly/react-core';
+
+import { getHostStatus } from '../../selectors/host/selectors';
+
+// Status titles. Could be i18n'ed at some point.
+const statusTextMap = {
```
suomiy: \+ the string mapping could be done there

src/components/BareMetalHosts/StatusComponents.js
```diff
@@ -0,0 +1,89 @@
...
+  ) : (
+    <IconAndText icon="pficon-error-circle-o" text={getStatusText(getHostStatus(host))} />
+  );
+
+GenericError.propTypes = {
+  host: PropTypes.object.isRequired,
+};
+
+// Generic status component as a fallback
+export const GenericStatus = ({ host }) => <React.Fragment>{getStatusText(getHostStatus(host))}</React.Fragment>;
```
suomiy: why the `React.Fragment` ?

suomiy:
sorry, I see it depends on the other PR now..

PR #335, opened 2019-04-10
[Cluster dashboard] Indicate when hosts are in progress

honza:
Depends on #293 [MERGED]


honza:
Needs a rebase before any reviews

honza:
Ready for reviews

src/utils/status/host/hostStatus.js
```diff
@@ -1,6 +1,5 @@
-import { HOST_STATUS_READY } from './constants';
+import { getHostStatus as getHostStatusSelector } from '../../../selectors';
 
-// TODO(honza): Fill this out
 export const getHostStatus = host => ({
-  status: HOST_STATUS_READY,
+  status: getHostStatusSelector(host),
```
suomiy: this is going the other way around. It should be implemented here.

honza: I'm not sure I follow. Could you explain a bit more? 

There's really nothing to implement. It just passes the value into this status object.

suomiy: sorry, I wrote it confusingly. I just meant that https://github.com/kubevirt/web-ui-components/pull/293 needs to be resolved first and it will call the status selector.

honza:
Dependent PR merged

PR #337, opened 2019-04-10
finalize vmware v2v

suomiy:
- create conversion pod and other objects

There is still a lot of work to be done here and I will tackle tackle deeper issues in followups. This PR is already big enough.

sibling: https://github.com/kubevirt/web-ui/pull/288


suomiy:
`finalize` means it works. Not that it is finished :)

src/components/Wizard/CreateVmWizard/providers/index.js
```diff
@@ -21,6 +27,14 @@ const getProviderHelp = basicSettings => {
 };
 
 export const importProviders = (basicSettings, operatingSystems, WithResources, k8sCreate, k8sGet, k8sPatch) => ({
+  [PROVIDER_VMWARE_NAMESPACE_ALERT_KEY]: {
+    id: 'vcenter-namespace-alert',
+    type: CUSTOM,
+    title: ' ',
+    CustomComponent: () => <Alert type="warning">Please select a namespace.</Alert>,
+    isVisible: basicVmSettings =>
+      isImageSourceType(basicVmSettings, PROVISION_SOURCE_IMPORT) && !settingsValue(basicSettings, NAMESPACE_KEY),
```
mareklibra: There was plan to make this requirement more generic, ie. by disabling subsequent dropdowns if `namespace` is not selected. @rawagner, any news?

src/components/Wizard/CreateVmWizard/providers/index.js
```diff
@@ -29,12 +43,17 @@ export const importProviders = (basicSettings, operatingSystems, WithResources,
     choices: providerList,
     required: true,
     onChange: (...opts) => onProviderChanged(k8sCreate, k8sGet, ...opts),
-    isVisible: basicVmSettings => isImageSourceType(basicVmSettings, PROVISION_SOURCE_IMPORT),
+    isVisible: basicVmSettings =>
+      settingsValue(basicSettings, NAMESPACE_KEY) && isImageSourceType(basicVmSettings, PROVISION_SOURCE_IMPORT),
```
mareklibra: Likewise - can become unmaintainable if not resolved at single place

src/components/Wizard/CreateVmWizard/providers/v2vvmwareController.js
```diff
@@ -11,161 +14,33 @@ export const startV2VVMWareController = async ({ k8sCreate, k8sGet, namespace })
...
-    createOperator(params).then(undefined, rejectLogger('V2V VMWare controller'));
-    /* eslint-enable promise/catch-or-return */
+    createServiceAccount(params).catch(rejectLogger('Service Account'));
+    createRole(params).catch(rejectLogger('Role'));
+    createRoleBinding(params).catch(rejectLogger('Role Binding'));
+    createOperator(params).catch(rejectLogger('V2V VMWare controller'));
   }
 };
 
-const createServiceAccount = ({ k8sCreate, namespace }) =>
```
mareklibra: I am not big fan of melting this code into the rest of our code-base as it sends incorrect message for other contributors.

My original intention was to keep it isolated as it should not be part of UI at all.
We just don't have better place to put it now.
As soon as it is found, we should remove it.

src/k8s/objects/v2v/vmware/conversionPod.js
```diff
@@ -0,0 +1,50 @@
...
+  metadata: {
+    generateName: CONVERSION_GENERATE_NAME,
+    namespace,
+  },
+  spec: {
+    serviceAccountName,
+    containers: [
+      {
+        name: CONVERSION_BASE_NAME,
+        image: 'quay.io/nyoxi/kubevirt-v2v-conversion:1.12.1-1-gf665c0a',
```
mareklibra: Please add TODO here for better visibility.
It's a huge gap which need to be resolved together with the v2v-vmware repo

src/k8s/requests/v2v/constants.js
```diff
@@ -0,0 +1,9 @@
+export const CONVERSION_POD_TEMP_MOUNT_PATH = '/var/tmp';
+export const CONVERSION_POD_VDDK_MOUNT_PATH = '/opt/vmware-vix-disklib-distrib';
+
+export const CONVERSION_BASE_NAME = 'kubevirt-v2v-conversion';
+export const CONVERSION_GENERATE_NAME = `${CONVERSION_BASE_NAME}-`;
+
+export const CONVERSION_PROGRESS_ANNOTATION = 'v2vConversionProgress';
```
mareklibra: We should not use `Annotations` for status info.

The `status` section of an object is better place.

src/k8s/requests/v2v/importVmware.js
```diff
@@ -0,0 +1,263 @@
...
+};
+
+export const importVmwareVm = async (getSetting, networks, storages, { k8sCreate, k8sPatch }) =>
+  [
+    createProviderSecret,
+    createConversionPodSecret,
+    resolveRolesAndServiceAccount,
+    resolveStorages,
+    startConversionPod,
+  ].reduce(
```
mareklibra: Vivat simplicity! Wouldn't it be more readable for future generations to avoid such constructs?

src/components/Wizard/CreateVmWizard/providers/index.js
rawagner: no news, havent had time to look at it. Here is a design https://docs.google.com/document/d/1ks4H-ImD-r0s5eV4FsP_73kRdAP6sKJTRbj8nClpyp0/edit#heading=h.o4npgyz4fz5s which shows that by disabling dropdowns/other fields we would 'guide' user to correct flow.

src/components/Wizard/CreateVmWizard/providers/index.js
suomiy: yes, I am aware of that. I will do a large refactoring of BasicSettingsTab and vmWare provider in the followup. It is becoming quite unmaintainable

src/components/Wizard/CreateVmWizard/providers/v2vvmwareController.js
suomiy: Yup, it is not good but we need a place for that and have some separation of code for now. We already have v2v in requests.

src/k8s/requests/v2v/constants.js
suomiy: I just found out that it is not possible to write to status from mere pod. This needs to be done from a controller and currently we use only a pod.

src/k8s/requests/v2v/importVmware.js
suomiy: I think it is quite readable. And it gives a clear contract for each step what data to expect. I have one refactoring lined up, which will simplify this use case a bit (error rollback).

src/k8s/objects/v2v/vmware/conversionPod.js
suomiy: done

suomiy:
rebased

src/k8s/requests/v2v/constants.js
mareklibra: The "conversion pod" has access to the k8s api, it's just about setting privileges.
Annotations are definitely not good place to keep status info.
Whatever sets this information within Annotation now, can set it in the Status as well.

src/components/Wizard/CreateVmWizard/providers/v2vvmwareController.js
mareklibra: I am afraid that we will have hard time to get rid of it in the future ...
I still prefer collocation this "unrelated" stuff somewhere.

suomiy: It is already getting used in two places, so I would rather have a proper place for that in the project structure (reusability). I know it is not good to overuse this stuff in the UI, but we can enforce that in merging process.

src/k8s/requests/v2v/constants.js
suomiy: I don't know the specifics, but there is nothing to do here until this is implemented by the conversion pod.

nyoxi: After investigation using status is not possible. It is not possible to introduce new fields into the status structure -- we can only change existing fields. This is risky (data can be overwritten by other components) plus all info will be lost when the pod terminates.

Using labels is not good either as it has strict limitation on what the values can be and is not designed for storing general data.

Seems to me that using annotations is the only option.

PR #338, opened 2019-04-10
Align byte units to tectonic ones.

rawagner:
@AyushAmbastha changing storage units to tectonic ones as suggested by @andybraren in https://github.com/kubevirt/web-ui-components/pull/318#issuecomment-481282991 and network to MBps format

### DIFF
--- a/src/utils/tests/utils.test.js
+++ b/src/utils/tests/utils.test.js
@@ -3,21 +3,21 @@ import { formatBytes, formatCores, formatNetTraffic } from '../utils';
 describe('unit format functions', () => {
   it('formats bytes', () => {
     expect(formatBytes(2)).toEqual({ value: 2, unit: 'B' });
-    expect(formatBytes(3 * 1024)).toEqual({ value: 3, unit: 'KiB' });
-    expect(formatBytes(3 * 1024 ** 2)).toEqual({ value: 3, unit: 'MiB' });
-    expect(formatBytes(3 * 1024 ** 3)).toEqual({ value: 3, unit: 'GiB' });
-    expect(formatBytes(3 * 1024 ** 4)).toEqual({ value: 3, unit: 'TiB' });
-    expect(formatBytes(3 * 1024 ** 5)).toEqual({ value: 3, unit: 'PiB' });
-    expect(formatBytes(3 * 1024 ** 6)).toEqual({ value: 3 * 1024, unit: 'PiB' });
-    expect(formatBytes(3 * 1024 ** 7)).toEqual({ value: 3 * 1024 ** 2, unit: 'PiB' });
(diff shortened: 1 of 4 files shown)


AyushAmbastha:
Hey @rawagner! Thanks and acknowledged :)

src/utils/tests/utils.test.js
```diff
@@ -3,21 +3,21 @@ import { formatBytes, formatCores, formatNetTraffic } from '../utils';
...
   it('formats bytes in preferred units', () => {
-    expect(formatBytes(3 * 1024 * 1024, 'KiB')).toEqual({ value: 3 * 1024, unit: 'KiB' });
+    expect(formatBytes(3 * 1024 * 1024, 'Ki')).toEqual({ value: 3 * 1024, unit: 'Ki' });
   });
   it('formats cores', () => {
     expect(formatCores(2)).toEqual({ value: 2, unit: 'cores' });
   });
   it('formats net traffic', () => {
-    expect(formatNetTraffic(3 * 1024 * 1024)).toEqual({ value: 3, unit: 'MiBps' });
+    expect(formatNetTraffic(3 * 1000 * 1000)).toEqual({ value: 3, unit: 'MBps' });
```
suomiy: why do we show the network in `MBps` instead of `MiBps`?

andybraren: @suomiy The last time I fell into this rabbit hole (a few months ago) I learned that networking metrics are still generally measured in Mbps (Megabits per second, note the lowercase b) largely because it has long-been the convention that "everyone" uses, even in cloud-oriented consoles where they use *iB units for everything else.

It's not the greatest reason, but it is what users currently expect for networking as far as I know. I've personally can't remember seeing `Mibps` used elsewhere, but I'd be happy to be wrong.

suomiy: > The last time I fell into this rabbit hole (a few months ago) I learned that networking metrics are still generally measured in Mbps (Megabits per second, note the lowercase b) largely because it has long-been the convention that "everyone" uses, even in cloud-oriented consoles where they use *iB units for everything else.

Ok then. If it was decided like that.

> I personally can't remember ever seeing Mibps used elsewhere, but I'd be happy to be wrong.

A lot of software uses IEC binary units (KiB/s, MiB/s), e.g. curl, torrent clients.

andybraren: Good to know! As a relative outsider I like KiB/s and MiB/s and would be happy to switch to that if users find it natural. Looking at what OKD and the Prometheus Dashboard [currently do](https://github.com/kubevirt/web-ui-components/pull/318), it seems like they're using Kbps, right? They're a little ambiguous.

suomiy: to me it seems that Prometheus dashboard uses kB/s (kBps)

PR #339, opened 2019-04-10
Add BaremetalHostRole component

jtomasek:


src/selectors/host/selectors.js
```diff
@@ -47,3 +47,6 @@ export const getHostStatusText = host => {
   const hostStatus = getHostStatus(host);
   return statusTextMap[hostStatus] ? statusTextMap[hostStatus] : hostStatus;
 };
+
+export const getHostRole = hostMachine =>
+  get(hostMachine, ['metadata', 'labels', 'machine.openshift.io/cluster-api-machine-role'], '-');
```
rawagner: returning `-` is this is very UI specific. Can you just return null and in `BaremetalHostRole` do 
```
getHostRole(machine) || DASHES
```

jtomasek: Done

PR #340, opened 2019-04-10
Do not map vm status off to status ok

yaacov:
Currently we map `[VM_STATUS_OFF, VM_STATUS_UNKNOWN]` to `STATUS_RESULT_OK`.

This PR adds mapping for missing statuses to off and warning icon states.

Ref: https://jira.coreos.com/browse/CNV-1466

Screenshot:
Screenshot_2019-04-11 ClusterOverview ClusterOverview – React Cosmos

Screenshot_2019-04-11 Dashboard Inventory InventoryItemStatus InventoryItemStatus – React Cosmos


yaacov:
@suomiy @rawagner hi,
we need to map machine down to something other than ok.

We currently have this statuses: 
https://github.com/kubevirt/web-ui-components/blob/master/src/components/Dashboard/Inventory/InventoryItemStatus.js#L47
https://github.com/kubevirt/web-ui-components/blob/master/src/components/Dashboard/Inventory/utils.js#L9

Do you think warning is the right option ? should it be error ? should we create a 5'th status icon for off ? do we have state off somewhere else, do we have a mockup for off state ?

suomiy:
> Do you think warning is the right option ? should it be error ? should we create a 5'th status icon for off ? do we have state off somewhere else, do we have a mockup for off state ?

`VM_STATUS_OFF` as a warning doesn't make sense to me. I vote for 5th down state. (Better to ask designers though)


`VM_STATUS_OTHER` will never be resolved because it is used only as a helper status in `getSimpleVmStatus` so please remove it. Could you please also rename it to `VM_SIMPLE_STATUS_OTHER` ? so it is not that confusing

`VM_STATUS_UNKNOWN` makes sense as a warning

suomiy:
can you do a bit more renaming? 

`VM_STATUS_ALL` -> `VM_SIMPLE_STATUS_ALL`
`VM_STATUS_TO_TEXT` -> `VM_SIMPLE_STATUS_TO_TEXT`

this needs to be also changed in web-ui

yaacov:
@matthewcarleton hi,

We have vm state off, but we do not have an inventory item state off, shoudl we add it ?

Attached is a demo of inventory state off for VMs:
Screenshot_2019-04-10 ClusterOverview ClusterOverview – React Cosmos(1)

Does this make sense to you ?

lizsurette:
> We have vm state off, but we do not have an inventory item state off, shoudl we add it ?

FYI @andybraren

yaacov:
> VM_STATUS_ALL -> VM_SIMPLE_STATUS_ALL
> VM_STATUS_TO_TEXT -> VM_SIMPLE_STATUS_TO_TEXT

> this needs to be also changed in web-ui

@suomiy I will do this a 2 new PR's that we can merge the same time in web-ui and web-ui-components, does that make sense to you ?

src/components/Dashboard/Inventory/InventoryItemStatus.js
```diff
@@ -42,14 +42,24 @@ const InProgressStatus = ({ count }) => (
...
 const Status = ({ Component, count, ...props }) => count > 0 && <Component count={count} {...props} />;
 
-export const InventoryItemStatus = ({ ok, warn, error, inProgress }) => (
+export const InventoryItemStatus = ({ ok, warn, error, inProgress, off }) => (
   <div className="kubevirt-inventory__row-status">
     <Status Component={InProgressStatus} count={inProgress} />
     <Status Component={ErrorStatus} count={error} />
     <Status Component={WarnStatus} count={warn} />
     <Status Component={OkStatus} count={ok} />
+    <Status Component={OffStatus} count={off} />
```
suomiy: ok should be probably the last one.

suomiy:
> @suomiy I will do this a 2 new PR's that we can merge the same time in web-ui and web-ui-components, does that make sense to you ?

It is up to you. I would put it here as it is a small change and a bit relevant to this PR.

src/components/Dashboard/Inventory/InventoryItemStatus.js
yaacov: Make sense :tulip:

suomiy:
\+ also  `VM_STATUS_OTHER` -> `VM_SIMPLE_STATUS_OTHER`

andybraren:
@lizsurette @yaacov @suomiy

Yes, we should have an `off` state for VMs within the Inventory card:
`VM_STATUS_OFF` -> Off icon (as your screenshot shows)

We'd like for the Inventory card's statuses to be as close as possible to what a user would see if they looked at the VM List page. For that reason, I would prefer to use the same status icon in the Inventory card that `VM_STATUS_UNKNOWN` shows in the VM List page. If we currently use `pficon-warning-triangle-o` then that's what we should use in the Inventory card. If we use `pficon-unknown` then we should use that instead.

yaacov:
I added the new inventory "off" state looking like the vm "off", only bigger because all the other icons in inventory are big.

@suomiy @rawagner 
I tried to use "x2" in the "off" icon but it did not make the icon bigger ( used special class css instead ) am I missing something with the "size" on pficon ?

yaacov:
@andybraren hi,
do we need some pop up text to appear when we hover over the status icons ?

for example the word "Off" when hovering over the "off" icon ? @suomiy @rawagner ?

suomiy:
> We'd like for the Inventory card's statuses to be as close as possible to what a user would see if they looked at the VM List page. For that reason, I would prefer to use the same status icon in the Inventory card that VM_STATUS_UNKNOWN shows in the VM List page. If we currently use pficon-warning-triangle-o then that's what we should use in the Inventory card. If we use pficon-unknown then we should use that instead.

There is not so much space to show an icon for every state we have, like `VM_STATUS_UNKNOWN`. I think it makes more sense to show a specific set of generic set of icons (`VM_STATUS_OFF` should be probably one of them). Then the user can go to the VMs page to inspect what the warnings, etc. mean.

suomiy:
> for example the word "Off" when hovering over the "off" icon ? @suomiy @rawagner ?

My vote is that it is not necessary.

yaacov:
@suomiy is there a "correct" way to make the off icon size bigger ?

yaacov:
@suomiy @rawagner I think it's ready, please review.

suomiy:
> @suomiy is there a "correct" way to make the off icon size bigger ?

patternfly icon is just a switch between PatternflyIcon and FontAwesomeIcon components.

Only the FontAwesomeIcon component supports size property, so your only option after that is the css.

Can you please also remove the size property on inProgress icon and apply the css?

sass/components/Dashboard/inventory.scss
```diff
@@ -43,3 +43,7 @@
 .kubevirt-inventory__row-status-item-icon--warn {
   color: goldenrod;
 }
+
+.kubevirt-inventory__row-status-item-icon--off {
+  font-size: 1.5em;
```
suomiy: I tried it in cosmos and maybe it could be a tad bigger

src/components/Dashboard/Inventory/InventoryItemStatus.js
```diff
@@ -42,13 +42,23 @@ const InProgressStatus = ({ count }) => (
```
suomiy: InProgressStatus is also ` type="pf"` component. Please try the sizes.

yaacov:
> Can you please also remove the size property on inProgress icon and apply the css?

Sure :plate_with_cutlery:

src/components/Dashboard/Inventory/InventoryItemStatus.js
```diff
@@ -42,13 +42,23 @@ const InProgressStatus = ({ count }) => (
...
+    <span className="kubevirt-inventory__row-status-item-text">{count}</span>
+  </div>
+);
+
+OffStatus.propTypes = StatusPropTypes;
+
 const Status = ({ Component, count, ...props }) => count > 0 && <Component count={count} {...props} />;
 
-export const InventoryItemStatus = ({ ok, warn, error, inProgress }) => (
+export const InventoryItemStatus = ({ ok, warn, error, inProgress, off }) => (
```
suomiy: please showcase this in the fixture

suomiy:
build failing..

yaacov:
@suomiy I do not understand why I get `Ki` instead of `KiB` looking ...

yaacov:
@suomiy not sure we can merge this ... something is wrong with the `KiB` units, looking what happen :-(

suomiy:
> @suomiy I do not understand why I get Ki instead of KiB looking ...

it was merged recently. Please rebase on the recent master

yaacov:
@suomiy I'm rebased on master and get `Ki` is it the way it suppose to be ?

yaacov:
ahh... https://github.com/kubevirt/web-ui-components/pull/338 

@suomiy it's ok, somehow this snap changes were not part of #338 but we do need them.

suomiy:
Yes you have it correct. The build is failing on master.

suomiy:
we merged wrong snapshots before ^^

andybraren:
@suomiy

> > We'd like for the Inventory card's statuses to be as close as possible to what a user would see if they looked at the VM List page. For that reason, I would prefer to use the same status icon in the Inventory card that VM_STATUS_UNKNOWN shows in the VM List page. If we currently use pficon-warning-triangle-o then that's what we should use in the Inventory card. If we use pficon-unknown then we should use that instead.
> 
> There is not so much space to show an icon for every state we have, like `VM_STATUS_UNKNOWN`. I think it makes more sense to show a specific set of generic set of icons (`VM_STATUS_OFF` should be probably one of them). Then the user can go to the VMs page to inspect what the warnings, etc. mean.

Acknowledged - this is something we're working on, and likely involves a deeper dive on our end into all of the possible statuses, how to prioritize them, when to show which, etc. In the short term we will very likely propose removing the green "okay" status from the Inventory card entirely (for this and other reasons). We'll share that design proposal soon.

andybraren:
> Are you okay with showing no status if okay part is removed? Because I am not sure we have much else to show for example in PVCs. Only pending icon when the pvc is pending.

Yes, that's totally fine. 👍

PR #342, opened 2019-04-10
Adding Health in Dashboard

AyushAmbastha:
Added the health folder in dashboard so that it's generic and can be used across components. Working remains the same.


src/components/Dashboard/Health/strings.js
```diff
@@ -0,0 +1,4 @@
+export const HEALTHY= 'OCS is Healthy';
```
rawagner: these values are specific for Storage dashboard

src/components/Dashboard/Health/HealthItem.js
```diff
@@ -3,31 +3,33 @@ import PropTypes from 'prop-types';
 import { Icon } from 'patternfly-react';
 import { get } from 'lodash';
 
+import { HEALTHY, DEGRADED, ERROR, UNKNOWN } from './strings';
+
 const HealthStatus = {
```
rawagner: this is specific for Storage Health and should not be in this generic Health

src/components/Dashboard/Health/HealthItem.js
```diff
@@ -3,31 +3,33 @@ import PropTypes from 'prop-types';
...
-    message: 'Cannot get OCS health',
+    message: UNKNOWN,
     iconname: 'exclamation-triangle',
     classname: 'error',
   },
 };
 
-const HealthBody = ({ data }) => {
-  const status = get(data, 'healthy.data.result[0].value[1]', '3');
+export const HealthItem = ({ data }) => {
```
rawagner: HealthItem should receive status (OK/WARNING/ERROR) and message so it can be reused for all dashboards

AyushAmbastha:
@rawagner Agreed, Will amend the PR. Thanks!

src/components/Dashboard/Health/HealthItem.js
```diff
@@ -3,31 +3,33 @@ import PropTypes from 'prop-types';
 import { Icon } from 'patternfly-react';
 import { get } from 'lodash';
 
+import { HEALTHY, DEGRADED, ERROR, UNKNOWN } from './strings';
+
 const HealthStatus = {
   0: {
-    message: 'OCS is Healthy',
+    message: HEALTHY,
```
suomiy: is the key always a number 0-3. Can you use an array instead?

src/components/Dashboard/Health/HealthItem.js
```diff
@@ -3,31 +3,33 @@ import PropTypes from 'prop-types';
...
   },
 };
 
-const HealthBody = ({ data }) => {
-  const status = get(data, 'healthy.data.result[0].value[1]', '3');
+export const HealthItem = ({ data }) => {
+  const status = get(data, 'healthy', '3');
   const klass = HealthStatus[+status];
   return klass ? (
     <React.Fragment>
```
suomiy: still has ocs classNames

src/components/Dashboard/Health/HealthItem.js
suomiy: > nevermind

may not be needed once @rawagner suggestion is implemented

> HealthItem should receive status (OK/WARNING/ERROR) and message so it can be reused for all dashboards

src/components/Dashboard/Health/strings.js
AyushAmbastha: What do you suggest the messages should be? Is 'System is healthy' fine?

rawagner: just keep these strings in Storage Overview, we do not need them in generic Dashboard Health. When you will want to use Dashboard Health you will specify the message

AyushAmbastha:
@rawagner @suomiy , I've made the requested changes and amended the PR. Please review.

BEFORE-

Old-health

AFTER-

New-health


The changes I've made are-
1. Made health card generic by adding Heath in the Dashboard folder. HealthItem now receives `icon`, `message` and `classname`. These values are passed after being mapped specific for Cluster or Storage Overview.
2. Made a change in the dashoard.scss file to remove the gap between the Health and the Compliance card.
3. Wrote fixtures and tests for compliance card by adding new file 'ComplianceBody.js' and removed complianceData from the index.js file.
4. Seperate 'strings.js' file for Cluster and Storage Overview which is used when Health card is rendered.

Reference Doc - https://docs.google.com/presentation/d/15nDG3VuQRy6vJbMX4yq6ZvJ5G2Qawaz9glnLT8GEvX4/edit#slide=id.g4fa43fba25_0_0

src/components/Dashboard/Health/HealthItem.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Icon } from 'patternfly-react';
+
+export const HealthItem = ({ icon, classname, message }) => (
+  <React.Fragment>
```
rawagner: no need for `React.Fragment`

src/components/Dashboard/Health/HealthItem.js
```diff
@@ -0,0 +1,24 @@
...
+    <div className="kubevirt-ocs-health__icon">
```
rawagner: all styles are refering to `ocs`. We should use `kubevirt-dashboard-health` styles

src/components/ClusterOverview/Health/Health.js
```diff
@@ -1,26 +1,56 @@
...
-  <DashboardCard>
-    <DashboardCardHeader>
-      <DashboardCardTitle>Cluster Health</DashboardCardTitle>
-    </DashboardCardHeader>
-    <DashboardCardBody className="kubevirt-health__body" isLoading={!loaded} LoadingComponent={InlineLoading}>
-      <HealthBody data={data} />
-    </DashboardCardBody>
-  </DashboardCard>
-);
+const HealthStatus = {
```
rawagner: Please do not change data format from the previous one, we are not using the same as in OCS

src/components/ClusterOverview/Compliance/ComplianceBody.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Icon } from 'patternfly-react';
+
+const healtyIcon = <Icon type="fa" size="2x" name="check-circle" className="kubevirt-ocs-health__icon--ok" />;
```
rawagner: these styles are ocs specific

src/components/Dashboard/Health/HealthItem.js
```diff
@@ -0,0 +1,24 @@
...
+      <Icon type="fa" size="2x" name={icon} className={`kubevirt-ocs-health__icon--${classname}`} />
+      <span className="kubevirt-ocs-health__row-status-item-text">{message}</span>
+    </div>
+  </React.Fragment>
+);
+
+HealthItem.defaultProps = {
+  message: 'Cant get health',
```
rawagner: I dont think these are sensible default props. I would just mark all props as required

src/components/Dashboard/Health/HealthItem.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Icon } from 'patternfly-react';
+
+export const HealthItem = ({ icon, classname, message }) => (
```
rawagner: Health item should be able to display loading state

src/components/ClusterOverview/Health/Health.js
rawagner: I would split the PR to two - first one would contain new generic HealthBody/Item for Dashboard + Health in StorageOverview. I will create second one which changes ClusterOverview health to use HealthBody/Item from Dashboard - as this will probably require changes in web-ui too

AyushAmbastha:
@rawagner Rolled back the changes for the cluster overview and made the requested changes.

PR #351, opened 2019-04-12
Use PF4 graphs

rawagner:


sass/components/Dashboard/capacity.scss
```diff
@@ -1,13 +1,16 @@
 .kubevirt-capacity__items {
   display: flex;
-  justify-content: space-between;
+  justify-content: space-evenly;
   flex-wrap: wrap;
 }
 
 .kubevirt-capacity__item {
```
suomiy: the donuts are overflowing outside the card

suomiy: and the items are all jumbled together on smaller devices

suomiy: \+ the graphs of all items should be centered similarly

sass/components/Dashboard/utilization.scss
```diff
@@ -26,12 +26,20 @@
...
+.kubevirt-utilization__item-chart--narrow {
   text-align: center;
 }
 
-.kubevirt-utilization__item-chart {
+.kubevirt-utilization__item-chart--wide {
   border-left-width: 1px;
   border-left-style: solid;
   border-left-color: rgb(209, 209, 209);
+  padding-right: 20px;
```
suomiy: rem/em?

src/components/Dashboard/Capacity/CapacityItem.js
```diff
@@ -27,43 +21,38 @@ export const CapacityItem = ({ id, title, used, total, unit, formatValue, Loadin
...
-          }}
-          unloadBeforeLoad
-          tooltip={tooltip}
-          title={{
-            primary: primaryTitle,
-            secondary: secondaryTitle,
-          }}
-        />
+      <div id={prefixedId('donut-chart', id)} className="kubevirt-capacity__graph">
+        <svg viewBox="0 0 200 200">
```
suomiy: can we make variables out of the height and the width?

src/components/Dashboard/Capacity/CapacityItem.js
```diff
@@ -27,43 +21,38 @@ export const CapacityItem = ({ id, title, used, total, unit, formatValue, Loadin
...
+          <ChartDonut
+            data={data}
+            labels={datum => `${datum.x}: ${datum.y}`}
+            theme={ChartTheme.light.blue}
+            height={200}
+            width={200}
+            standalone={false}
+          />
+          <ChartLabel textAnchor="middle" style={{ fontSize: 25 }} x={100} y={85} text={primaryTitle} />
+          <ChartLabel textAnchor="middle" style={{ fontSize: 15 }} x={100} y={110} text={secondaryTitle} />
```
suomiy: and derive these from these variables?

src/components/Dashboard/Utilization/UtilizationItem.js
```diff
@@ -49,13 +51,21 @@ export class UtilizationItem extends React.PureComponent {
...
-      };
+      const chartData = data.map((val, index) => ({ x: index, y: Number(val.toFixed(1)) }));
       actual = `${Math.round(data[data.length - 1])} ${unit}`;
-      chart = <SparklineChart id={id} data={chartData} axis={axis} unloadBeforeLoad />;
+      chart = (
+        <Chart
+          id={id}
+          height={60}
+          width={300}
+          containerComponent={<ChartVoronoiContainer labels={datum => `${datum.y} ${unit}`} />}
```
suomiy: the label is huge! Can something be done about that? On the other hand the pies' labels are quite small.

sass/components/Dashboard/utilization.scss
rawagner: I actually removed the padding in latest patch as it seems to be not necessary

src/components/Dashboard/Capacity/CapacityItem.js
rawagner: created constants out of width and height

src/components/Dashboard/Capacity/CapacityItem.js
rawagner: done

sass/components/Dashboard/capacity.scss
rawagner: The issue is happening in cosmos playground - i think that we need to import base.css from patternfly-react but them Im facing issue with webpack not being able to load fonts..

src/components/Dashboard/Utilization/UtilizationItem.js
rawagner: I tried to make the tooltips similar, but it will need more work in the future

src/components/Dashboard/Capacity/CapacityItem.js
```diff
@@ -27,43 +24,63 @@ export const CapacityItem = ({ id, title, used, total, unit, formatValue, Loadin
...
+            style={{ fontSize: 25 }}
+            x={CHART_WIDTH / 2}
+            y={CHART_HEIGHT / 2.3}
+            text={primaryTitle}
+          />
+          <ChartLabel
+            textAnchor="middle"
+            style={{ fontSize: 15 }}
+            x={CHART_WIDTH / 2}
+            y={CHART_HEIGHT / 1.8}
```
suomiy: shouldn't this reather be `CHART_HEIGHT / 2 - SOME_CONSTANT` ? and the `x` one also?

sass/components/Dashboard/capacity.scss
suomiy: can we fix it so it doesn't break without that css?

mareklibra:
Let's fix styling in cosmos in a follow-up. Full build works fine and fix is required asap.

PR #354, opened 2019-04-15
aligned data resiliency,capacity and performance cards

afreen23:
alignednow

### DIFF
--- a/sass/components/Dashboard/utilization.scss
+++ b/sass/components/Dashboard/utilization.scss
@@ -35,3 +35,7 @@
   border-left-style: solid;
   border-left-color: rgb(209, 209, 209);
 }
+
+.kubevirt-utilization__card {
+  height: 23.38em;
+}
--- a/src/components/StorageOverview/Utilization/Utilization.js
+++ b/src/components/StorageOverview/Utilization/Utilization.js
@@ -36,7 +36,7 @@ export const Utilization = ({ iopsUtilization, latencyUtilization, throughputUti
       <DashboardCardHeader>
         <DashboardCardTitle>Perfrormance</DashboardCardTitle>
       </DashboardCardHeader>
-      <DashboardCardBody>
+      <DashboardCardBody className="kubevirt-utilization__card">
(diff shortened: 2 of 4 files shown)


sass/components/Dashboard/utilization.scss
```diff
@@ -35,3 +35,7 @@
   border-left-style: solid;
   border-left-color: rgb(209, 209, 209);
 }
+
+.kubevirt-utilization__card {
```
suomiy: I don't think this is gonna work on all devices. + this does not belong to `Dashboard/`

suomiy:
why does the performance need to be the same height in the first place?  Also, if you want the resiliency and capacity the same height, you should put them in the same card and fix them with css (they can still look like two cards).

suomiy:
>  , you should put them in the same card grid

please see Health & Compliance in ClusterOverview for inspiration

andybraren:
> > , you should put them in the same card grid
> 
> please see Health & Compliance in ClusterOverview for inspiration

Looping in @yuvalgalanti. I wasn't a part of these discussions, but I believe the group decided to visually separate the Health and Data Resiliency cards to not imply any connection between the two. The combined Health/Compliance card of the Cluster Dashboard can stay as-is for now.

jelkosz:
Looping @matthewcarleton in since he is now enhancing the styles to make sure it is in line with his work.

suomiy:
> the group decided to visually separate the Health and Data Resiliency cards to not imply any connection between the two. 

I wasn't suggesting that.  I just mentioned Health & Compliance as an example how to do the layout.

afreen23:
should I close this PR since for now alignment is not possible ..in feedback somebody suggested but on setting that recovery rate is cutting...

andybraren:
@suomiy Ah, I misunderstood, thank you for clarifying.

matthewcarleton:
@afreen23 I think we can close this now. I've updated the layout to just use a grid gap which helps with the spacing.

PR #356, opened 2019-04-15
Add new host states

flofuchs:
Adding 3 new provisioning states:
- match profile (progress state)
- provisioning error (error state)
- power management error (error state)

### DIFF
--- a/src/components/BareMetalHosts/fixtures/BaremetalHostStatus.fixture.js
+++ b/src/components/BareMetalHosts/fixtures/BaremetalHostStatus.fixture.js
@@ -1,30 +1,23 @@
 import { BaremetalHostStatus } from '../BaremetalHostStatus';
 
+const createFixture = (name, state) => ({
+  component: BaremetalHostStatus,
+  name,
+  props: {
+    host: {
+      status: {
+        provisioning: {
+          state,
+        },
+      },
+    },
+  },
+});
+
 export default [
-  {
-    component: BaremetalHostStatus,
-    name: 'Show Generic Status',
-    props: {
-      host: {
-        status: {
-          provisioning: {
-            state: 'unknown state',
-          },
-        },
-      },
-    },
-  },
-  {
-    component: BaremetalHostStatus,
-    name: 'Show Generic Success',
-    props: {
-      host: {
-        status: {
-          provisioning: {
-            state: 'provisioned',
-          },
-        },
-      },
-    },
-  },
+  createFixture('Show Generic Status', 'unknown state'),
+  createFixture('Show Generic Success', 'provisioned'),
+  createFixture('Show progress state for `match profile` status', 'match profile'),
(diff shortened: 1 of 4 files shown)


src/components/BareMetalHosts/fixtures/BaremetalHostStatus.fixture.js
```diff
@@ -27,4 +27,43 @@ export default [
       },
     },
   },
+  {
+    component: BaremetalHostStatus,
+    name: 'Show progress state for `match profile` status',
+    props: {
+      host: {
```
suomiy: please can you create a helper function which creates this structure?

src/components/BareMetalHosts/tests/BaremetalHostStatus.test.js
```diff
@@ -17,4 +20,16 @@ describe('<BaremetalHostStatus />', () => {
     const component = shallow(testBaremetalHostStatusSuccess());
     expect(component).toMatchSnapshot();
   });
+  it('renders a match profile state', () => {
```
suomiy: it is expanding.. Can you please match the snapshots in a for loop. You can look at VmStatus.test.js for inspiration.

flofuchs: Thanks for the hint. Done

src/components/BareMetalHosts/fixtures/BaremetalHostStatus.fixture.js
flofuchs: Done

src/components/BareMetalHosts/tests/BaremetalHostStatus.test.js
```diff
@@ -5,16 +5,13 @@ import { BaremetalHostStatus } from '..';
...
 
 describe('<BaremetalHostStatus />', () => {
-  it('renders a generic status message', () => {
-    const component = shallow(testBaremetalHostStatus());
-    expect(component).toMatchSnapshot();
-  });
-  it('renders a generic success message', () => {
-    const component = shallow(testBaremetalHostStatusSuccess());
-    expect(component).toMatchSnapshot();
+  it('renders the correct subcomponent', () => {
```
suomiy: it should be inside the forEach for easier debugging  + please also include the fixture index in `it` message + optionally the status

flofuchs: Good point, thanks. I made the message more descriptive by adding the status but omitted the index because that would make the message ugly and wouldn't add to its meaning.

suomiy: index add uniqueness to the test case. This will break once you check for the same statuses in multiple fixtures (different conditions - same status)

PR #369, opened 2019-04-16
check if value is null for utilization graph

cloudbehl:
@rawagner @suomiy sometime value provided by prometheus is null on the particular timestamp. So if that's the case the storage dashboard is breaking.

### DIFF
--- a/src/components/Dashboard/Utilization/UtilizationItem.js
+++ b/src/components/Dashboard/Utilization/UtilizationItem.js
@@ -50,7 +50,7 @@ export class UtilizationItem extends React.PureComponent {
     if (isLoading) {
       chart = <LoadingComponent />;
     } else if (data) {
-      const chartData = data.map((val, index) => ({ x: index, y: Number(val.toFixed(1)) }));
+      const chartData = data.map((val, index) => ({ x: index, y: val ? Number(val.toFixed(1)) : 0 }));
       actual = `${Math.round(data[data.length - 1])} ${unit}`;
       chart = (
         <Chart


src/components/Dashboard/Utilization/UtilizationItem.js
```diff
@@ -50,7 +50,7 @@ export class UtilizationItem extends React.PureComponent {
     if (isLoading) {
       chart = <LoadingComponent />;
     } else if (data) {
-      const chartData = data.map((val, index) => ({ x: index, y: Number(val.toFixed(1)) }));
+      const chartData = data.map((val, index) => ({ x: index, y: Number(val && val.toFixed(1)) }));
```
rawagner: how is this `x`  without `y` value going to visualize in graph ? I think it would be better to first filter all data without value

mareklibra: I would move this logic upwards, i.e. to `getUtilizationVectorStats()` or to other place where the source data are being parsed.

Please note, numeric zero `0` is valid value here but it is `false` if transformed to boolean.

rawagner: `parseNumber` from `utils` can be used here

mareklibra: IIUC, `val` can be null. In that case `null.toFixed(1)` will fail.
In other flows, the `y` can become `null` which is not preferred either.

**I still prefer moving this logic to the "parser".**
Or at least: `y: val ? Number(val.toFixed(1)) : 0` . This is no 100% but covers expected cases (expects `val` to be numeric or falsish).

PR #370, opened 2019-04-16
Render Add Host link as host status when in discovered state

jtomasek:
This adds a temporary placeholder link which should in future point
to dialog which lets user register a newly discovered host

### DIFF
--- a/src/components/BareMetalHosts/BaremetalHostStatus.js
+++ b/src/components/BareMetalHosts/BaremetalHostStatus.js
@@ -2,9 +2,17 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { getHostStatus } from '../../utils/status/host/hostStatus';
-import { GenericStatus, GenericError, GenericProgress, GenericSuccess, ValidationError } from './StatusComponents';
+import {
+  AddDiscoveredHostLink,
+  GenericStatus,
+  GenericError,
+  GenericProgress,
+  GenericSuccess,
+  ValidationError,
+} from './StatusComponents';
 import {
   HOST_STATUS_VALIDATION_ERROR,
+  HOST_STATUS_DISCOVERED,
   HOST_STATUS_ALL_PROGRESS,
   HOST_STATUS_ALL_ERROR,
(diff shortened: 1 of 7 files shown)


suomiy:
> This adds a temporary placeholder link which should in future point
to dialog which lets user register a newly discovered host

Shouldn't this rather be an action? It is a bit weird IMO to expect the user to click on the status for an action. Especially when the rest of the UI behaves differently.

jtomasek:
> > This adds a temporary placeholder link which should in future point
> > to dialog which lets user register a newly discovered host
> 
> Shouldn't this rather be an action? It is a bit weird IMO to expect the user to click on the status for an action. Especially when the rest of the UI behaves differently.

This is done to match the wireframes. I think the goal is to make it as easy as possible to figure out what the next action with a newly discovered node is.

suomiy:
> > > This adds a temporary placeholder link which should in future point
> > > to dialog which lets user register a newly discovered host
> > 
> > 
> > Shouldn't this rather be an action? It is a bit weird IMO to expect the user to click on the status for an action. Especially when the rest of the UI behaves differently.
> 
> This is done to match the wireframes. I think the goal is to make it as easy as possible to figure out what the next action with a newly discovered node is.

@andybraren could you please give explanation to that? I am a bit worried about the UI consistency.

andybraren:
@suomiy @jtomasek's description is correct. Exposing the obvious/necessary next step was the goal, and we do something similar in the Machines column if an autoscaler isn't available to attach a Machine to the Host automatically.

We agree that this behavior, along with all the other functionality that our MetalKube designs have put into Status cells (like clicking an error to see info, or clicking in-progress statuses to see details) is currently a little inconsistent. We'll be coordinating with the OKD designers on this, and will likely make updates to their [Resource Statuses design doc](http://openshift.github.io/openshift-origin-design/web-console/4.0-designs/status/status) in the future.

(My guess is that we may change the text "Add host" to something like "Discovered" in the future and make the "Add Host" an action within a popover and/or the action kebab, but for now this PR is fine to merge.)

PR #376, opened 2019-04-16
Dashboards - Fix loading disks

rawagner:
Fixed loading disks by as prop names changed for Overview Dashboard. Im also showing 0 Disks (not loading) in case we have response but it cannot be parsed (ie prometheus is not available or disk metrics are not in prometheus etc)


honza:
#373

PR #377, opened 2019-04-16
WIP Make capacity items smaller

rawagner:


sass/components/Dashboard/capacity.scss
```diff
@@ -9,8 +9,8 @@
   flex-direction: column;
   text-align: center;
   flex-grow: 0;
-  width: 11em;
-  height: 15em;
+  width: 9em;
```
suomiy: why do we set the width and height like this? Shouldn't this rather depend on the donut chart which has a set width and height, 200px x 200px? and add padding/margin to that?

rawagner: AFAIK viewbox sets width and height within SVG image which is then scaled to fit the container. I'm trying to fit all 4 capacity items into one row on smaller display which will be used for demo. But we need to test it yet.

suomiy: in that case I think it might be easier to use row with `3|3|3|3` cols for smaller displays (resolved by query). I would try to match it all by the queries (sm, md, lg) if possible. IF not, we could also resolve half by query and half by flexbox.

lizsurette: > why do we set the width and height like this? Shouldn't this rather depend on the donut chart which has a set width and height, 200px x 200px? and add padding/margin to that?

@matthewcarleton any thoughts here? We see on even regular mac displays at full screen that each capacity chart takes up so much space that there is spill over and it's pushing the utilization card off of the screen completely. Any good ideas on reducing the size of these capacity charts to that 4 will fit at a normal display width?

matthewcarleton: @lizsurrette I have been looking at this today. Let me keep digging and figure out the best approach. I agree setting width/height is probably not ideal.

matthewcarleton:
@rawagner this should be covered in [this pr](https://github.com/kubevirt/web-ui-components/pull/383)

rawagner:
closing in favor of https://github.com/kubevirt/web-ui-components/pull/383

PR #382, opened 2019-04-17
Remove Overview/HealthBody and use Dashboard/HealthBody in Compliace

rawagner:


src/components/ClusterOverview/index.js
```diff
@@ -4,9 +4,5 @@ export { ClusterOverviewContext } from './ClusterOverviewContext';
 // TODO move all these mocked data to component fixtures once they are not needed in web-ui
 // we do not include fixture files in build so we cannot reuse their mocked data
 export const complianceData = {
```
mareklibra: Can we get rid of this at all?

rawagner: Compliance is still just hardcoded, so no, we cannot - or I could hardcode it into compliance component itself (and not connect it to provider).

mareklibra: ok, got it. Let's keep it then.

PR #383, opened 2019-04-17
details/charts

matthewcarleton:


matthewcarleton:
@rawagner just need some help on the class I'm adding on the `dl`

sass/components/Dashboard/capacity.scss
```diff
@@ -29,9 +28,5 @@
 .kubevirt-capacity__graph {
   display: flex;
   flex-direction: column;
-
-  svg {
```
matthewcarleton: @rawagner did you say we needed this overflow?

rawagner: Yes, otherwise tooltips are cut see 
svg_tooltip

rawagner:
I see a few issues

1. if Health is showing multiple errors with additional message it makes height of Health and Compliance card different
health_css

2. Storage dashboard - Capacity and Data Resiliency cards are not aligned anymore
storage_size


Arent those Capacity charts too small ? Ie CPU - 12% Used - the text is really tiny

sass/components/Dashboard/capacity.scss
matthewcarleton: updated

matthewcarleton:
> I see a few issues
> 
> 1. if Health is showing multiple errors with additional message it makes height of Health and Compliance card different
>    health_css
> 2. Storage dashboard - Capacity and Data Resiliency cards are not aligned anymore
>    storage_size
> 
> Arent those Capacity charts too small ? Ie CPU - 12% Used - the text is really tiny
@rawagner 
The height issue should be fixed now. I have a react issue that I'll need some help with :)
I made the charts a bit wider, it's a tricky balance of keeping them narrow enough so they fit on one line. We could adjust the one capacity chart for storage to be larger but that might be a little strange.

rawagner:
@matthewcarleton fixed the build
1. and 2. are resolved
I see that Events card has horizontal scroll now which is not what we want right ? It can even be seen on this screenshot https://github.com/kubevirt/web-ui-components/pull/383#discussion_r277987470 - Chrome  72.0.3626.109 Full HD display

rawagner:
> > @rawagner
> > The height issue should be fixed now. I have a react issue that I'll need some help with :)
> > I made the charts a bit wider, it's a tricky balance of keeping them narrow enough so they fit on one line. We could adjust the one capacity chart for storage to be larger but that might be a little strange.

I agree, they should be the same size, otherwise it would be a bit weird. I think that previous size was ok ( maybe scale down to like 80%) but the issue was that the items didn't wrap very well - you could end up with 3 up and 1 down - which can still happen
cap_wrap

I think that if we make the items bigger but make sure that they wrap only to 2 up 2 down 
cap_wrap_ok

or 1 for a row 
cap_wrap_ok2
 
it would be good

matthewcarleton:
@rawagner I agree there is more work to be done here for responsive behaviour. The charts, the column breakdowns in top consumers, and just generally how card content reacts at different viewport sizes. I think it's out of scope for this PR but something that should definitely be addressed post Summit :) Happy to help out with that.

PR #389, opened 2019-04-18
Hotfix missing components/ClusterOverview/compliance

mareklibra:
Forgotten "git add". Proper fix is required.


mareklibra:
@suomiy , please provide a follow-up to fix that properly.
I guess one "git add" was skipped.

suomiy:
not needed and removed in https://github.com/kubevirt/web-ui-components/pull/400

PR #392, opened 2019-04-18
Correct hostname display in VM overview tab

pcbailey:
This PR does the following:

- Changes the "FQDN" label to "Hostname" in the VM overview tab
- If cloud init is used and hostname field is populated, sets vmi.spec.hostname to the hostname provided
- Sets vmi.spec.hostname to the VM name is cloud init isn't being used or the hostname field is empty

Bug: https://bugzilla.redhat.com/1688124

### DIFF
--- a/src/components/Details/VmDetails/VmDetails.js
+++ b/src/components/Details/VmDetails/VmDetails.js
@@ -16,10 +16,10 @@ import {
   getWorkloadProfile,
   getVmiIpAddresses,
   getOperatingSystemName,
-  getHostName,
   getFlavor,
   getId,
   getDescription,
+  getHostname,
 } from '../../../selectors';
 import {
   getUpdateDescriptionPatch,
@@ -179,8 +179,7 @@ export class VmDetails extends React.Component {
     const vmIsOff = isVmOff(vmStatus);
     const nodeName = getNodeName(launcherPod);
     const ipAddresses = vmIsOff ? [] : getVmiIpAddresses(vmi);
-    const hostName = getHostName(launcherPod);
-    const fqdn = vmIsOff || !hostName ? DASHES : hostName;
+    const hostname = (!vmIsOff && getHostname(vmi)) || DASHES;
     const template = getVmTemplate(vm);
     const id = getId(vm);
     const sortedBootableDevices = getBootableDevicesInOrder(vm);
@@ -288,8 +287,8 @@ export class VmDetails extends React.Component {
             </dl>
             <div className="kubevirt-vm-details__other-details">
               <dl className="kubevirt-vm-details__details-list">
-                <dt>FQDN</dt>
-                <dd id={prefixedId(id, 'fqdn')}>{fqdn}</dd>
+                <dt>Hostname</dt>
+                <dd id={prefixedId(id, 'hostname')}>{hostname}</dd>
 
                 <dt>Namespace</dt>
                 <dd id={prefixedId(id, 'namespace')}>{NamespaceResourceLink ? <NamespaceResourceLink /> : DASHES}</dd>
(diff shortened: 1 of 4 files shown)


src/components/Details/VmDetails/VmDetails.js
```diff
@@ -173,8 +173,8 @@ export class VmDetails extends React.Component {
     const vmIsOff = isVmOff(vmStatus);
     const nodeName = getNodeName(launcherPod);
     const ipAddresses = vmIsOff ? [] : getVmiIpAddresses(vmi);
-    const hostName = getHostName(launcherPod);
-    const fqdn = vmIsOff || !hostName ? DASHES : hostName;
+    const hostname_ = getHostname(vmi);
+    const hostname = vmIsOff || !hostname_ ? DASHES : hostname_;
```
suomiy:
```js
(!vmIsOff &&  getHostname(vmi)) || DASHES
``` 
might be simpler

src/k8s/request.js
```diff
@@ -407,26 +407,27 @@ const addNetworks = (vm, template, getSetting, networks) => {
...
 
-    if (getSetting(USE_CLOUD_INIT_CUSTOM_SCRIPT_KEY)) {
-      cloudInit.setUserData(getSetting(CLOUD_INIT_CUSTOM_SCRIPT_KEY));
-    } else {
-      cloudInit.setPredefinedUserData({
-        hostname: getSetting(HOST_NAME_KEY),
-        sshAuthorizedKeys: getSetting(AUTHKEYS_KEY),
-      });
-    }
+  vm.spec.template.spec.hostname = vmHostname || getSetting(NAME_KEY);
```
suomiy: Can we make it more generic? 

I would rather add a method to cloudInit called getHostname. Which would search for a hostname in the whole cloud-config, e.g.: `^hostname:\s+(\S+)`

this way we would also cover user defined scripts


Please see https://cloudinit.readthedocs.io/en/latest/topics/modules.html#set-hostname

Thinks to consider:
- script must start with  `#cloud-config` (we could add support for other types later)
- to take multipart config into account. and find the `#cloud-config` one (mm can there be mutliple ones of this type?).
- fqdn has a priority in cloud config - so we could parse the hostname out of it  (??)
- getHostname should probably call `build` to resolve the headers.

src/components/Details/VmDetails/VmDetails.js
pcbailey: 100% agreed. Done!

src/k8s/request.js
pcbailey: I talked with @jelkosz about parsing user-provided cloud-init scripts while I was working on the feature and he advised that it's outside the scope of this particular bug. We can see about adding support for it in separate PR, though.

src/k8s/request.js
```diff
@@ -366,26 +366,31 @@ const addNetworks = (vm, template, getSetting, networks) => {
...
 
-    if (getSetting(USE_CLOUD_INIT_CUSTOM_SCRIPT_KEY)) {
-      cloudInit.setUserData(getSetting(CLOUD_INIT_CUSTOM_SCRIPT_KEY));
-    } else {
-      cloudInit.setPredefinedUserData({
-        hostname: getSetting(HOST_NAME_KEY),
-        sshAuthorizedKeys: getSetting(AUTHKEYS_KEY),
-      });
-    }
+  vm.spec.template.spec.hostname = getSetting(HOST_NAME_KEY) || getSetting(NAME_KEY);
```
suomiy: you can move this line to the beginning and remove the one at line 370

src/k8s/request.js
suomiy: ok can you please create an issue for that and/or continue on this?

src/k8s/request.js
pcbailey: Done!

PR #396, opened 2019-04-22
Drop VM detail edit values on cancel

pcbailey:
If the user entered values into fields in the edit mode of the VM details screen, clicked the Cancel button, and then re-entered edit mode, the previously entered values would still be populated. This PR backs up the saved values before entering edit mode and restores them if editing is cancelled.

Bug: https://bugzilla.redhat.com/1698474

### DIFF
--- a/src/components/Details/VmDetails/VmDetails.js
+++ b/src/components/Details/VmDetails/VmDetails.js
@@ -1,6 +1,6 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
-import { get } from 'lodash';
+import { cloneDeep, get } from 'lodash';
 import { Button, Alert, FieldLevelHelp, Icon, Tooltip, OverlayTrigger } from 'patternfly-react';
 import classNames from 'classnames';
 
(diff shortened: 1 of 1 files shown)


src/components/Details/VmDetails/VmDetails.js
```diff
@@ -50,14 +50,21 @@ export class VmDetails extends React.Component {
...
   }
 
-  setEditing = editing =>
-    this.setState({
-      editing,
-    });
+  setEditing = editing => this.setState({ editing });
+
+  onCancel = () => {
+    this.setEditing(false);
```
suomiy: can we have just one `setState` call here?

src/components/Details/VmDetails/VmDetails.js
```diff
@@ -50,14 +50,21 @@ export class VmDetails extends React.Component {
...
+    this.setState(prevState => ({ form: prevState.preEditFormValues }));
+  };
+
+  onEdit = () => {
+    this.setState(prevState => ({ preEditFormValues: cloneDeep(prevState.form) }), () => this.setEditing(true));
```
suomiy: and here? is there an advantage calling it  after the first `setState`?

src/components/Details/VmDetails/VmDetails.js
pcbailey: Done!

src/components/Details/VmDetails/VmDetails.js
pcbailey: No. My thought process was to use setEditing() in case some other relevant logic is put there in the future, but in hindsight that's just another version of premature optimization. So, you're right. It's better to just do this as simply as possible right now and worry about that case if it ever happens in the future. =)

src/components/Details/VmDetails/VmDetails.js
```diff
@@ -50,14 +50,20 @@ export class VmDetails extends React.Component {
       k8sError: null,
       templateError: null,
       form: {},
+      preEditFormValues: {},
```
suomiy: nit: can we call it `preEditForm`? So it has the same shape as `form` - it is less confusing IMO.

pcbailey: Done. I struggled with what to call it. I like yours better. =)

PR #405, opened 2019-04-24
Change to y axis of Top Projects card

AyushAmbastha:
Added 2 extra points on the Y-axis and increased domain such that no extra space is seen.
Screenshot from 2019-04-24 14-07-50


rawagner:
@AyushAmbastha do you want to backport this to `v2.0.0-beta.0` ?

AyushAmbastha:
@rawagner I don't know If you find this a critical problem or not, that's why I sent the PR to master at the moment. If you think it should be included in `v2.0.0-beta.0` as well, I'll send the PR there.

PR #406, opened 2019-04-24
v2v: fix connecting to new instance & enhance

suomiy:
- sort vms
- show duplicate vms only once


suomiy:
\+ the versions will be updated to latest once the images in quay get updated

PR #407, opened 2019-04-24
 enhance k8s methods

suomiy:
- rollback already created objects when the request fails
- change results history approach
- enhance resultsTab

### DIFF
--- a/sass/components/Wizard/create-vm-wizard.scss
+++ b/sass/components/Wizard/create-vm-wizard.scss
@@ -14,6 +14,11 @@
   text-align: left;
 }
 
+.kubevirt-create-vm-wizard__result-tab-row--error {
+  background-color: #ffe6e6;
+  border-color: #c9190b;
+}
+
 .kubevirt-create-vm-wizard__custom-cloud-script-textarea {
   min-height: 100px;
 }
--- a/src/components/Wizard/CreateVmWizard/ResultTabRow.js
+++ b/src/components/Wizard/CreateVmWizard/ResultTabRow.js
@@ -1,8 +1,9 @@
 import React from 'react';
+import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import { ExpandCollapse } from 'patternfly-react';
 
-export const ResultTabRow = ({ title, content, expanded }) => {
+export const ResultTabRow = ({ title, content, isExpanded, isError }) => {
   const resolvedContent = typeof content === 'object' ? JSON.stringify(content, null, 1) : content;
 
   if (!title && !content) {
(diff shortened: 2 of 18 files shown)


suomiy:
mainly for v2v flow, but applies for other flows as well

sass/components/Wizard/create-vm-wizard.scss
```diff
@@ -14,6 +14,11 @@
   text-align: left;
 }
 
+.kubevirt-create-vm-wizard__result-tab-row--error {
+  background-color: mistyrose;
+  border-color: red;
```
matthewcarleton: can we use #c9190b here to be consistent.

sass/components/Wizard/create-vm-wizard.scss
```diff
@@ -14,6 +14,11 @@
   text-align: left;
 }
 
+.kubevirt-create-vm-wizard__result-tab-row--error {
+  background-color: mistyrose;
```
matthewcarleton: Can we use #ffe6e6

sass/components/Wizard/create-vm-wizard.scss
suomiy: done

sass/components/Wizard/create-vm-wizard.scss
suomiy: done

src/k8s/util/enhancedK8sMethods.js
```diff
@@ -0,0 +1,121 @@
...
+export class EnhancedK8sMethods {
+  constructor({ k8sCreate, k8sPatch, k8sKill }) {
+    this._history = [];
+    this._k8sCreate = k8sCreate;
+    this._k8sPatch = k8sPatch;
+    this._k8sKill = k8sKill;
+  }
+
+  _appendHistory = (historyItem, enhancedOpts) => {
+    if (!enhancedOpts || !enhancedOpts.disableHistory) {
```
mareklibra: The use of `enhancedOpts` is confusing:
-  effectively, it's boolean. What about `isSomething` naming?
- based on its name, I would expect that it "turns on"  the feature. In fact, it disables it.

src/k8s/util/enhancedK8sMethods.js
```diff
@@ -0,0 +1,121 @@
...
+          break;
+        default:
+          break;
+      }
+    });
+
+    return currentUnfilteredState.filter(a => a);
+  };
+
+  cleanup = async () => {
```
mareklibra: naming: what about using "rollback" terminology?

src/components/Wizard/CreateVmWizard/ResultTabRow.js
```diff
@@ -1,29 +1,39 @@
...
 };
 
 ResultTabRow.propTypes = {
   title: PropTypes.string,
   content: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
-  expanded: PropTypes.bool,
+  isExpanded: PropTypes.bool,
+  isError: PropTypes.bool,
 };
+//
```
mareklibra: leftover?

mareklibra:
I think this is a good idea to implement wrap the low-level methods for rollback.

src/k8s/util/enhancedK8sMethods.js
suomiy: it is called `enhancedOpt` because there is already one set of opts for the request. It is an object so it can be easier to extend in the future.

`disableHistory` is a command to the request. It just tells the request to not store history of it. `IsHistoryDisabled` doesn't sound right to me. 

\+ history is on by default, because that is the meaning of this class

Better name doesn't come to my mind...

src/k8s/util/enhancedK8sMethods.js
suomiy: renamed to rollback

src/k8s/util/enhancedK8sMethods.js
mareklibra: ok
Honestly, I really hope we will not need to farther decorate these low-level function. Sounds reasonable now, but might become nightmare when overused.

PR #411, opened 2019-04-25
plotted y values for performance card graphs

gnehapk:
With these changes, performance card will look like -
performance-card

### DIFF
--- a/src/components/Dashboard/Utilization/UtilizationItem.js
+++ b/src/components/Dashboard/Utilization/UtilizationItem.js
@@ -22,7 +22,7 @@ export class UtilizationItem extends React.PureComponent {
   };
 
   render() {
-    const { id, title, data, maxY, unit, isLoading, LoadingComponent } = this.props;
+    const { id, title, data, maxY, unit, isLoading, LoadingComponent, decimalPoints } = this.props;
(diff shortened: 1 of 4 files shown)


src/components/Dashboard/Utilization/UtilizationItem.js
```diff
@@ -46,12 +46,27 @@ export class UtilizationItem extends React.PureComponent {
...
     if (isLoading) {
       chart = <LoadingComponent />;
     } else if (data) {
       const chartData = data.map((val, index) => ({ x: index, y: val ? Number(val.toFixed(1)) : 0 }));
       actual = `${Math.round(data[data.length - 1])} ${unit}`;
+      if (decimalPoints > -1) {
+        chartAxis = (
+          <ChartAxis
+            dependentAxis
+            tickValues={[0, Number((maxY / 2).toFixed(decimalPoints)), Number(maxY.toFixed(decimalPoints))]}
```
rawagner: if/else block could calculate just `tickValues` which would result in one `ChartAxis` (easier to maintain if the future)

gnehapk:
Done the review changes. @rawagner Please review.

PR #416, opened 2019-04-26
Add status component

rawagner:
Status component is reused for  Vm status, Node status and BareMetalHost status


src/components/VmStatus/VmStatus.js
```diff
@@ -21,62 +20,39 @@ import {
   getVmStatus,
 } from '../../utils/status/vm';
 import { getId, getVmImporterPods } from '../../selectors';
+import { Status, LinkStatus } from '../Status/Status';
```
mareklibra: just `../Status` is enough, let's make use the flexibility given by of the webpack's resolver.

rawagner:
@mareklibra fixed all, see new commit.

PR #421, opened 2019-04-29
v2v: Compose image name dynamically

mareklibra:
Allows downstream patches.

### DIFF
--- a/src/config/index.js
+++ b/src/config/index.js
@@ -1,4 +1,11 @@
+import { get } from 'lodash';
+
+const getRegistryUrl = () => get(window.SERVER_FLAGS, 'Registry', 'quay.io/nyoxi'); // TODO: upstream should be moved under quay.io/kubevirt
+const getV2vImageTag = () => get(window.SERVER_FLAGS, 'ImageTagV2V', 'latest');
+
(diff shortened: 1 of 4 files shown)


suomiy:
prettier complains..

src/config/index.js
```diff
@@ -1,4 +1,19 @@
+import { get } from 'lodash';
+
+const getRegistry = () => {
+  const registry = get(window, 'SERVER_FLAGS.registry');
+  if (!registry) {
```
suomiy: we should not forget to setup SERVER_FLAGS for tests and cosmos next time. Or alternatively set a default value here.

src/config/index.js
```diff
@@ -1,4 +1,19 @@
...
+    throw new Error('kubevirt: window.SERVER_FLAGS.registry is not set!');
+  }
+  return registry;
+};
+
+const getRegistryV2v = () => get(getRegistry(), 'v2v.url');
+const getImageTagConversion = () => get(getRegistry(), 'v2v.conversionTag');
```
suomiy: image tags could have a default value latest

mareklibra:
Info: Working on better approach via ConfigMap

mareklibra:
Finished, @rawagner, @suomiy , can you please have a look?

PR #427, opened 2019-05-02
Converting vm-status from using link to pf4 popover

yaacov:
Converting status from pf-3 "link" component, to pf4 "popover" component.

Ref: http://openshift.github.io/openshift-origin-design/web-console/4.0-designs/status/status.html

Screenshot:
Peek 2019-05-27 22-00


yaacov:
@rawagner @suomiy please review.

yaacov:
The BZ state that we should use the name "Import" :

> because the Web Console describes this process as importing (Provision Source is "Import", user selects "VM to import"), not conversion or migration.

But we already use state "Import" to state importing from an outside repository.

@matthewcarleton can you help with finding a better descriptions for the two "Import" ing states,
first is from outside repository and the second is from a source VM inside the cluster ?

@rawagner @suomiy please keep me honest here about the meaning of the two states.

src/components/VmStatus/VmStatus.js
```diff
@@ -51,7 +51,7 @@ const StateImporting = ({ additionalText, ...props }) => (
 
 const StateV2VConversionInProgress = ({ progress, ...props }) => (
   <LinkStatus icon="import" {...props}>
-    V2V Conversion In Progress
+    Virtual Machine Import In Progress
```
suomiy: yup this is not really distinguishable from `Importing` state few lines above

suomiy:
##### VM_STATUS_IMPORTING
Newly created VM will be booted from PVC/datavolume which is being imported from some source (Container/URL/PVC) by CDI.

##### VM_STATUS_V2V_CONVERSION_IN_PROGRESS  

The whole VM is imported from VMware and all VM disks are being converted from VMware format to libvirt format. I guess it could be also called importing.

But it should be possible to distinguish between these state. Let's say you have 50 VMs and you want to see which ones are being initialized by CDI and which ones are being migrated from VMware.

> @matthewcarleton can you help with finding a better descriptions for the two "Import" ing states,

should there be also different icon for each type ?

matthewcarleton:
@suomiy we could offer two different icons with the same importing string. On hover we could clarify the difference between the two.

suomiy:
@matthewcarleton 
> we could offer two different icons 

+1

> with the same importing string. On hover we could clarify the difference between the two.

I still prefer having two different importing strings, because it is more clear than on hover. Also the new user might not know which icon belongs to which state.

matthewcarleton:
@suomiy ya I agree it would be nice. I think the biggest challenge is space in the list view. We are fairly limited :( If we can keep it short then that could work. Does it makes sense to use "importing" and "converting" with more info on hover? Converting doesn't sound great to me.

andybraren:
@matthewcarleton Regarding hovering over statuses, this hasn't been added to the OpenShift Design repo yet, but we're generally in agreement that clicking to reveal a popover with additional information is probably the more accessible and forward-looking way to go. Here's a design from Metal3 where we do this:

2a-2b-maint-status-messages

Using a slightly different icon to hint that the two "Importing" statuses are different makes sense to me, and clicking could clarify the source and/or progress in a popover.

suomiy:
> ya I agree it would be nice. I think the biggest challenge is space in the list view. We are fairly limited

I tried different viewports and it seems that currently we are limited with 20 characters. Which could of course change once we add more columns.

> Does it makes sense to use "importing" and "converting" with more info on hover? Converting doesn't sound great to me.

Yes, it is not much specific. 

What about `Importing (VMware)` ?


> Regarding hovering over statuses, this hasn't been added to the OpenShift Design repo yet, but we're generally in agreement that clicking to reveal a popover with additional information is probably the more accessible and forward-looking way to go. Here's a design from Metal3 where we do this:

Please take into account that currently there are some redirections (e.g. to virt-launcher pod) when clicking on the status.

 I guess we could incorporate these links into the popover  (there can be multiple links in case of importing) .

matthewcarleton:
thanks @andybraren ya I agree we need to come to terms with the amount of info we can convey through an icon and a short string of text. I like the idea of presenting things in a popover that can offer more clarity.
@suomiy I think importing and importing (VMware) works. We do risk truncating text but even that can hint that there is more info that they should view.
The main purpose of the different string/icon is really to get the user to engage with the popover to help them understand exactly what is going on and I think we achieve that with these options. Once educating they will understand the difference as they scan the list.
I wonder if a colour change to the current icon could work (blue). The only issue is we use the same icon for migrating as well so we might be introducing confusion if we change it for that and not migrating.

matthewcarleton:
@suomiy @yaacov FYI:
Summit-slide-import-variations
I'm not sure we need to change the icon. We currently use that icon for other actions too (migration) to represent progress which is still happening here. It might be a bit odd to have a unique icon just because it's coming from VMWare. WDYT?

yaacov:
Hi, all.

Thanks @matthewcarleton and @andybraren for the mockups :exclamation: 

A - Thanks, "Importing (VMware)" sounds good to me :+1: 
B - Looking for a way to combine "popover" with "linkto" ... :-)

rawagner:
> B - Looking for a way to combine "popover" with "linkto" ... :-)

@yaacov you can use `OverlayStatus` component https://github.com/kubevirt/web-ui-components/blob/master/src/components/Status/Status.js#L22 See also usage in https://github.com/kubevirt/web-ui-components/blob/master/src/components/NodeStatus/NodeStatus.js#L31

andybraren:
> I guess we could incorporate these links into the popover (there can be multiple links in case of importing) .

Sounds good to me, here’s an example of two “actions” in a popover from a Metal3 design:

IMG_0025

> Please take into account that currently there are some redirections (e.g. to virt-launcher pod) when clicking on the status.

@suomiy Oh, is that the only place where clicking on a status redirects to a different page? I think we should standardize on showing a popover when a blue status is clicked and optionally including a link/action within the popover to the “Pod Details”. I definitely want to make sure we clean up any places where the behavior is different from the above.

> I'm not sure we need to change the icon. We currently use that icon for other actions too (migration) to represent progress which is still happening here. It might be a bit odd to have a unique icon just because it's coming from VMWare. I'm going to ask this question at summit get some feedback as well. WDYT?

@matthewcarleton +1, definitely worth asking about at Summit to see if we’re overloading `pficon-in-progress`.

I’m out of the loop regarding the “Migrating” status, but if a different icon for migrating makes sense, maybe `pf-icon-migration` could be used. Migrating still seems like an “In Progress” state though and could be further explained in a popover, so I’m not 100% sure.

I agree that “Importing (VMware)” makes sense, and it should use the same icon as Importing (they both start from the same “Import” Provision Source in the Create VM wizard, after all). The conversion that happens when importing from VMware seems like a technical detail that can be explained in the popover.

suomiy:
> @yaacov you can use OverlayStatus component

I think the designs are refererencing Popover from pf4:
http://patternfly-react.surge.sh/patternfly-4/components/popover/

rawagner:
> > @yaacov you can use OverlayStatus component
> 
> I think the designs are refererencing Popover from pf4:
> http://patternfly-react.surge.sh/patternfly-4/components/popover/

That's fine. We can use the PF3 and later switch the `OverlayStatus` to PF4

suomiy:
> That's fine. We can use the PF3 and later switch the OverlayStatus to PF4

IMO it would be easier to use Popover component because it already has a support for Header, Body and Footer with a look and feel from the designs. 

We can put the status title in the Header and the message with the links into the Body.

yaacov:
WIP until happy with the links inside popover thing.

yaacov:
@rawagner @suomiy @andybraren please review:

Peek 2019-05-07 14-28

src/components/Status/Status.js
```diff
@@ -35,29 +41,70 @@ OverlayStatus.defaultProps = {
...
+    <React.Fragment>
+      <dt>Message:</dt>
+      <dd>{message}</dd>
+    </React.Fragment>
   ) : (
-    <Status icon={icon}>{children}</Status>
+    ''
   );
 
+  const lintSection = linkTo ? (
```
rawagner: typo - linkSection

src/components/Status/Status.js
```diff
@@ -35,29 +41,64 @@ OverlayStatus.defaultProps = {
...
+export const LinkStatus = ({ icon, children, title, information, linkMessage, linkTo, message }) => {
+  if (!(linkTo || title)) {
+    return <Status icon={icon}>{children}</Status>;
+  }
+
+  const infoSection = information ? <dd>{information}</dd> : '';
+
+  const messageSection = message ? (
+    <React.Fragment>
+      <dt>Message:</dt>
```
rawagner: I'd rather do this with `<div>`

src/components/VmStatus/VmStatus.js
```diff
@@ -63,8 +68,13 @@ StateV2VConversionInProgress.propTypes = {
 };
 
 const StateV2VConversionError = ({ ...props }) => (
-  <LinkStatus icon="error-circle-o" {...props}>
-    V2V Conversion Error
+  <LinkStatus
+    icon="error-circle-o"
+    title="Importing (VMWare) Error"
+    information="Virtual Machine is being converted from VMware format to libvirt format."
```
rawagner: shouldn't there be some error message ?

rawagner: Can you add messages to other statuses ? Or at least create a followup issue

yaacov: > Can you add messages to other statuses ?

:+1: I just don't know what to write ...

@matthewcarleton help :-) 
I need help writing messages for the statuses, 
I do not what the message to just repeat the status, but I can't think of something usefull, can you help ?

src/components/Status/Status.js
yaacov: :smile:

src/components/Status/Status.js
yaacov: do we have a class for this format ? should I create a new one ?

rawagner: new one

yaacov: :+1: added a new class ...

src/components/Status/Status.js
```diff
@@ -1,8 +1,14 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
-import { Icon, OverlayTrigger, Button } from 'patternfly-react';
+import { Icon, OverlayTrigger, Button, Popover } from 'patternfly-react';
```
suomiy: please use Popover from '@patternfly/react-core'

it should be easier to use and there should not be a need for `OverlayTrigger` if the trigger is just `onClick`

src/components/Status/Status.js
```diff
@@ -35,29 +41,64 @@ OverlayStatus.defaultProps = {
...
 };
 
-export const LinkStatus = ({ icon, children, linkMessage, linkTo }) =>
-  linkTo ? (
-    <Status icon={icon}>
-      <Link to={linkTo} title={linkMessage}>
-        {children}
-      </Link>
-    </Status>
+export const LinkStatus = ({ icon, children, title, information, linkMessage, linkTo, message }) => {
```
suomiy: The LinkStatus needs to be more expressive. You should test this with multiple links and progresses.

I would suggest API like this 

```jsx
<LinkStatus icon={icon} title={title} message={message} >
  <StatusField title={title}>  {/*  optional - name of the pod? */}
    <Progress progres={progress} />  {/* this can  be progress bar in the future */}
    <Link linkTo={linkTo} /> {/*  link to the pod? */}
  </StatusField>
  <StatusField>
    <Link />
  </StatusField>
</LinkStatus>
```

src/components/VmStatus/VmStatus.js
```diff
@@ -50,8 +50,13 @@ const StateImporting = ({ additionalText, ...props }) => (
 );
 
 const StateV2VConversionInProgress = ({ progress, ...props }) => (
-  <LinkStatus icon="import" {...props}>
-    V2V Conversion In Progress
+  <LinkStatus
+    icon="import"
+    title="Importing (VMWare)"
+    information="Virtual Machine is being converted from VMware format to libvirt format."
```
suomiy: btw I think these messages could be passed from `vmStatus.js` and changed accordingly there (can be customized if the message is parsed from pod, etc.)

src/components/VmStatus/VmStatus.js
yaacov: add some info for states:
@rawagner @matthewcarleton please review

src/components/Status/Status.js
suomiy: thoughts?

yaacov: Nice :exclamation:  I like it

src/components/Status/Status.js
suomiy: please look at the documentation on how to use it properly: http://patternfly-react.surge.sh/patternfly-4/components/popover/

...I am just going along with the documentation and have not used it myself >.<

andybraren: Using the PF4 Popover component with its built-in header, body, and footer would be great! Here’s another link: https://www.patternfly.org/v4/documentation/react/components/popover/

andybraren:
Hi again, my how this PR has grown 😄

@matthewcarleton is likely quite busy at Summit this week and would have a better idea of status message text, but here are some things I noticed from your GIF (thank you for posting it):

1. The title of the popover should be the same as the text of the status.
2. The "Link" at the bottom should be hyperlinked text like "View pod events" or "View VM details" instead of showing the path.
3. I don't think the word "Message" needs to be shown - the message is just whatever text appears within the popover. The percent completion when importing could be shown within a `<p>` until we (maybe) start using progress bars in the UI.

Here's a mockup using PF4's popover. The exact text and where the links should point to are all up for debate - this is just a general idea.

bit-pf4-status-popovers

suomiy:
@andybraren 

looks good. just please take into account there can be multiple links/progresses  - which are usually tied together (through some pod). Not sure, if to show that in the footer or in the body together, as suggested in https://github.com/kubevirt/web-ui-components/pull/427#discussion_r281599733

yaacov:
@andybraren wow, thanks ! your mockups are very helpful :+1:

matthewcarleton:
Thanks @andybraren it's a busy week here at Summit. +1 to everything you pointed out. If we have multiple links that can not be self explained by their text I'd suggest a descriptive sentence with a link embedded in the text. 

For example: The virtual machine could not be imported from VMWare. Go to the details page to learn more (link).

This approach is the best a11y option.

andybraren:
Here's another quick mockup of what progress bars and two actions in the footer could look like (wrapping at smaller screen sizes is fine). As Matt said, including links within the text content of the popover is also a good, accessible approach for things like "Learn more" which could link to help documentation.

I'd suggest that we place any links to object pages inside the footer (like "View pod details") and we try to keep that consistent across all future status popovers we make going forward. At least for now.

Is this along the lines of what you were thinking @suomiy?

bit-pf4-status-popovers

suomiy:
>  If we have multiple links

VMware import has only one link currently. But cdi imports can have multiple.

> I'd suggest that we place any links to object pages inside the footer (like "View pod details") and we try to keep that consistent across all future status popovers we make going forward. At least for now.

> Is this along the lines of what you were thinking @suomiy?

+1

rawagner:
@yaacov @suomiy guys, looking at referenced bugzilla - it is targeting CNV 2.0. Please do not do any big changes such as PF4 components, stick to original scope so we will be able to backport it to stable branch. Followup PR can address new features/PF4 components. @jelkosz

yaacov:
> Please do not do any big changes such as PF4 components, stick to original scope so we will be able to backport it to stable branch

@suomiy @rawagner For this PR continue using the PF3 popover ?

suomiy:
> @suomiy @rawagner For this PR continue using the PF3 popover ?

IMO you can just do a small patch basically swapping `V2V Conversion in Progress` with `Importing (VMware)` and continue with this one for master

yaacov:
> do a small patch basically swapping V2V Conversion in Progress with Importing (VMware) 

Nice ! , will continue here with PF4, and make a new small one for the BZ.

andybraren:
@yaacov @rawagner @suomiy Just to clarify, will the Status Popover implementation use the PF4 design or PF3? I'm wondering what our mockups should include going forward.

It sounds like the 4.2 Console will be switching over to PF4 for all tables, which might make PF3 status popovers look a little out of place.

yaacov:
@andybraren , my vote is PF4, we have  https://github.com/kubevirt/web-ui-components/pull/447 for backward compatibility, so we do not need to worry about it here.

matthewcarleton:
@yaacov @andybraren @suomiy if possible it would be great if we could stick to PF4. We are in the process of updating our designs to align with the PF4 popovers.

yaacov:
Updates:

- Will use the new PF4 `OverlayStatus` from https://github.com/kubevirt/web-ui-components/pull/473
- Will add a `StatusField` component for the different sections inside the OverlayStatus ( https://github.com/kubevirt/web-ui-components/pull/427#discussion_r281599733 )

- Look up best way to add progress  `StatusField`  ( https://github.com/kubevirt/web-ui-components/blob/master/src/components/VmStatus/VmStatus.js#L61 ) ( https://github.com/kubevirt/web-ui-components/pull/427#issuecomment-490099773 )
-  Look up best way to get descriptions and notes from caller components instead of hard coded here.

cc// @rawagner @suomiy

suomiy:
> Updates:
> 
> * Will use the new PF4 `OverlayStatus` from #473
> * Will add a `StatusField` component for the different sections inside the OverlayStatus ( [#427 (comment)](https://github.com/kubevirt/web-ui-components/pull/427#discussion_r281599733) )

+1

> * Look up best way to add progress  `StatusField`  ( https://github.com/kubevirt/web-ui-components/blob/master/src/components/VmStatus/VmStatus.js#L61 ) ( [#427 (comment)](https://github.com/kubevirt/web-ui-components/pull/427#issuecomment-490099773) )

you could use pf4 Progress http://patternfly-react.surge.sh/patternfly-4/components/progress/. 

Although, we are using pf3 `ProgressBar` in dashboard. Not sure which one to choose but  we probably should unite them.

rawagner:
> you could use pf4 Progress http://patternfly-react.surge.sh/patternfly-4/components/progress/.
> 
> Although, we are using pf3 `ProgressBar` in dashboard. Not sure which one to choose but we probably should unite them.


Lets go for PF4 one and use it in dashboard too

suomiy:
\+ I think it would be useful to have the percentage number inside the progress bar

lizsurette:
> + I think it would be useful to have the percentage number inside the progress bar

I think to avoid awkwardness with text-color needing to change based on whether the bar is full or not, we are recommending that the percentage is above the bar in the top right:
https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/progress-bar

src/components/VmStatus/strings.js
```diff
@@ -0,0 +1,25 @@
...
+export const IMPORTING_ERROR_VMWARE = 'Import Error (VMware)';
+export const VIEW_POD_EVENTS = 'View pod events';
+export const VIEW_POD_DETAILS = 'View pod details';
+export const VIEW_VM_EVENTS = 'View VM events';
+export const UNKNOWN = 'Unknown';
+export const OFF = 'Off';
+export const MIGRATING = 'Migration';
+export const IMPORTING_VMWARE_MESSAGE =
+  'The virtual machine is being imported from VMWare. Disks will be converted to the libvirt format.';
+export const IMPORTING_ERROR_VMWARE_MESSAGE = 'The virtual machine could not be imported from VMWare.';
```
apinnick: Typo. VMware, not VMWare.

src/components/VmStatus/strings.js
```diff
@@ -0,0 +1,25 @@
...
+export const IMPORTING_PENDING_VMWARE = 'Import Pending (VMware)';
+export const IMPORTING_ERROR_VMWARE = 'Import Error (VMware)';
+export const VIEW_POD_EVENTS = 'View pod events';
+export const VIEW_POD_DETAILS = 'View pod details';
+export const VIEW_VM_EVENTS = 'View VM events';
+export const UNKNOWN = 'Unknown';
+export const OFF = 'Off';
+export const MIGRATING = 'Migration';
+export const IMPORTING_VMWARE_MESSAGE =
+  'The virtual machine is being imported from VMWare. Disks will be converted to the libvirt format.';
```
apinnick: VMware, not VMWare.

src/components/VmStatus/strings.js
```diff
@@ -0,0 +1,25 @@
...
+export const IMPORTING_ERROR_VMWARE_MESSAGE = 'The virtual machine could not be imported from VMWare.';
+export const IMPORTING_MESSAGE = 'The virtual machine is being imported. Disks will be copied from source image.';
```
apinnick: If the disks are being copied as part of the import process, change to "Disks are being copied from the source image."

src/components/VmStatus/strings.js
yaacov: Thanks :heart:

src/components/VmStatus/strings.js
```diff
@@ -0,0 +1,25 @@
...
+export const IMPORTING_ERROR_MESSAGE = 'The virtual machine could not be imported.';
+export const VMI_WAITING_MESSAGE = 'The virtual machine is waiting for resources to become available.';
+export const STARTING_MESSAGE = 'The virtual machine wait for state to become ready.';
```
apinnick: This looks weird. I'm not sure what it means. What is the VM waiting for? Ready state?

src/utils/status/common.js
```diff
@@ -1 +1,5 @@
 export const NOT_HANDLED = null;
+
+export const STATUS_INFO = 'INFO';
+export const STATUS_SUCCESS = 'SUCCESS';
+export const STATUS_DANGER = 'DANGER';
```
apinnick: What would be considered a danger?

src/components/VmStatus/strings.js
```diff
@@ -0,0 +1,25 @@
...
+export const IMPORTING = 'Importing';
+export const IMPORTING_VMWARE = 'Importing (VMware)';
+export const IMPORTING_PENDING_VMWARE = 'Import Pending (VMware)';
+export const IMPORTING_ERROR_VMWARE = 'Import Error (VMware)';
+export const VIEW_POD_EVENTS = 'View pod events';
+export const VIEW_POD_DETAILS = 'View pod details';
+export const VIEW_VM_EVENTS = 'View VM events';
+export const UNKNOWN = 'Unknown';
+export const OFF = 'Off';
+export const MIGRATING = 'Migration';
```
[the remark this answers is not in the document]
yaacov: Wow, my bad :-) thanks!

src/components/VmStatus/strings.js
yaacov: @apinnick hi, 
This text is problematic because I do not have useful information to add on the abv. version "Starting".

What I want to do here is to give a "long" version of "Starting" if that makes sense ?
Basically, the VM is waiting for the readyness tests to be valid.

Does "Pod is waiting for virtual machine readyness tests to pass." sounds better ?

apinnick: Since the rest of the messages do not mention the pod, maybe best not to mention it here if we can find a workaround. What happens when the VM is 'ready'? It starts running? It migrates?

src/utils/status/common.js
yaacov: @apinnick Thanks!

Ahh, these are not user-facing texts, it's internal constants that are strings.
In this case, it will be shown to the user as one of the "Error" strings.

apinnick: NP

src/components/VmStatus/strings.js
yaacov: > What happens when the VM is 'ready'? It starts running? It migrates?

It starts running :-)

apinnick: OK. I suggest this: "The virtual machine is preparing to start", if it's not important for the user to know about the readiness tests.

yaacov: :+1: :heart:

matthewcarleton: +1 on not mentioning the pod. Can we shorten this to "The virtual machine is starting" ? Do we need to be so specific about the preparing part?

apinnick: If this message does not need to be substantially different from the "Starting" message, sure, that's fine with me.

yaacov:
-  Using the new PF4 PopoverStatus :birthday:
-  Added a StatusField component for the different sections inside the PopoverStatus.
-  Added progress to the popover.
-  Currently using hard-coded status messages 

@rawagner @suomiy please review.

yaacov:
Update PR description:
-  Title is now "Converting vm-status from using link to pf4 popover"
- Added screenshot:
Peek 2019-05-27 17-06

src/components/VmStatus/strings.js
yaacov: @apinnick @matthewcarleton hi, thanks.

Updated to "The virtual machine is starting" , attached screenshots :-)

"The virtual machine is starting"
Screenshot_2019-05-27 VmStatusWrapper Starting – React Cosmos(1)

"The virtual machine is preparing to start"
Screenshot_2019-05-27 VmStatusWrapper Starting – React Cosmos

matthewcarleton: oh hmm now that I'm seeing this I take it back, the second one is probably better. It would be even nicer if we could be a bit more detailed. Is there any other information available on what is happening?
Also, do we want to say 'view pod events` or just `view events` Is this the users first exposure to the VM running on a pod? Would they understand that the pod = the VM? 
Sorry for all the nitpicks :)

yaacov: > Sorry for all the nitpicks :)

Not at all :-) 

> Is there any other information available on what is happening?

a - Starting means, we have all the necessary resources to start the VM, and is currently in the process of creating the vm's storage, net and compute resources.

b - The "fail description Ready" represent a description string we get from k8s API call, It should have some relevant information about this specific vm's state.

> Also, do we want to say 'view pod eventsor justview events` 

I copied it from https://github.com/kubevirt/web-ui-components/pull/427#issuecomment-490099773 , happy to change to what ever makes more sense :smile: 

> oh hmm now that I'm seeing this I take it back,

Yes, seeing things in context changes the view :-) I can make new screen shots with different texts so we can evaluate several options.

matthewcarleton: Thanks @yaacov!
Could we say " This virtual machine will start shortly. Preparing storage, networking, and compute resources"
where does "View Pod events" go? That should influence what this string says IMO :)

yaacov: > Could we say " This virtual machine will start shortly. Preparing storage, networking, and compute resources"

+1 

> where does "View Pod events" go? That should influence what this string says IMO :)

It goes to the page of the "Pod" that runs the VM into the "Events" tab ...

yaacov: @matthewcarleton I like it now :-)

Screenshot_2019-05-27 VmStatusWrapper Starting – React Cosmos

andybraren: Does the VM Details page include an Events tab (like our designs do)? If yes, maybe changing this text to "View VM events" and linking to that page makes the most sense. If not, then "View pod events" seems okay for now.

yaacov: > Does the VM Details page include an Events tab (like our designs do)?

I think the reason we send users to the pod, is that while vm is not yet running, the vm events is empty, ant the pod events are more useful.

@matthewcarleton @rawagner @suomiy maybe we should have two links ( or more ? ) one for each entity that makes sense in this context ?

andybraren: That makes total sense, I'd say "View pod events" then. 👍

Our design does support two actions/links in the footer (example [here](http://openshift.github.io/openshift-origin-design/web-console/4.0-designs/status/status.html#status-popovers)) but since VM events would be empty/useless in this state, just linking to pod events seems correct to me.

andybraren:
The PF4 popovers look great @yaacov, thank you for sharing that (super useful) GIF! 👍

Two (last?) comments:

1. The `Running` status probably doesn't need a popover, since it doesn't provide any additional help or information that the user would find useful. Do you agree @matthewcarleton?

2. The `Running` status should use the `fa-refresh` icon instead of `pficon-on-running`, documented [here](http://openshift.github.io/openshift-origin-design/web-console/4.0-designs/status/status.html). Maybe that's out of scope of this PR, though?

(P.S. @matthewcarleton I mentioned using `pf-icon-migration` instead of `pf-in-progress` for Migrating [3 weeks ago](https://github.com/kubevirt/web-ui-components/pull/427#issuecomment-489479699), and I see this GIF now uses it for Importing too. I'm not against this even though they're similar to "In Progress", but whatever we decide we should update the [VM List](http://openshift.github.io/openshift-origin-design/web-console/knikubevirt/vm-list/vm-list) and [Status Icon Usage](http://openshift.github.io/openshift-origin-design/web-console/4.0-designs/status/status.html#icon-usage) design documentation to keep track).

yaacov:
@andybraren @matthewcarleton hi,

1.  I agree, ( note: it does currently give a link to the pods events )

2. done :-)

p.s.
Updated the PR to use migration icon for the migrations,  and import icon for the Importing, WDYT ?

Peek 2019-05-27 19-57

yaacov:
screenshot of running status without link and with new icon:
Screenshot_2019-05-27 VmStatusWrapper Running – React Cosmos(1)

src/components/VmStatus/strings.js
matthewcarleton: +1 ya that makes sense to me. Pod events it is!

matthewcarleton:
@andybraren +1 on no pop up for the running status :)
@yaacov I think you are using the restart icon for running in that last screenshot. 
Maybe I'm wrong but I thought we were using the `pficon-in-progress` icon for all import/migration status types. 
We have the effort of updating the list view to PF4 in general so that might be a good time to comb over all of this and ensure all the icons align.

yaacov:
@matthewcarleton 

>  I think you are using the restart icon for running in that last screenshot.

It spouse to be `refresh` :-) https://github.com/kubevirt/web-ui-components/pull/427/files#diff-cbd603a3cce7a5976745fe0c00f41583R212, will look into it :+1: 

>  I thought we were using the pficon-in-progress icon for all import/migration status types.

Didn't understand the comments above :-) , np, `pficon-in-progress` icon for all import/migration status types. :cake:  [Icon usage](http://openshift.github.io/openshift-origin-design/web-console/4.0-designs/status/status.html#icon-usage)

yaacov:
@matthewcarleton @andybraren  this is an update screenshot:
Peek 2019-05-27 22-00

state | icon
-------|--------
all migrating and importing + starting | `in-progress`
running | `ok`
off        | `off`
all errors | `error-circle-o`
pending and waiting | `hourglass-half`

andybraren:
I'm not very familiar with Pending or Waiting as they relate to VMs, but using `fa-hourglass-half` like you did makes sense for those labels. And I agree `fa-refresh` should be used for Running.

Thank you @yaacov, looks great to me!

matthewcarleton:
agreed! This looks awesome. Thanks for doing this @yaacov

src/components/Status/Status.js
```diff
@@ -1,67 +1,117 @@
...
+StatusLinkField.defaultProps = {
+  title: null,
+};
+StatusLinkField.propTypes = {
+  title: PropTypes.string,
+  linkTo: PropTypes.string.isRequired,
+};
+
+export const StatusProgressField = ({ title, progress, barType }) => {
+  const variants = {
```
suomiy: this object can be static

we could also export this or/and the constants from here.

 I am not sure the mapping is necessary though - we could have just used the original constants

src/components/VmStatus/VmStatus.js
```diff
@@ -126,74 +102,137 @@ VmStatuses.propTypes = {
...
-        </StateError>
-      );
-    default:
-      return <StateUnknown />; // Let's hope this state is tentative and will fit former conditions soon
-  }
-};
+  const additionalText =
+    verbose && statusDetail.pod && statusDetail.pod.metadata.labels ? getAdditionalImportText(statusDetail.pod) : null;
+
+  const statusProps = {
```
suomiy: IMO I think it would be better to not use the wrapper and create ` <PopoverStatus ... >` or `<Status ...>` accordingly here (we do not need special components like StateV2VConversionInProgress). 

But the approach with the wrapper is a bit confusing (just a lot of `if`s) and doesn't bring much of  additional functionality - so I think it is easier to just declare the components here.

suomiy: plus it is a bit inefficient to create so much markup in each render

src/components/VmStatus/strings.js
```diff
@@ -0,0 +1,24 @@
+export const RUNNING = 'Running';
```
suomiy: +1 for extracting the strings

src/components/VmStatus/VmStatus.js
suomiy: the ifs could be also put inside the components - if progress == null - do not render the component

suomiy: thoughts on all of this?

src/utils/status/vm/constants.js
```diff
@@ -4,6 +4,7 @@ export const VM_STATUS_STARTING = 'VM_STATUS_STARTING';
 export const VM_STATUS_VMI_WAITING = 'VM_STATUS_VMI_WAITING';
 export const VM_STATUS_IMPORTING = 'VM_STATUS_IMPORTING';
 
+export const VM_STATUS_CONVERSION_IN_PROGRESS = 'VM_STATUS_CONVERSION_IN_PROGRESS';
```
suomiy: what is this status for? Who sets it?

src/components/Status/Status.js
```diff
@@ -1,67 +1,117 @@
...
 import { Link } from 'react-router-dom';
-import { Popover } from '@patternfly/react-core';
+import { Popover, Progress, ProgressVariant, ProgressSize } from '@patternfly/react-core';
 import { Icon, Button } from 'patternfly-react';
 import PropTypes from 'prop-types';
 
-export const Status = ({ icon, children }) => (
+import { STATUS_DANGER, STATUS_SUCCESS, STATUS_INFO } from '../../utils/status/common';
+
+const StatusField = ({ content }) => <div className="kubevirt-status__field">{content}</div>;
```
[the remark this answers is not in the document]
yaacov: yes :+1:

src/components/Status/Status.js
yaacov: > I am not sure the mapping is necessary though - we could have just used the original constants

makes sense :smile:

src/components/VmStatus/strings.js
rawagner: VM events shows also launcher pod events see https://github.com/kubevirt/web-ui/blob/master/frontend/public/kubevirt/components/vm/vm-detail.jsx#L39 so we dont have link to pod events page anymore

yaacov: > so we dont have link to pod events page anymore

@rawagner should we:
1 - Convert all links to `vm events` instead of pod events ?
2 - Convert all labels from "View pod events" to just "View events" ?

suomiy: I think it makes sense to show the link to the pod at least in a case of an error - for inspecting the pod

yaacov: @rawagner @suomiy 
All links except pod-error will link to vm event and label "View events"
In case of pod-error link to pod events and label "View pod events" WDYT ?

suomiy: I would link to VM events and to Pod Overview..

src/components/VmStatus/VmStatus.js
yaacov: > thoughts on all of this?

Nice, it is a lighter implementation :dark_sunglasses: , reverting back to switch :smile:

src/utils/status/vm/constants.js
yaacov: > what is this status for? Who sets it?

:cry: My bad, removing,

 :hibiscus: the string is `"VM_STATUS_CONVERSION_IN_PROGRESS"` the constant is `VM_STATUS_V2V_CONVERSION_IN_PROGRESS`

src/components/VmStatus/VmStatus.js
```diff
@@ -17,74 +20,91 @@ import {
...
-);
-
-const StateV2VConversionInProgress = ({ progress, ...props }) => (
-  <LinkStatus icon="import" {...props}>
-    Importing (VMware)
-  </LinkStatus>
+const VmStatusInProgress = ({ header, message, children, progress, linkTo, linkMessage }) => (
+  <PopoverStatus icon="in-progress" header={header}>
+    <StatusDescriptionField>{message}</StatusDescriptionField>
+    {children || ''}
```
suomiy: why does the empty string have to be here?

src/components/VmStatus/VmStatus.js
```diff
@@ -17,74 +20,91 @@ import {
...
+    {progress !== null ? <StatusProgressField title={IMPORTING} progress={progress} /> : ''}
```
suomiy: `progress && <StatusProgressField title={IMPORTING} progress={progress} /> ` ?

src/components/VmStatus/VmStatus.js
```diff
@@ -126,74 +144,140 @@ VmStatuses.propTypes = {
...
+    case VM_STATUS_V2V_CONVERSION_PENDING:
       return (
-        <StateV2VConversionInProgress {...statusDetail} linkTo={getSubPagePath(statusDetail.pod, PodModel, 'events')} />
+        <VmStatusPending
+          header={IMPORTING_PENDING_VMWARE}
+          message={IMPORTING_VMWARE_MESSAGE}
+          linkMessage={VIEW_VM_EVENTS}
+          linkTo={linkToVMEvents}
+        >
+          {statusDetail.message ? <StatusDescriptionField>{statusDetail.message}</StatusDescriptionField> : ''}
```
suomiy: can we do without empty strings?

src/components/VmStatus/VmStatus.js
```diff
@@ -126,74 +144,140 @@ VmStatuses.propTypes = {
...
+      );
+    case VM_STATUS_IMPORT_ERROR:
+      return (
+        <VmStatusError
+          header={IMPORTING_ERROR}
+          message={IMPORTING_ERROR_MESSAGE}
+          linkMessage={VIEW_VM_EVENTS}
+          linkTo={linkToVMEvents}
+        >
+          <React.Fragment>
```
suomiy: no need for fragments

src/components/Status/tests/__snapshots__/Status.test.js.snap
```diff
@@ -10,3 +10,38 @@ exports[`<Status /> renders correctly 1`] = `
...
+
+exports[`<StatusField> renders correctly 3`] = `
+<Fragment>
+  <div>
+    some less important content
+  </div>
+</Fragment>
+`;
+
+exports[`<StatusField> renders correctly 4`] = `<Fragment />`;
```
suomiy: renders `Fragment` only?

yaacov: @suomiy , Thanks for the  reviews :+1: , removed the unnecessary checks :beers: :cake:

yaacov:
@rawagner @suomiy please review.

p.s.
@suomiy thanks for the empty strings comment, looks much nicer now :beer:

src/components/VmStatus/VmStatus.js
```diff
@@ -19,72 +22,87 @@ import {
...
+};
+VmStatusError.propTypes = {
+  header: PropTypes.string.isRequired,
+  message: PropTypes.string.isRequired,
+  linkMessage: PropTypes.string.isRequired,
+  linkTo: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+  children: PropTypes.any,
 };
 
 export const VmStatuses = props => {
```
suomiy: We should also move the importerPods statuses into the popover. Then we should either show the same exact status + popover in the vmDetails or show the inline version with just the content there and without the popover part. @andybraren @matthewcarleton thoughts?

not sure if you want to do it in this PR or leave it for the followup...

andybraren: For now let's use the same [Status Popover pattern](http://openshift.github.io/openshift-origin-design/web-console/4.0-designs/status/status.html#status-popovers) within VM Details as well (so no inline version).

FYI, we have some longer-term plans to improve Resource Details pages console-wide using some of our Dashboard card-related work, which would allow us to show error statuses/messages inline. But we'll save that for later. Another popover is fine. 🙂

yaacov:
@rawagner ping :-)

src/components/VmStatus/VmStatus.js
matthewcarleton: +1 @andybraren

PR #430, opened 2019-05-04
Delete new row when cancel is clicked

rawagner:
https://bugzilla.redhat.com/show_bug.cgi?id=1646953

### DIFF
--- a/src/components/Table/EditableDraggableTable.js
+++ b/src/components/Table/EditableDraggableTable.js
@@ -43,6 +43,21 @@ class EditableDraggableTable extends React.Component {
     }
   }
 
+  onRowDelete = rowData => {
+    const { id } = rowData;
+    const editing = false;
+    const rows = cloneDeep(this.props.rows.filter(row => row.id !== id));
+
+    this.flagUpdate(rows, editing);
+    this.setState({ editing });
+
+    this.props.onChange(rows, {
+      type: ON_DELETE,
+      editing,
+      id,
+    });
+  };
+
   // needed for refiring row renders
   // eslint-disable-next-line no-return-assign
   flagUpdate = (rows, editingInProgress) => rows.forEach(row => (row.editingInProgress = editingInProgress));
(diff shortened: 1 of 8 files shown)


src/components/Wizard/CreateVmWizard/CreateVmWizard.js
```diff
@@ -116,9 +116,10 @@ export class CreateVmWizard extends React.Component {
 
   lastStepReached = () => this.state.activeStepIndex === this.getLastStepIndex();
 
-  onStepDataChanged = (tabKey, value, valid) => {
+  onStepDataChanged = (tabKey, value, valid, editing = false) => {
     const validatedTabData = validateTabData(tabKey, value, valid);
     this.safeSetState(state => ({
+      editing,
```
suomiy: I think the tab should merely request to lock the createVmWizard and not take Wizard state into account. This is very useful when using asynchronous actions.

I suggest something like:

```js
{
  stepData: {
    [tabKey]: {
      lockStep
    }
  }
}

```

src/components/Table/EditableDraggableTable.js
```diff
@@ -96,17 +97,24 @@ class EditableDraggableTable extends React.Component {
...
-      const index = findIndex(rows, { id });
-
-      rows[index] = cloneDeep(rows[index].backup);
-      delete rows[index].backup;
+      const index = findIndex(this.props.rows, { id });
+      let rows;
+
+      let type = ON_CANCEL;
+      if (this.props.rows[index].newRow) {
+        rows = this.props.rows.filter(row => row.id !== id);
```
suomiy: can't we rather call onDelete method from here?

rawagner: extracted onDelete to separate function which is now reused where needed

src/components/Wizard/CreateVmWizard/CreateVmWizard.js
rawagner: done, `lockStep` is under `[tabKey]` now

PR #435, opened 2019-05-06
Time Axis - Utilization Card (Storage Overview)

AyushAmbastha:
Implemented a common time axis for the Utilization Card.

Screenshot from 2019-05-06 12-25-49

JIRA Task - https://jira.coreos.com/browse/CNV-1654

PS. Will be sending a follow up PR having the tests for the rest of the cards.


src/components/StorageOverview/Utilization/Utilization.js
```diff
@@ -55,13 +61,26 @@ export const Utilization = ({
...
 
   return (
     <DashboardCard>
       <DashboardCardHeader>
-        <DashboardCardTitle>Performance</DashboardCardTitle>
+        <DashboardCardTitle>Utilization</DashboardCardTitle>
       </DashboardCardHeader>
       <DashboardCardBody>
+        <Row>
+          <Col className="kubevirt-utilization__time-duration time-duration--left">Time </Col>
```
rawagner: className should look like `className="kubevirt-utilization__time-duration kubevirt-utilization__time-duration--left"`. Same applies for `right`. It will conform to BEM and there will be no need to disable it in `scss`

AyushAmbastha: I tried that and it still gave the `Invalid component selector` error and hence had to disable it.

rawagner: does not happen here..

AyushAmbastha: `Invalid component selector                                                                          plugin/selector-bem-pattern
          ".kubevirt-utilization__time-duration.kubevirt-utilization__time-duration--left"`

Error ^

rawagner:
```suggestion
          <Col className="kubevirt-utilization__time-duration kubevirt-utilization__time-duration--left">Time </Col>
```

sass/components/Dashboard/utilization.scss
```diff
@@ -50,5 +50,22 @@
 }
 
 .kubevirt-utilization__item-narrow-title {
-  font-size: 0.875rem;
+  font-size: 1rem;
+}
+
+.kubevirt-utilization__time-duration {
```
rawagner:
```suggestion
.kubevirt-utilization__time-duration {
  font-size: 0.8em;
  padding-bottom: 1rem;
}

.kubevirt-utilization__time-duration--left {
  float: left;
  padding-left: 1rem;
}

.kubevirt-utilization__time-duration--right {
  float: right;
  padding-right: 1rem;
}
```

src/components/StorageOverview/Utilization/Utilization.js
rawagner: @AyushAmbastha see suggested change which is passing the validation

AyushAmbastha:
@rawagner Amended the PR with the suggested changes. Please take a look.

src/selectors/prometheus/selectors.js
```diff
@@ -24,6 +24,11 @@ export const getUtilizationVectorStats = response => {
   return values && Array.isArray(values) ? values.map(timeValuePair => parseNumber(timeValuePair[1])) : null;
 };
 
+export const getUtilizationVectorTime = response => {
+  const values = get(response, 'data.result[0].values');
+  return values && Array.isArray(values) ? values.map(timeValuePair => parseNumber(timeValuePair[0] * 1000)) : null;
```
mareklibra: Shouldn't it be `parseNumber(timeValuePair[0]) * 1000` ?

src/components/StorageOverview/Utilization/Utilization.js
```diff
@@ -55,13 +61,28 @@ export const Utilization = ({
...
+        <Row>
+          <Col className="kubevirt-utilization__time-duration kubevirt-utilization__time-duration--left">Time </Col>
+          <Col className="kubevirt-utilization__time-duration kubevirt-utilization__time-duration--right">
+            6 hour average
+          </Col>
+        </Row>
+        <ChartAxis
+          scale={{ x: 'time' }}
+          tickValues={timestamps}
+          tickFormat={x => formatToShortTime(x)}
```
mareklibra: The `x` can be potentially `null`.

src/components/StorageOverview/Utilization/tests/Utilization.test.js
```diff
@@ -1,16 +1,16 @@
...
-import { render } from 'enzyme';
+import { shallow } from 'enzyme';
 
 import { Utilization } from '../Utilization';
 import { default as UtilizationFixtures } from '../fixtures/Utilization.fixture';
 
 describe('<Utilization />', () => {
   it('renders correctly', () => {
-    const component = render(<Utilization {...UtilizationFixtures.props} />);
+    const component = shallow(<Utilization {...UtilizationFixtures.props} />);
```
mareklibra: I am fine with this change but can you briefly explain reasons for it, please?

mareklibra:
@AyushAmbastha, @rawagner ,please excuse my late review of this.
I have a few concerns above, can you please respond? The best way is a follow now.

mareklibra:
If you need this in 2.0.0, a backport is required. Not sure if it is possible from process perspective now as there is just a Jira ticket for it and not BZ.

rawagner:
@mareklibra dashboard isn't in 2.0.0 so this is only for master :)

mareklibra:
Ooops, you are right :-)

AyushAmbastha:
> @AyushAmbastha, @rawagner ,please excuse my late review of this.
> I have a few concerns above, can you please respond? The best way is a follow now.

That's absolutely fine @mareklibra , I'll just look into the questions asked and get back to you shortly.

src/selectors/prometheus/selectors.js
AyushAmbastha: `parseNumber` will just return the value given to it in the `Number` form after checking if its NaN, so `parseNumber(timeValuePair[0] * 1000))`  and  `parseNumber(timeValuePair[0]) * 1000` will probably do the same thing and hence I didn't notice this. If you feel the latter is better to understand, we can change it.

src/components/StorageOverview/Utilization/Utilization.js
AyushAmbastha: The Axis defaults to the times `12am, 6am, 12pm` and so on when the value of x is `null`

src/components/StorageOverview/Utilization/tests/Utilization.test.js
AyushAmbastha: If we use `render`, it will render everything, including the child components resulting in a huge snapshot which will not be useful. Using `shallow` , only the single component is rendered and we get a smaller and more useful snapshots.

PR #438, opened 2019-05-06
Add a controller address validation

honza:
Validate the controller address in the create a new baremetal host dialog

### DIFF
--- a/src/components/CreateBaremetalHostDialog/CreateBaremetalHostDialog.js
+++ b/src/components/CreateBaremetalHostDialog/CreateBaremetalHostDialog.js
@@ -4,8 +4,10 @@ import { Modal, Col, Button, Alert } from 'patternfly-react';
 
 import { FormFactory } from '../Form/FormFactory';
 import { createBaremetalHost } from '../../k8s/requests';
+import { settingsValue } from '../../k8s/selectors';
 import { TinyInlineLoading } from '../Loading/Loading';
 import { CREATE_HOST_FORM_TEXT } from './strings';
+import { validateBmcURL } from '../../utils/validations';
 
 const formFields = {
   name: {
@@ -17,6 +19,7 @@ const formFields = {
     id: 'controller',
     title: 'Management Controller Address',
     required: true,
+    validate: form => validateBmcURL(settingsValue(form, 'controller')),
   },
   username: {
     id: 'username',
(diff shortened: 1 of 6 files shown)


src/components/CreateBaremetalHostDialog/CreateBaremetalHostDialog.js
```diff
@@ -17,6 +18,7 @@ const formFields = {
     id: 'controller',
     title: 'Management Controller Address',
     required: true,
+    validate: form => validateBmcURL(form.controller.value),
```
rawagner: you can use `settingsValue(form, 'controller')`

honza: That seems like an awfully specific (i.e. not general) function.

src/utils/validations.js
```diff
@@ -110,3 +112,38 @@ export const validateVmwareURL = value => {
...
+export const validateBmcURL = value => {
+  if (!value) {
+    return getValidationObject(EMPTY_ERROR);
+  }
+
+  const hasProtocol = value.includes('://');
+  const hasPort = value.match(':(?!/)');
+
+  if (hasProtocol) {
+    const startsWithAllowedProtocol = startsWith(value, 'ipmi') || startsWith(value, 'idrac');
```
suomiy: nitpick: could be also done this way
 
```js
if(hasProtocol && !['impi', 'idrac'].find(allowedProtocol => startsWith(value, allowedProtocol)){
  return getValidationObject(BMC_PROTOCOL_ERROR);
}
```

and `['impi', 'idrac']` could be  extracted to a constant.

src/components/CreateBaremetalHostDialog/CreateBaremetalHostDialog.js
rawagner: it is actually meant to be used with `FormFactory` so it applies here too. Problem is that we need to change function name and name of the parameters and move it to better place to make it seem more generic :) I will refactor it.

src/utils/validations.js
```diff
@@ -110,3 +112,38 @@ export const validateVmwareURL = value => {
...
+
+    if (!startsWithAllowedProtocol) {
+      return getValidationObject(BMC_PROTOCOL_ERROR);
+    }
+  }
+
+  if (hasPort) {
```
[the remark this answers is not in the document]
honza: Done

src/utils/validations.js
honza: Done

src/components/CreateBaremetalHostDialog/CreateBaremetalHostDialog.js
honza: Done, I'm happy to update it once it's refactored.

PR #440, opened 2019-05-07
Fix vddk mount path

mareklibra:
The source .tar already contains the "vmware-vix-disklib-distrib" directory.


mareklibra:
TODO: I will backport, please do not merge for visibility

suomiy:
@mareklibra I would advise not to change that because that would mean the pod would be mounted to the whole `/opt/` directory which I think is a bit too much

mareklibra:
Content of the source .tar contains the `vmware-vix-disklib-distrib` directory.
Without that, we end up with `/opt/vmware-vix-disklib-distrib/vmware-vix-disklib-distrib/*.*`

mareklibra:
We want to keep the source .tar untouched, but mounting to `/opt` limits us from extensions in the future.
So we will mount elsewhere and symlink under `/opt`

PR #441, opened 2019-05-07
Fix vddk mount path

mareklibra:
The source .tar already contains the "vmware-vix-disklib-distrib" directory.


mareklibra:
backport of: #441

PR #445, opened 2019-05-09
Display error on duplicate VM name in wizard

pcbailey:
This PR adds a validation check to ensure that a duplicate VM name is not provided.

Bug: https://bugzilla.redhat.com/1697333

Depends on: https://github.com/kubevirt/web-ui/pull/367

### DIFF
--- a/src/components/Wizard/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/Wizard/CreateVmWizard/CreateVmWizard.js
@@ -56,7 +56,7 @@ const getUpdatedValidatedState = (prevProps, prevState, props, state, extra) => 
 
   const validatedStepData = ALL_TAB_KEYS.reduce((resultStepData, tabKey) => {
     const { value, valid } = newState.stepData[tabKey];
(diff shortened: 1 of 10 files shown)


src/components/Wizard/CreateVmWizard/CreateVmWizard.js
```diff
@@ -56,7 +57,7 @@ const getUpdatedValidatedState = (prevProps, prevState, props, state, extra) =>
 
   const validatedStepData = ALL_TAB_KEYS.reduce((resultStepData, tabKey) => {
     const { value, valid } = newState.stepData[tabKey];
-    resultStepData[tabKey] = validateTabData(tabKey, value, valid);
+    resultStepData[tabKey] = validateTabData(tabKey, value, valid, { [VIRTUAL_MACHINES_KEY]: extra.virtualMachines });
```
suomiy: why use some extra for that, can't we rather just pass the props for better convenience and to be more generic?

pcbailey: Done!

PR #446, opened 2019-05-10
Created dropdown component and integrated it with top consumer card

gnehapk:
ref: https://github.com/kubevirt/web-ui/pull/370

After adding the dropdowns to top consumer card, it looks like this  -
top-consumer-with-drodowns

### DIFF
--- a/sass/components/Dashboard/dropdown.scss
+++ b/sass/components/Dashboard/dropdown.scss
@@ -1,0 +1,3 @@
+.kubevirt-dropdown__group {
+  margin-right: 0.4em;
+}
--- a/src/components/Dashboard/DashboardCard/DashboardCardActionsBody.js
+++ b/src/components/Dashboard/DashboardCard/DashboardCardActionsBody.js
@@ -1,0 +1,10 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const DashboardCardActionsBody = ({ children }) => (
+  <div className="kubevirt-dashboard__actions-body">{children}</div>
+);
+
+DashboardCardActionsBody.propTypes = {
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+};
--- a/src/components/StorageOverview/TopConsumers/TopConsumers.js
+++ b/src/components/StorageOverview/TopConsumers/TopConsumers.js
@@ -16,6 +16,8 @@ import {
 
 import { InlineLoading } from '../../Loading';
 import { formatToShortTime } from '../../../utils';
+import { Dropdown } from '../../Form/Dropdown';
+import { DashboardCardActionsBody } from '../../Dashboard/DashboardCard/DashboardCardActionsBody';
 
 import {
   DashboardCard,
@@ -23,15 +25,22 @@ import {
   DashboardCardHeader,
   DashboardCardTitle,
 } from '../../Dashboard/DashboardCard';
+
+import { PROJECTS, STORAGE_CLASSES, PODS, BY_USED_CAPACITY, BY_REQUESTED_CAPACITY } from './strings';
 import { StorageOverviewContext } from '../StorageOverviewContext/StorageOverviewContext';
 import { getTopConsumerVectorStats } from '../../../selectors/prometheus/storage';
 
-const TopConsumersBody = ({ topConsumers }) => {
+const metricTypes = [PROJECTS, STORAGE_CLASSES, PODS];
+const sortBy = [
+  { name: BY_USED_CAPACITY, refObj: 'byUsedCapacity' },
+  { name: BY_REQUESTED_CAPACITY, refObj: 'byRequestedCapacity' },
+];
+
+export const TopConsumersBody = ({ topConsumerStats, metricType, sortByOption }) => {
   let results = 'No data available';
-  const topConsumerStats = get(topConsumers, 'data.result', []);
 
   if (topConsumerStats.length) {
-    const stats = getTopConsumerVectorStats(topConsumerStats);
+    const stats = getTopConsumerVectorStats(topConsumerStats, metricType);
     const { chartData, legends, maxCapacity, unit } = stats;
     const yTickValues = [
       0,
@@ -67,7 +76,7 @@ const TopConsumersBody = ({ topConsumers }) => {
               <ChartGroup>{chartLineList}</ChartGroup>
               <ChartAxis tickFormat={x => formatToShortTime(x)} style={{ tickLabels: { fontSize: 8, padding: 5 } }} />
               <ChartAxis
-                label={`Requested capacity(${unit})`}
+                label={`${sortByOption.name}(${unit})`}
                 dependentAxis
                 tickValues={yTickValues}
                 style={{ tickLabels: { fontSize: 8, padding: 5 }, axisLabel: { fontSize: 8, padding: 25 } }}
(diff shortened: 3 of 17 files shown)


gnehapk:
Unit tests for Top consumer's Dropdown will be sent in next PR.

gnehapk:
@rawagner Please review.

src/components/Dashboard/DashboardCard/DashboardCardDropdownBody.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const DashboardCardDropdownBody = ({ children }) => <div className="kubevirt-dropdown__items">{children}</div>;
```
rawagner: I'd rather name this more generic, like `DashboardCardActionsBody`

src/components/StorageOverview/TopConsumers/TopConsumers.js
```diff
@@ -90,30 +98,129 @@ const TopConsumersBody = ({ topConsumerStats }) => {
...
-  <DashboardCard>
-    <DashboardCardHeader>
-      <DashboardCardTitle>Top Projects by Requested Capacity</DashboardCardTitle>
-    </DashboardCardHeader>
-    <DashboardCardBody isLoading={!topConsumerLoaded} LoadingComponent={LoadingComponent}>
-      <TopConsumersBody topConsumerStats={topConsumerStats} />
-    </DashboardCardBody>
-  </DashboardCard>
-);
+export class TopConsumers extends React.PureComponent {
```
rawagner: you need to pass specific props in https://github.com/kubevirt/web-ui-components/blob/c4331bf25e916fdb3dfd0dc671fc1f6eb19d8a78/src/components/StorageOverview/TopConsumers/TopConsumers.js#L228 not just `{...props}` otherwise `PureComponent` wont function as intended

sass/components/Dashboard/dropdown.scss
```diff
@@ -0,0 +1,11 @@
+.kubevirt-dropdown__items {
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+  margin: 0.6em 0;
+
+  .btn-group.btn-group-justified {
```
rawagner: it would be better to pass new argument to `Dropdown` (ie `groupClassName`) if you need to style it

src/components/Dashboard/DashboardCard/DashboardCardDropdownItem.js
```diff
@@ -0,0 +1,34 @@
...
+
+    const filterDropdown = {
+      id,
+      type: DROPDOWN,
+      choices: dropdownOptions,
+      onChange: newValue => onDropdownChange(dropdownType, newValue),
+      value: selectedOption,
+      className: 'kubevirt-dashboard-card-dropdown',
+    };
+    return getFormElement(filterDropdown);
```
suomiy: can you please create the dropdown element directly here instead of calling `getFormElement`. It is easier to read.

suomiy: Not sure if this component is needed at all. Can't we just use this inside TopConsumers? I see that the only difference is the classname. Also not sure if DashboardCardDropdownBody is necessary.

src/components/StorageOverview/TopConsumers/TopConsumers.js
```diff
@@ -90,30 +98,129 @@ const TopConsumersBody = ({ topConsumerStats }) => {
...
+
+      case STORAGE_CLASSES:
+        switch (this.state.sortBy) {
+          case BY_USED_CAPACITY:
+            return [];
+          case BY_REQUESTED_CAPACITY:
+            return this.getCapacity(this.props.slClassesRequestedCapacity);
+          default:
+            // eslint-disable-next-line no-console
+            console.log(`Unknown metric ${this.state.sortBy.name}`);
```
suomiy: isn't sortBy a string?

src/components/StorageOverview/TopConsumers/TopConsumers.js
```diff
@@ -90,30 +98,129 @@ const TopConsumersBody = ({ topConsumerStats }) => {
...
+
+      case PODS:
+        switch (this.state.sortBy) {
+          case BY_USED_CAPACITY:
+            return this.getCapacity(this.props.podsUsedCapacity);
+          case BY_REQUESTED_CAPACITY:
+            return [];
+          default:
+            // eslint-disable-next-line no-console
+            console.log(`Unknown metric ${this.state.sortBy.name}`);
```
suomiy: not sure if these console statements are really necessary? Can another value really occur here? Isn't this in our control?

src/components/StorageOverview/TopConsumers/TopConsumers.js
```diff
@@ -90,30 +98,129 @@ const TopConsumersBody = ({ topConsumerStats }) => {
...
+    return (
+      <DashboardCard>
+        <DashboardCardHeader>
+          <Row>
+            <Col lg={4} md={4} sm={4} xs={4}>
+              <DashboardCardTitle>Top Consumers</DashboardCardTitle>
+            </Col>
+            <Col lg={8} md={8} sm={8} xs={8}>
+              <DashboardCardDropdownBody>
+                <DashboardCardDropdownItem
```
suomiy: IMO this should be disabled if the data isLoading

src/components/Dashboard/DashboardCard/DashboardCardDropdownBody.js
gnehapk: ack

src/components/StorageOverview/TopConsumers/TopConsumers.js
gnehapk: ack

src/components/Dashboard/DashboardCard/DashboardCardDropdownItem.js
gnehapk: @suomiy There is a function 'getFormElement` inside FormFactory.js file which takes care of creating all the form components. So, I don't feel there is a need to rewrite code for creating dropdown component.

`DashboardCardDropdownItem` is a generic component which needs to be used across various cards in OCS dashboard for eg. Top Consumer, Performance, Capacity, Alerts. Hence, I have created a component. There as some cards where we need more than one dropdown(i.e. 2 or 3). `DashboardCardDropdownBody ` is parent component where you can pass as many dropdowns you required for the card(like we have done for Utilization card).

src/components/StorageOverview/TopConsumers/TopConsumers.js
gnehapk: yup, will correct it.

sass/components/Dashboard/dropdown.scss
gnehapk: ack

src/components/StorageOverview/TopConsumers/TopConsumers.js
gnehapk: Agreed. Can be removed.

src/components/StorageOverview/TopConsumers/TopConsumers.js
gnehapk: ack

src/components/Dashboard/DashboardCard/DashboardCardDropdownItem.js
suomiy: > There is a function 'getFormElement` inside FormFactory.js file which takes care of creating all the form components. So, I don't feel there is a need to rewrite code for creating dropdown component.

I think it would help the code readability because It just obfuscate things. It is not about rewriting the code for creating dropdown component. It is about using the dropdown component.

> DashboardCardDropdownItem is a generic component which needs to be used across various cards in OCS dashboard for eg. Top Consumer, Performance, Capacity, Alerts. Hence, I have created a component. There as some cards where we need more than one dropdown(i.e. 2 or 3). DashboardCardDropdownBody is parent component where you can pass as many dropdowns you required for the card(like we have done for Utilization card).

I don't think `width: 36% !important;` is specifically reusable. And I couldn't find the css for `kubevirt-dashboard-card-dropdown`. 

If it is made reusable as especially styled component, then please let these components in.

gnehapk: @suomiy ok. Do you mean to write the following code -
` 
<Dropdown
          id={id}
          value={value || defaultValue}
          onChange={onChange}
          choices={choices}
          disabled={disabled}
          className={className}
        />
`
with values populated in the component?
I feel, the intention of the `getFormElement` function is to return the form elements based on the type passed. @rawagner Please correct me, if I am wrong.  Just exploring the better approach to do it.

 Also,if you refer https://github.com/kubevirt/web-ui-components/blob/master/src/components/ClusterOverview/TopConsumers/TopConsumers.js#L123, the same way dropdown component has been created.

rawagner: @gnehapk @suomiy I dont mind using `getFormElement`. We should remove export from `getFormElement` if we dont want to use it outside `FormFactory`.

suomiy: > Do you mean to write the following code -
`<Dropdown id={id} value={value || defaultValue} onChange={onChange} choices={choices} disabled={disabled} className={className} />`

yes, with the props you need

> We should remove export from getFormElement if we dont want to use it outside FormFactory

> I feel, the intention of the getFormElement function is to return the form elements based on the type passed. @rawagner Please correct me, if I am wrong. Just exploring the better approach to do it.

this is mainly useful in `EditableDraggableTable` until that is refactored to pf4 - then we will see. So we need the export. 

I think using the underlying elements is preferable when dynamic approach is not required - i.e. this use case

> Also,if you refer https://github.com/kubevirt/web-ui-components/blob/master/src/components/ClusterOverview/TopConsumers/TopConsumers.js#L123, the same way dropdown component has been created.

I guess this would be also good to unify. Could you please include this in this PR?  I suspect TopConsumers should be similar for both dashboards, right?

gnehapk:
@rawagner I have incorporated the review comments. Please review.

src/components/Dashboard/DashboardCard/DashboardCardActionsBody.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const DashboardCardActionsBody = ({ children }) => <div className="kubevirt-dropdown__items">{children}</div>;
```
rawagner: className should be renamed to reflect the name of the component ie `kubevirt-dashboard-actions__body` and it also shouldn't be in `dropdown.scss` as it represents generic actions which have nothing to do with dropdown

sass/components/Dashboard/dropdown.scss
rawagner: This is still not addressed.  `btn-group.btn-group-justified` is specific to dropdowns but `DashboardCardActionsBody` may contain anything. When you need to style dropdowns in `DashboardCardActionsBody`  you need to style them directly which will require new Dropdown prop `groupClassName` which will be applied to `ButtonGroup` in `Dropdown.js`

gnehapk:
@rawagner Please review.

src/components/Dashboard/DashboardCard/DashboardCardDropdownItem.js
```diff
@@ -0,0 +1,36 @@
...
+
+export class DashboardCardDropdownItem extends React.PureComponent {
+  render() {
+    const { id, dropdownOptions, onDropdownChange, dropdownType, selectedOption, disabled } = this.props;
+
+    return (
+      <Dropdown
+        id={id}
+        value={selectedOption}
+        onChange={newValue => onDropdownChange(dropdownType, newValue)}
```
suomiy: this is unnecessary - this should be handled by the component which uses it. Btw is this component really necessary? - we could move the margin css to the parent component DashboardCardActionsBody which will apply it to its children.

sass/components/Dashboard/dropdown.scss
```diff
@@ -0,0 +1,3 @@
+.kubevirt-dropdown__group {
+  margin-right: 0.4em;
```
suomiy: shouldn't this be rather left? because we have the title first

src/components/StorageOverview/TopConsumers/TopConsumers.js
```diff
@@ -92,37 +98,160 @@ const TopConsumersBody = ({ topConsumers }) => {
...
+  };
+
+  getCapacity = data => get(data, 'data.result', []);
+
+  getCurrentMetric = () => {
+    switch (this.state.type) {
+      case PROJECTS:
+        switch (this.state.sortBy) {
+          case BY_USED_CAPACITY:
+            return this.getCapacity(this.props.projectsUsedCapacity);
```
suomiy: maybe object mapping (of sortBy to project prop name, etc.) would be easier to read than these many switches?

sass/components/Dashboard/dropdown.scss
suomiy: ow sorry, probably not - in that case we need to disable the margin for the first or last child (depends on left x right)

src/components/Dashboard/DashboardCard/DashboardCardDropdownItem.js
gnehapk: @suomiy this is being handled by the component only. We are passing the function definition which should be called when dropdown option changes and just mapping that function in the `Dropdown`

suomiy: You are passing a function and a value which you want the function to be called with.

That is passing a responsibility on something which doesn't have to have that responsibility. You might as well just pass a function/lambda with this binded value. 

This will be also easier to read from TopConsumers.

gnehapk: @suomiy Can you please explain with an example here

suomiy:
```jsx
      <DashboardCardDropdownItem
        dropdownOptions={sortBy}
        onDropdownChange={this.changeState}
        dropdownType="sortBy"
        selectedOption={this.state.sortBy}
      />
```

vs

```jsx
      <DashboardCardDropdownItem
        dropdownOptions={sortBy}
        onDropdownChange={value => this.changeState("sortBy", value)}
        selectedOption={this.state.sortBy}
      />
```

is more readable and there is less Magic involved

suomiy: I would rather discard that component altogether though as mentioned before

gnehapk: @suomiy Will take care of this function binding change in the follow up PR. There are other tasks, which is being blocked by this. Hence, need to merge this as soon as possible.

suomiy: these are pretty small changes, and IMO we could tackle them here 

\+ this is a refactoring so it makes sense to have them in this PR

src/components/StorageOverview/TopConsumers/TopConsumers.js
gnehapk: @suomiy will also take this up in the follow up PR.

gnehapk:
@rawagner Added all the tests and incorporated all the review comments. Please review.

PR #451, opened 2019-05-14
make details components safe for async requests

suomiy:
- forbid state updates on unmounted component

@rhrazdil this should prevent the known console.err of VmTemplateDetails


suomiy:
tests passed

PR #459, opened 2019-05-16
Added dropdown for alerts card in storage dashboard

afreen23:
Added dropddown which filters the alerts as per the severity: All, Critical, and Warning.

### DIFF
--- a/src/components/StorageOverview/Alerts/Alerts.js
+++ b/src/components/StorageOverview/Alerts/Alerts.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Row, Col } from 'patternfly-react';
 
 import {
   DashboardCard,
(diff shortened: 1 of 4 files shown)


gnehapk:
@rawagner Please review.

src/components/StorageOverview/Alerts/Alerts.js
```diff
@@ -1,39 +1,94 @@
...
-            <AlertItem key={`alert-${index}`} alert={alert} />
-          ))}
-        </AlertsBody>
-      </DashboardCardBody>
-    </DashboardCard>
-  ) : null;
-};
+  onDropdownOptionChange = newVal => this.setState({ type: newVal }, this.getAlerts);
+
+  filterAlertsBy = (allAlerts, filter) => {
```
suomiy: could you please rename it  at least to filterBy ?

src/components/StorageOverview/Alerts/Alerts.js
```diff
@@ -1,39 +1,94 @@
...
+const alertsFilterOptions = [ALL, CRITICAL, WARNING];
 
-export const Alerts = ({ alertsResponse, className }) => {
-  if (!Array.isArray(alertsResponse)) {
-    return null;
+export class Alerts extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {
+      type: alertsFilterOptions[0],
```
suomiy: why not just put `ALL` here?

src/components/StorageOverview/Alerts/Alerts.js
```diff
@@ -1,39 +1,94 @@
...
-        <AlertsBody>
-          {alerts.map((alert, index) => (
-            <AlertItem key={`alert-${index}`} alert={alert} />
-          ))}
-        </AlertsBody>
-      </DashboardCardBody>
-    </DashboardCard>
-  ) : null;
-};
+  onDropdownOptionChange = newVal => this.setState({ type: newVal }, this.getAlerts);
```
suomiy: what is the `this.getAlerts` callback for?

src/components/StorageOverview/Alerts/Alerts.js
```diff
@@ -1,39 +1,94 @@
...
+
+  filterAlertsBy = (allAlerts, filter) => {
+    const filteredAlerts = allAlerts.filter(alert => getAlertSeverity(alert) === filter);
+    /* handles the case when no alerts of a particular filter are present */
+    return filteredAlerts.length !== 0 ? filteredAlerts : allAlerts;
```
suomiy: IMO we should show zero alerts if the user wishes the filter to be that way.

src/components/StorageOverview/Alerts/strings.js
```diff
@@ -0,0 +1,3 @@
+export const ALL = 'All';
```
suomiy: this should probably go under `components/Dashboard`

src/components/StorageOverview/Alerts/tests/__snapshots__/Alerts.test.js.snap
```diff
@@ -142,11 +259,90 @@ exports[`<Alerts /> renders correctly with Provider 1`] = `
   <div
```
matthewcarleton: You could simplify this by using flex (which I think is already applied to `.kubevirt-dashboard__card-title`) and remove all of the bootstrap grid markup.

src/components/StorageOverview/Alerts/Alerts.js
afreen23: yeah even I tried that but the issue is when no alerts i.e [] array is passed then the entire alert component is gone since inside render we have this condition `return alerts.length > 0` and `null` is returned.
So if user wants to select any other filter, he wont be able to do that.

suomiy: We could change that condition so that alerts card is visible when some alert exists, i.e. `allAlerts.length > 0`

src/components/StorageOverview/Alerts/Alerts.js
afreen23: just thought its better to put the reference to the first element instead of hardcoding so that in future if any changes would occur it wont affect that...

src/components/StorageOverview/Alerts/Alerts.js
afreen23: ah! no need for this...will remove it

src/components/StorageOverview/Alerts/strings.js
afreen23: right now `strings.js` file for each component is inside that component so followed the same pattern...if u want i will move it !

src/components/StorageOverview/Alerts/Alerts.js
afreen23: ack

src/components/StorageOverview/Alerts/tests/__snapshots__/Alerts.test.js.snap
afreen23: @matthewcarleton  can you please specify what to simplify..sorry did not get you

matthewcarleton: Sure! 
What happens when you remove the boostrap grid markup? Does it not get the right layout? 
`.kubevirt-dashboard__card-title` has the following css:
```
display: flex;
justify-content: space-between;
align-items: center;
```
so if you have the following markup:
```
<div class="card-pf-heading kubevirt-dashboard__card-heading">
  <h2 class="card-pf-title kubevirt-dashboard__card-title">
     Alerts
  </h2>
  <!--dropdown markup here :)-->
</div>
```
I would expect to see the layout you need.

afreen23: okay..so when I remove bootstrap grid I ma getting like this..which is not desired since dropdown is occupying all remaining space
alerts

matthewcarleton: ah ya
I don't think you'll need 
```<div class="btn-group btn-group-justified">```
wrapping it either.

afreen23: @matthewcarleton this is the snapshot file generated by snaphot tests. I cannot change it.

matthewcarleton: ah ok my apologies for making this more confusing. I'm not very familiar with how this is built I am only judging the output. This may be out of scope for this issue, I do think it's worth revisiting though. We are using PF4 to layout these pages but then using PF3 bootstrap grid inside the cards. We can simplify it all quite a bit, I'm happy to help!

src/components/StorageOverview/Alerts/Alerts.js
afreen23: @suomiy is it looks good now

src/components/StorageOverview/Alerts/Alerts.js
suomiy: I see it the other way around. It will be broken if someone changes that array in the future (adding element or changing the order of them). + it will be pretty explicit if we put `ALL` there and we do not need to jump to array to find out about that value.

src/components/StorageOverview/Alerts/Alerts.js
suomiy: I mean the variable :), because it could be mistaken for the filter function

src/components/StorageOverview/Alerts/strings.js
suomiy: these constants will be reused by other dashboards, so it would be good if you could move it

gnehapk:
@rawagner Please review.

PR #460, opened 2019-05-16
update-capacity-card-css

matthewcarleton:
fixes issue where text doesn't break gracefully. Also visual tweaks (margin, line-height).

### DIFF
--- a/sass/components/Dashboard/capacity.scss
+++ b/sass/components/Dashboard/capacity.scss
@@ -18,6 +18,10 @@
 
 .kubevirt-capacity__item-description {
   font-size: 0.6875em;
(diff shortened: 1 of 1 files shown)


suomiy:
@matthewcarleton I just noticed that when I create a vm with a long name (e.g. 100+ characters (max is 253)), it overlaps with other fields in vm details page. I guess this could be the case with other fields as well. Could you please look into that?

matthewcarleton:
@suomiy I checked it out and I don't see that. Here's what I see. Maybe I misunderstood you. What browser are you testing with?
Screen Shot 2019-05-17 at 9 11 42 AM

PR #463, opened 2019-05-17
Fix name line break

matthewcarleton:


sass/components/VmDetails/vm-details.scss
```diff
@@ -39,6 +39,11 @@
   width: 400px;
   flex: 1;
   flex-basis: auto;
+  margin-right: 8px;
```
suomiy: hmm, I guess we could also add this to `kubevirt-vm-details__details-list` just to be sure?

suomiy: meaning the word-break, not sure about the margins..

sass/components/VmDetails/vm-details.scss
```diff
@@ -39,6 +39,11 @@
...
+  
+  & > * {
```
suomiy: plus we also need this here

```
/* stylelint-disable plugin/selector-bem-pattern */

/* stylelint-enable */
```

because the tests are crashing

sass/components/VmDetails/vm-details.scss
matthewcarleton: There is already margin-right:20px; there which I think is ok. I just added a minimal amount just in case we have this long string for the name.

suomiy: Ok. Can we have a  `word-break` there?

matthewcarleton: oh sorry didn't see your second comment :)

PR #471, opened 2019-05-20
Do not close Create VM Wizard when clicked outside of modal

rawagner:
https://bugzilla.redhat.com/show_bug.cgi?id=1710888

### DIFF
--- a/src/components/Wizard/CreateVmWizard/CreateVmWizard.js
+++ b/src/components/Wizard/CreateVmWizard/CreateVmWizard.js
@@ -289,6 +289,7 @@ export class CreateVmWizard extends React.Component {
     return (
       <Wizard.Pattern
         show
+        backdrop="static"
         onHide={this.onHideWrapper}
         steps={this.wizardStepsNewVM}
         activeStepIndex={this.state.activeStepIndex}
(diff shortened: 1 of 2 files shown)


suomiy:
IMO this is a feature, not a bug :) Users should know what to expect from `escape` key.

suomiy:
not sure about the backdrop click

mareklibra:
I like this change.
Former behaviour was annoying to me.

andybraren:
No other modal (that I can find) closes itself when clicking on the backdrop, so I agree that annoying behavior should be changed. 🙂

I'm not sure about disabling the `Esc` key though. The new [PF4 wizard](https://www.patternfly.org/v4/documentation/react/components/wizard/) supports using `Esc` to close for accessibility reasons I believe, and existing modals within the Console can be `Esc`'d to close too. Is hitting `Esc` and accidentally closing the modal a frequent problem?

matthewcarleton:
+1 to @andybraren comment about the "esc" key. For more information: https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html and here's a good write up on it too https://simplyaccessible.com/article/closing-modals/

rawagner:
@andybraren @matthewcarleton thank you for the input! I changed it to not close when user clicks outside of modal (which was pretty annoying), `esc` is unchanged.

PR #473, opened 2019-05-22
Fix baremetal status popover

yaacov:
Currently BareMetalHosts Error status is broken.

-  Fix use of PF4 popover.
-  Syncing with `openshift/console` to patternfly-4 core version `3.16` from `1.43.x`.

Notes:
-  patternfly-4 v `1.x` has a broken `Popover`.
-  web-ui version of pf4 react is currently 2.4.1:
   https://github.com/kubevirt/web-ui/blob/master/frontend/package.json#L63
- openshift console use 3.16.16:
   https://github.com/openshift/console/blob/master/frontend/package.json#L66

Screenshot:
Screenshot from 2019-05-22 13-30-17

### DIFF
--- a/config/cosmos.webpack.config.js
+++ b/config/cosmos.webpack.config.js
@@ -43,7 +43,8 @@ module.exports = {
       {
         test: /\.(jpg|jpeg|png|gif|svg|eot|otf|ttf|woff|woff2)$/,
         loader: 'file-loader',
-      },{
+      },
+      {
         test: /\.css$/,
         use: ['style-loader', 'css-loader'],
       },
--- a/config/cosmos.webpack.config.js
+++ b/config/cosmos.webpack.config.js
@@ -44,6 +44,16 @@ module.exports = {
         test: /\.(jpg|jpeg|png|gif|svg|eot|otf|ttf|woff|woff2)$/,
         loader: 'file-loader',
       },
+      {
+        test: /\.css$/,
+        exclude: /node_modules/,
+        use: ['style-loader', 'css-loader'],
+      },
+      {
+        test: /\.css$/,
+        include: /node_modules/,
+        use: ['style-loader', 'css-loader'],
+      },
     ],
   },
 };
--- a/config/cosmos.webpack.config.js
+++ b/config/cosmos.webpack.config.js
@@ -43,15 +43,8 @@ module.exports = {
       {
         test: /\.(jpg|jpeg|png|gif|svg|eot|otf|ttf|woff|woff2)$/,
         loader: 'file-loader',
-      },
-      {
+      },{
         test: /\.css$/,
-        exclude: /node_modules/,
-        use: ['style-loader', 'css-loader'],
-      },
-      {
-        test: /\.css$/,
-        include: /node_modules/,
         use: ['style-loader', 'css-loader'],
       },
     ],
--- a/sass/components/Status/status.scss
+++ b/sass/components/Status/status.scss
@@ -7,6 +7,6 @@
   margin: 0;
 }
 
-.kubevirt-status__span {
+.kubevirt-status__popover {
   cursor: pointer;
 }
--- a/sass/components/Status/status.scss
+++ b/sass/components/Status/status.scss
@@ -6,3 +6,7 @@
   padding: 0;
   margin: 0;
 }
+
+.kubevirt-status__span {
+  cursor: pointer;
+}
--- a/src/components/BareMetalHosts/StatusComponents.js
+++ b/src/components/BareMetalHosts/StatusComponents.js
@@ -28,7 +28,7 @@ GenericSuccess.defaultProps = GenericStatus.defaultProps;
 export const GenericError = ({ status, text, errorMessage }) =>
   errorMessage ? (
     <OverlayStatus icon="error-circle-o" header={<div>{text}</div>}>
-      {<div>{errorMessage}</div>}
+      <div>{errorMessage}</div>
     </OverlayStatus>
   ) : (
     <Status icon="error-circle-o">{text}</Status>
@@ -47,7 +47,7 @@ GenericProgress.defaultProps = GenericStatus.defaultProps;
 // (TODO) Add details for validation errors in Popover component
 export const ValidationError = ({ status, text, errorMessage }) => (
   <OverlayStatus icon="error-circle-o" header={<div>{text}</div>}>
-    {<div>{errorMessage}</div>}
+    <div>{errorMessage}</div>
   </OverlayStatus>
 );
 
--- a/src/components/BareMetalHosts/StatusComponents.js
+++ b/src/components/BareMetalHosts/StatusComponents.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Popover } from '@patternfly/react-core';
 
 import { Status, OverlayStatus } from '../Status';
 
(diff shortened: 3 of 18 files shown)


yaacov:
@suomiy @rawagner please review.

yaacov:
@dgutride Hi,

We use Jest with this config:
https://github.com/kubevirt/web-ui-components/blob/master/config/jest.config.js

When upgrading to `@patternfly/react-core": "3.x"` and running jest I get:
```
This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /home/yzamir/go/src/github.com/kubevirt/web-ui-components/node_modules/@patternfly/react-core/dist/js/@patternfly/patternfly/components/AboutModalBox/about-modal-box.css:2
    .pf-c-about-modal-box {
    ^
```

Do you know what am I missing ?

dgutride:
@yaacov  - you need to stub out the css includes - see our jest config in the sample app here:
https://github.com/patternfly/patternfly-react-seed/blob/master/jest.config.js#L29

yaacov:
@dgutride THANKS, it works now :heart:

yaacov:
WIP because grid in v3.x has changed since v1.x and the storage overview is now broken :-(

yaacov:
sorry, removing WIP, storage overview looks the same, it's not because of the new PF version ...

yaacov:
@suomiy @rawagner please review.

-  Syncing pf-4 react-core version to 2.4.1 like in web-ui
-  Leaving the config code to work with `pf@react-core` version `3.x` , because why not :-)

yaacov:
Ahh cosole using `3.16.16` :tada:
https://github.com/openshift/console/blob/master/frontend/package.json#L66

bumping back up to 3 ... :rocket:

src/components/BareMetalHosts/StatusComponents.js
```diff
@@ -23,26 +22,14 @@ export const GenericSuccess = ({ status, text, errorMessage }) => <Status icon="
...
-};
-
 // Generic error status component
 // If the errorMessage property isn't empty its contents are
 // shown in a Popover.
 export const GenericError = ({ status, text, errorMessage }) =>
   errorMessage ? (
-    <OverlayStatus icon="error-circle-o" text={text} overlay={<ErrorOverlay errorMessage={errorMessage} />} />
+    <OverlayStatus icon="error-circle-o" header={<div>{text}</div>}>
+      {<div>{errorMessage}</div>}
```
suomiy: `{` -||- `}` are superfluous here

suomiy:
it seems it is not possible to select the popover text (header, body) and it is instantly closed instead (tried only in cosmos)

src/components/BareMetalHosts/StatusComponents.js
yaacov: Thanks :blush:

yaacov:
> it seems it is not possible to select the popover text (header, body) and it is instantly closed instead (tried only in cosmos)

@suomiy , that is wired :sunflower: , downgraded pf@react-core to 3.16.10 ( current stable ) and it fixed it ...

Peek 2019-05-22 20-38

yaacov:
Squashed commits :sweet_potato:

yaacov:
When I chenaged the status overlay to use PF4 popover I broke nodestatus :-(
Added a fix for node status, it now looks like this:
Peek 2019-05-23 12-48

src/components/BareMetalHosts/StatusComponents.js
rawagner:
```suggestion
      <div>{errorMessage}</div>
```

src/components/BareMetalHosts/StatusComponents.js
```diff
@@ -59,7 +46,9 @@ GenericProgress.defaultProps = GenericStatus.defaultProps;
 // Validation Error component
 // (TODO) Add details for validation errors in Popover component
 export const ValidationError = ({ status, text, errorMessage }) => (
-  <OverlayStatus icon="error-circle-o" text={text} overlay={<ErrorOverlay errorMessage={errorMessage} />} />
+  <OverlayStatus icon="error-circle-o" header={<div>{text}</div>}>
+    {<div>{errorMessage}</div>}
```
rawagner:
```suggestion
    <div>{errorMessage}</div>
```

src/components/NodeStatus/NodeStatus.js
```diff
@@ -25,10 +24,14 @@ const UnderMaintenanceStatus = ({ node, maintenance, TimestampComponent }) => {
...
         Started: <TimestampComponent simple timestamp={created} />
       </div>
-    </Popover>
+    </React.Fragment>
   );
 
-  return <OverlayStatus icon="off" overlay={overlay} text="Under maintenance" />;
+  return (
+    <OverlayStatus icon="off" header="Under maintenance">
+      {overlay}
```
rawagner: i'd just move the whole overlay content here

src/components/NodeStatus/NodeStatus.js
```diff
@@ -55,10 +58,14 @@ const StoppingMaintenanceStatus = ({ node, maintenance, TimestampComponent }) =>
...
         Ended: <TimestampComponent simple timestamp={deleted} />
       </div>
-    </Popover>
+    </React.Fragment>
   );
 
-  return <OverlayStatus icon="off" overlay={overlay} text="Stopping maintenance" />;
+  return (
+    <OverlayStatus icon="off" header="Stopping maintenance">
+      {overlay}
```
rawagner: i'd just move the whole overlay content here

src/components/Status/Status.js
```diff
@@ -19,14 +20,16 @@ Status.propTypes = {
...
 
-export const OverlayStatus = ({ icon, text, overlay }) => (
-  <Status icon={icon}>
-    <OverlayTrigger overlay={overlay} placement="right" trigger={['click']} rootClose>
-      <Button className="kubevirt-status__button" bsStyle="link">
-        {text}
-      </Button>
-    </OverlayTrigger>
-  </Status>
+export const OverlayStatus = ({ icon, header, children }) => (
```
rawagner: we could rename this to Popover status now

src/components/Status/Status.js
```diff
@@ -19,14 +20,16 @@ Status.propTypes = {
...
+  <Popover position="right" headerContent={header} bodyContent={children}>
+    <span>
```
rawagner: `span` is unnecessary here, isn't it ?

config/cosmos.webpack.config.js
```diff
@@ -44,6 +44,16 @@ module.exports = {
         test: /\.(jpg|jpeg|png|gif|svg|eot|otf|ttf|woff|woff2)$/,
         loader: 'file-loader',
       },
+      {
+        test: /\.css$/,
+        exclude: /node_modules/,
+        use: ['style-loader', 'css-loader'],
+      },
+      {
+        test: /\.css$/,
```
rawagner: so we exclude `node_modules` in the first one and then we include them in the second ? I dont understand

src/components/NodeStatus/NodeStatus.js
yaacov: Nice :+1:

config/cosmos.webpack.config.js
yaacov: :smile:  copy paste from a tutorial (they wanted to convey a point) ... , yes it's the same as doing just one:
```
{
  test: /\.css$/,
  use: ['style-loader', 'css-loader'],
},
```

src/components/Status/Status.js
yaacov: Make sense :-)

src/components/Status/Status.js
yaacov: When I remove the `<span>` the trigger component is only the `icon`, and pressing the `button` element does not trigger the popover :-(

This may be a bug in the `popover` element, but "adding span" fix/hack is easy ... and solves the problem.

@rawagner WDYT, can we do it a nicer way, can we ignore this odd behaviur from popover, and assume it will be fixed upstream ?

Screenshot of problem:
Peek 2019-05-23 14-27

src/components/BareMetalHosts/StatusComponents.js
yaacov: I already fixed it when @suomiy commented about that :-( ... probably forgot to commit, now it's fixed and commited :-)

yaacov:
@suomiy @rawagner please review.

-  Re fixed the bracket thing :-)
-  Simplied the cosmos config, and removed the redundant overlay const.

Not sure what to do with the extra `<span>` used to force trigger area for status popover, removing it looks better, but breaks the component behaviur, help ?

src/components/Status/Status.js
rawagner: ah, ok. Popover expects one child. Can you add style to the span element to show pointer cursor ?

yaacov: wow, didn't notice the pointer shape, thanks :cake:

sass/components/Status/status.scss
```diff
@@ -6,3 +6,7 @@
   padding: 0;
   margin: 0;
 }
+
+.kubevirt-status__span {
```
rawagner: can you rename it to `kubevirt-status__popover` ? Otherwise LGTM, I will push it after the change :)

yaacov: :beers:

PR #475, opened 2019-05-23
Capacity Card Dropdown

AyushAmbastha:
Implemented the dropdown options for the Capacity card.
1. Total Capacity
2. Requested vs Used
3. VMs vs Pods

Screenshot from 2019-06-04 16-31-31

The queries haven't been tested yet and will be changed, if needed.

### DIFF
--- a/src/components/StorageOverview/Capacity/Capacity.js
+++ b/src/components/StorageOverview/Capacity/Capacity.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Row, Col } from 'patternfly-react';
 
 import { InlineLoading } from '../../Loading';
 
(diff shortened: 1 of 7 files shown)


src/components/StorageOverview/Capacity/Capacity.js
```diff
@@ -9,48 +11,131 @@ import {
...
 
 export class Capacity extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {
+      capacity: capacityTypes[0],
+    };
+  }
+
+  convertMetrics = (Total, Used, title) => {
```
rawagner: so this function converts {Total, Used, title} to {metric1, metric2, title} and then it is passed back to Total/Used/title in https://github.com/kubevirt/web-ui-components/pull/475/files#diff-0be8d3e2808ab296537ca229a7f5a817R104 . Why use it at all ?

src/components/StorageOverview/Capacity/Capacity.js
```diff
@@ -9,48 +11,131 @@ import {
...
 import { StorageOverviewContext } from '../StorageOverviewContext';
 import { CapacityItem } from '../../Dashboard/Capacity/CapacityItem';
 import { formatBytes } from '../../../utils';
 import { getCapacityStats } from '../../../selectors';
 import { CapacityBody } from '../../Dashboard/Capacity/CapacityBody';
+import { TOTAL, CAPACITY, VMSVSPODS } from './strings';
+
+const capacityTypes = [TOTAL, CAPACITY, VMSVSPODS];
+
+export const CapacityImplementation = ({ Total, Used, title, LoadingComponent }) => (
```
rawagner: why having a separate component ? Can't you just have the content of `CapacityImplementation` directly in `DashboardCardBody` ?

src/components/StorageOverview/Capacity/Capacity.js
```diff
@@ -9,48 +11,131 @@ import {
...
-              title="Total capacity"
-              used={getCapacityStats(capacityUsed)}
-              total={getCapacityStats(capacityTotal)}
-              formatValue={formatBytes}
-              LoadingComponent={LoadingComponent}
-              isLoading={!(capacityUsed && capacityTotal)}
-            />
-          </CapacityBody>
+          <CapacityImplementation
+            Total={capacitymetric.metric1}
```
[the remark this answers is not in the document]
AyushAmbastha: Will change it. Ack!

src/components/StorageOverview/Capacity/Capacity.js
AyushAmbastha: I used this function to convert the different metrics used for `total` and `used` to a common variable so that it can be passed to the function to implement the Donut chart easily.
Example - For `Requested vs Used` , values for total, used and title are different when compared to values for `VMs vs Pods` .  https://github.com/kubevirt/web-ui-components/pull/475/files#diff-0be8d3e2808ab296537ca229a7f5a817R67
The variable names are repeated as I couldn't think of any other suitable name.

src/components/StorageOverview/Capacity/Capacity.js
AyushAmbastha: I can do that, was trying to keep it consistent with the way the other dropdowns are implemented. As metrics are sent to a `...Body` component which renders the card.
Couldn't reuse `CapacityBody` so renamed it to `CapacityImplementation`.

src/components/StorageOverview/Capacity/Capacity.js
rawagner: ah, ok. But it could be slimmed down to 
```suggestion
  convertMetrics = (total, used, title) => ({total, used, title});
```

src/components/StorageOverview/Capacity/Capacity.js
rawagner: I still think it would be cleaner to have content directly in `DashboardCardBody` - you are not doing any other calculations but others (ie TopConsumers) do. So it's just unnecessary wrapping.

src/components/StorageOverview/Capacity/Capacity.js
```diff
@@ -1,38 +1,89 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-
-import { InlineLoading } from '../../Loading';
+import { Row } from 'patternfly-react/dist/js/components/Grid';
```
rawagner:
```suggestion
import { Row, Col } from 'patternfly-react';
```

AyushAmbastha:
@rawagner Done! :)

src/components/StorageOverview/Capacity/Capacity.js
```diff
@@ -1,38 +1,88 @@
...
             <CapacityItem
               id="capacity"
-              title="Total capacity"
-              used={getCapacityStats(capacityUsed)}
-              total={getCapacityStats(capacityTotal)}
+              title={capacitymetric.title}
+              used={getCapacityStats(capacitymetric.used)}
+              total={getCapacityStats(capacitymetric.total)}
               formatValue={formatBytes}
-              LoadingComponent={LoadingComponent}
```
rawagner: why did you remove passing LoadingComponent ? We need it to pass `LoadingInline` from `web-ui`

PR #476, opened 2019-05-24
Do not allow floats as memory value

rawagner:
https://bugzilla.redhat.com/show_bug.cgi?id=1710775

we should allow floats and different values than Gi for web-ui future versions


rawagner:
created issue for future releases https://github.com/kubevirt/web-ui-components/issues/479

PR #481, opened 2019-05-27
added dropdown for VMs in top consumers card

afreen23:
vmx
vms


rawagner:
lgtm. Lets wait for web-ui counterpart PR

PR #474, opened 2019-05-22
move CreateVmWizard state to redux

suomiy:
- refactoring
- bug fixes

sibling  PR: https://github.com/kubevirt/web-ui/pull/393

PR #483, opened 2019-05-29
 CreateVmDialog: V2V VMware report deployment status and all errors

suomiy:
depends on https://github.com/kubevirt/web-ui-components/pull/474

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1709775


mareklibra:
@suomiy , is targeted for v2.0.0, GA is very close.
This PR is based on #474 which will not make it to v2.0.0.

Let's discuss offline.

Cc @jelkosz

suomiy:
@mareklibra this was planned to be backported - but will be discussed further as the scope is a bit bigger

suomiy:
Some screenshots:

Deployment Failure:

deployment-fail

Deploying (no deployment exists):

deployment-starting


Deploying

deployment-deploying

Deployment Failed

deploy-fail2

This error will appear after some time (10mins) after the deployment reaches progress deadline and it will not try to restart the pod again.

deploy-fail3

matthewcarleton:
@suomiy can you explain what's happening with the error firing? Is the "deployment Error" text inside the accordion? Could we just move that outside the accordion and use default error text on the input? 
Can you provide screen shots of what content is inside of the accordion sections?

suomiy:
> @suomiy can you explain what's happening with the error firing? Is the "deployment Error" text inside the accordion? Could we just move that outside the accordion and use default error text on the input?
Can you provide screen shots of what content is inside of the accordion sections?



It is not just one error but a list of objects which were either
- not created and failed
- created & cleaned up
- created and failed to cleanup 

these objects are all related to the deployment and might give clues to what is wrong

the content with an example error (will be probably something different though)

one

two

matthewcarleton:
ah ok. What do you think of breaking the error text outside of the accordion? I think that deserves to be high level and just standard input error text so we don't see the overflow issue.

suomiy:
> What do you think of breaking the error text outside of the accordion?

I guess that could look a bit better

Should we do the same for the results page? Because it uses the same design and approach.
IMO it would be better to do it in a separate patch

matthewcarleton:
yes I'd agree. Although because this wizard will be updated to PF4 soon(?) maybe we can wait until then to put in that effort. WDYT?

suomiy:
ok, sounds good

PR #484, opened 2019-05-29
fix: reset OS when re-selecting VM for import

suomiy:
- Once operating system is picked, Operating system for another VM is now detected

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1710932

### DIFF
--- a/src/components/Wizard/CreateVmWizard/stateUpdate/providers/prefillVmStateUpdate.js
+++ b/src/components/Wizard/CreateVmWizard/stateUpdate/providers/prefillVmStateUpdate.js
@@ -47,20 +47,20 @@ export const prefilUpdateCreator = (prevProps, prevState, props, state) => {
       [VM_SETTINGS_TAB_KEY]: {
         value: {
           [NAME_KEY]: {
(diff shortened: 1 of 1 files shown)


suomiy:
separated from https://github.com/kubevirt/web-ui-components/pull/474 for faster bug fix

mareklibra:
lgtm

I have changed subject of this PR to be descriptive for humans and not-containing an URL.
Reference to BZ is added to the description field, annotated with `Fix:`

mareklibra:
lgtm

I have changed subject of this PR to be descriptive for humans and not-containing an URL.
Reference to BZ is added to the description field, annotated with `Fix:`

mareklibra:
@suomiy , can you please update commit message in a following way:
- first line contains short description of content, ideally <80 characters
- if needed (depends on the first line ^^^), add more details on other lines
- on a separate line, please reference the bug

We do not have code-style formally defained in this project but I believe this should be in.

suomiy:
this was a small quick fix, but you are right about the format

- fixed the commit message

PR #485, opened 2019-05-30
CreateVmWizard WMware - fix: fields become inactive after the VM is picked

suomiy:
- BZ https://bugzilla.redhat.com/show_bug.cgi?id=1710944


suomiy:
@mareklibra #487

PR #493, opened 2019-06-07
Fix save vCenter credentials functionality

pcbailey:
This PR fixes an issue where vCenter credentials are not saved unless the create VM wizard is completed, regardless of success or failure. These changes cause the kubevirt.io/temporary label to be removed from the secret if the "Remember vCenter credentials" box is checked and adds the label with a value of true if the "Remember vCenter credentials" box is unchecked; however, these changes are only made if the connection check is successful.

Bug: https://bugzilla.redhat.com/1714004

Depends on: kubevirt/web-ui#399

Original PR: https://github.com/kubevirt/web-ui-components/pull/489


mareklibra:
This backport comes without snapshots from #489.
As activity on the stable branch is getting lower and lower and bug-fix release is requested to be asap, let's accept that.

PR #503, opened 2019-06-26
CreateVmDialog uses Firehose

mareklibra:
Depends on: #502

### DIFF
--- a/src/components/Wizard/CreateVmWizard/providers/VMwareImportProvider/VMWareImportProvider.js
+++ b/src/components/Wizard/CreateVmWizard/providers/VMwareImportProvider/VMWareImportProvider.js
@@ -45,6 +45,7 @@ import {
 } from '../../../../../utils/status/v2vVMware';
 import { getVmwareField } from './selectors';
 import { correctVCenterSecretLabels } from '../../../../../k8s/requests/v2v/correctVCenterSecretLabels';
+import { flatten } from '../../../../../k8s/util/k8sMethodsUtils';
 
 export class VMWareImportProvider extends React.Component {
   state = {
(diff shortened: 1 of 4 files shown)


mareklibra:
No matter this is intended for openshift/console only, there will be follow-up updating web-ui to keep its master working.

src/components/Wizard/CreateVmWizard/providers/VMwareImportProvider/requestedResources.js
```diff
@@ -15,39 +15,39 @@ export const getVmWareProviderRequestedResources = state => {
...
         namespace,
         isList: false,
-      }),
-    };
+        prop: 'v2vvmware',
+      })
+    );
   }
 
   if (activeVcenterSecretName) {
```
suomiy: this should be converted as well

src/components/Wizard/CreateVmWizard/providers/VMwareImportProvider/VMWareImportProvider.js
```diff
@@ -161,9 +166,19 @@ export class VMWareImportProvider extends React.Component {
   }
```
suomiy: and flatten activeVcenterSecret

src/components/Wizard/CreateVmWizard/providers/VMwareImportProvider/VMWareImportProvider.js
```diff
@@ -107,7 +111,8 @@ export class VMWareImportProvider extends React.Component {
   });
 
   componentDidUpdate() {
-    const { vmwareToKubevirtOsConfigMap, v2vvmware } = this.props;
+    const v2vvmware = flatten(this.props, 'v2vvmware', {});
+    const vmwareToKubevirtOsConfigMap = flatten(this.props, 'vmwareToKubevirtOsConfigMap', undefined);
```
suomiy: do we really have to specify undefined value here?

mareklibra: no, leftover

src/components/Wizard/CreateVmWizard/providers/VMwareImportProvider/requestedResources.js
mareklibra: good catch, change introduced during rebase

PR #505, opened 2019-06-27
flatten activeVcenterSecret

mareklibra:


suomiy:
the snapshots need to be updated

PR #507, opened 2019-07-11
Configure V2V via ConfigMap

mareklibra:
This patch removes dependency on v2v-specific SERVER_FLAGS with the motivation
to call this code from upstream openshift/console without the need to
modify the bridge-process there.

The V2V deployment is newly configured via "v2v-vmware" ConfigMap within
the "openshift" namespace.

Example:
```
oc apply -f - <<EOF
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: openshift
  name: v2v-vmware
data:
  v2v-conversion-image: [V2V-CONVERSION-IMAGE]
  kubevirt-vmware-image: [KUBEVIRT-VMWARE-IMAGE]
  kubevirt-vmware-image-pull-policy: IfNotPresent
EOF
```

It is farther expected to have
- v2vvmware CRD deployed [1]
- the vddk-pvc created as before

The kubevirt-vmware deployment logic is recently expected to be separated
to an operator (does not exist yet).

[1] oc apply -f https://raw.githubusercontent.com/kubevirt/web-ui-operator/master/build/kubevirt-web-ui-ansible/roles/kubevirt_web_ui/files/kubevirt_v1alpha1_v2vvmware_crd.yaml


src/config/index.js
```diff
@@ -1,12 +1,24 @@
...
-const getRegistryUrl = () => get(window.SERVER_FLAGS, 'registry', 'quay.io/nyoxi'); // TODO: upstream should be moved under quay.io/kubevirt
-const getV2vImageTag = () => get(window.SERVER_FLAGS, 'v2vImageTag', 'latest');
-export const getV2vImagePullPolicy = () => get(window.SERVER_FLAGS, 'v2vImagePullPolicy', 'IfNotPresent');
+import { VMWARE_KUBEVIRT_VMWARE_CONFIG_MAP_NAME } from '../k8s/requests';
 
-export const VMWARE_TO_KUBEVIRT_OS_CONFIG_MAP_NAMESPACE = 'kube-public'; // note: common-templates are in the "openshift" namespace
-export const VMWARE_TO_KUBEVIRT_OS_CONFIG_MAP_NAME = 'vmware-to-kubevirt-os'; // single configMap per cluster, contains mapping of vmware guestId to common-templates OS ID
+// full image name of kubevirt-v2v-conversion
+// Presence and proper content of the ConfigMap is hard requirement, so ensure proper info makes it into the logs otherwise.
+export const getKubevirtV2vConversionContainerImage = kubevirtVmwareConfigMap =>
```
suomiy: these should probably go to selectors now

src/k8s/objects/v2v/vmware/conversionPod.js
```diff
@@ -2,7 +2,14 @@ import { getKubevirtV2vConversionContainerImage, getV2vImagePullPolicy } from '.
...
 import { CONVERSION_BASE_NAME, CONVERSION_GENERATE_NAME } from '../../../requests/v2v';
 
-export const buildConversionPod = ({ volumes, volumeMounts, namespace, serviceAccountName, secretName }) => ({
+export const buildConversionPod = ({
+  volumes,
+  volumeMounts,
+  namespace,
+  serviceAccountName,
+  secretName,
+  kubevirtVmwareConfigMap,
```
suomiy: can we please rather pass the `image` & `imagePullPolicy` to lower down the number of dependencies/imports

src/k8s/objects/v2v/vmware/vmWareDeployment.js
```diff
@@ -1,7 +1,7 @@
 import { getKubevirtV2vVmwareContainerImage, getV2vImagePullPolicy } from '../../../../config';
 import { DeploymentModel } from '../../../../models';
 
-export const buildVmWareDeployment = ({ name, namespace }) => ({
+export const buildVmWareDeployment = ({ name, namespace, kubevirtVmwareConfigMap }) => ({
```
suomiy: same here

src/k8s/requests/v2v/constants.js
```diff
@@ -7,3 +7,10 @@ export const CONVERSION_GENERATE_NAME = `${CONVERSION_BASE_NAME}-`;
 export const CONVERSION_PROGRESS_ANNOTATION = 'v2vConversionProgress';
 
 export const V2VVMWARE_DEPLOYMENT_NAME = 'v2v-vmware';
+
+export const VMWARE_KUBEVIRT_VMWARE_CONFIG_MAP_NAMESPACE = 'openshift'; // along the common-templates ...
```
suomiy: is the openshift namespace really the right one we want?

src/k8s/requests/v2v/constants.js
```diff
@@ -7,3 +7,10 @@ export const CONVERSION_GENERATE_NAME = `${CONVERSION_BASE_NAME}-`;
...
+export const VMWARE_KUBEVIRT_VMWARE_CONFIG_MAP_NAME = V2VVMWARE_DEPLOYMENT_NAME;
```
suomiy: can we copy the value instead? this could be unintentionally changed when updating V2VVMWARE_DEPLOYMENT_NAME

src/k8s/requests/v2v/startV2VvmwareController.js
```diff
@@ -12,21 +23,42 @@ const { info } = console;
...
 // The controller is namespace-scoped, especially due to security reasons
 // Let's make sure its started within the desired namespace (which is not by default).
 // The V2VVmware CRD is expected to be already created within the cluster (by Web UI installation)
+// TODO: The controller should be deployed by a provider and not via following UI code.
 export const startV2VVMWareController = async ({ namespace }, { k8sGet, k8sCreate, k8sKill, k8sPatch }) => {
   if (!namespace) {
     throw new Error('V2V VMWare: namespace must be selected');
   }
 
+  if (semaphor) {
```
suomiy: this does not take other namespaces and new versions into account

src/k8s/requests/v2v/constants.js
mareklibra: Any better idea? :-)

mareklibra: Must be accessible but still "safe"

- kubevirt is gone
- kube-public is dangerous
- kube-system is simply not right
- openshift already contains similar "common-templates" stuff

I do not want to introduce new one ...

src/k8s/requests/v2v/constants.js
mareklibra: honestly, that was my intention. But I can split.

src/k8s/requests/v2v/startV2VvmwareController.js
mareklibra: Can this happen within the single tab?

src/config/index.js
mareklibra: ok, will do

src/k8s/objects/v2v/vmware/conversionPod.js
mareklibra: I wanted to have flexibility and not pass all the params explicitly via the hierarchy. I can leverage destructuring

src/k8s/objects/v2v/vmware/vmWareDeployment.js
mareklibra: done

src/k8s/objects/v2v/vmware/conversionPod.js
mareklibra: done

src/k8s/requests/v2v/constants.js
mareklibra: done

src/k8s/requests/v2v/constants.js
suomiy: why is the `kube-public` dangerous?

I guess `openshift` will do, we do not want to introduce another namespace

src/k8s/requests/v2v/startV2VvmwareController.js
suomiy: I guess the versions probably not, but the namespaces certainly can - user can just select another namespace

src/k8s/requests/v2v/constants.js
mareklibra: I consider `kube-public` more like `/tmp` on unix.
I think there's higher probability that an "untrusted" user can get privileges to modify ConfigMaps in `kube-public` and so (s)he would be able to replace image (repo + name + tag) which is used for execution by another user.

Seems like `openshift` is the best option here so far.

src/k8s/requests/v2v/startV2VvmwareController.js
mareklibra: Right, this can happen. There is not much time for it, but it's possible. Will update the PR for this case.

src/k8s/requests/v2v/constants.js
suomiy: ok

src/k8s/requests/v2v/startV2VvmwareController.js
mareklibra: done

src/k8s/requests/v2v/startV2VvmwareController.js
```diff
@@ -53,8 +85,12 @@ export const startV2VVMWareController = async ({ namespace }, { k8sGet, k8sCreat
           ...nextResult,
         };
       },
-      Promise.resolve({ activeDeployment, namespace, name })
+      Promise.resolve({ activeDeployment, kubevirtVmwareConfigMap, namespace, name })
     );
+  } finally {
+    // eslint-disable-next-line no-console
+    console.log(`startV2VVMWareController for "${namespace}" namespace finished.`);
```
suomiy: IMO we should show this message only if the particular call deployed the controller and was not skipped because of the same version

src/k8s/requests/v2v/startV2VvmwareController.js
```diff
@@ -1,32 +1,64 @@
...
 // The V2VVmware CRD is expected to be already created within the cluster (by Web UI installation)
+// TODO: The controller should be deployed by a provider and not via following UI code.
 export const startV2VVMWareController = async ({ namespace }, { k8sGet, k8sCreate, k8sKill, k8sPatch }) => {
   if (!namespace) {
     throw new Error('V2V VMWare: namespace must be selected');
   }
 
+  if (semaphors[namespace]) {
+    // eslint-disable-next-line no-console
+    console.info(`startV2VVMWareController for "${namespace}" namespace already in progress. Skipping...`);
```
suomiy: not sure if this message is necessary

mareklibra: This functionality caused a lot of confusion so far, so I prefer to keep it for visibility.
No matter it's not a simple task, the user is unaware of it. 
Might help us with debugging ....

src/k8s/requests/v2v/startV2VvmwareController.js
mareklibra: Ok, this can add extra information in connection to the "initial" message

PR #516, opened 2019-07-17
Remove dot character from DNS-1123 validation

pcbailey:
The dot character is not allowed in a subdomain value as the dot character is the delimiter between subdomain values.


mareklibra:
lgtm, just tests need to be fixed

pcbailey:
That would be helpful. Wouldn't it. =) Done.

PR #519, opened 2019-07-18
Create VM Wizard: conditionally rendered Import provision source

mareklibra:
The V2V Import VMware option is supposed to be rendered just in case
that related infrastructure is deployed on the cluster.

In other words, check for presence of V2VVMware CRD is introduced.

As we have just a single provider (VMware) so far, the higher-level
"Import"-provision type is not rendered if the CRD is missing.

### DIFF
--- a/src/components/Wizard/CreateVmWizard/initialState/vmSettingsTabInitialState.js
+++ b/src/components/Wizard/CreateVmWizard/initialState/vmSettingsTabInitialState.js
@@ -48,8 +48,8 @@ export const getVmSettingsInitialState = props => ({
   valid: false,
 });
 
-export const getInitialVmSettings = (props = { createTemplate: false }) => {
-  const { createTemplate } = props;
+export const getInitialVmSettings = (props = { createTemplate: false, isV2vVmwareCrd: false }) => {
+  const { createTemplate, isV2vVmwareCrd } = props;
 
   const provisionSources = [
     PROVISION_SOURCE_PXE,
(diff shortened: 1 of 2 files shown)


src/components/Wizard/CreateVmWizard/initialState/vmSettingsTabInitialState.js
```diff
@@ -49,7 +49,7 @@ export const getVmSettingsInitialState = props => ({
 });
 
 export const getInitialVmSettings = (props = { createTemplate: false }) => {
```
suomiy: please add the `isV2vVmwareCrd` with `false` as a default. This could break things in the future if somebody for example chose to use it for `asHidden` which take distinction of `undefined` and `false` into account.

mareklibra:
> This approach is okay for now due to having only one import provider, but it should be changed in the future if other providers get included.

That's the plan. But so far the potentially empty list of providers would be confusing. In addition, this is meant as a workaround for potential project planing or cluster mis-configuration issues.

src/components/Wizard/CreateVmWizard/initialState/vmSettingsTabInitialState.js
mareklibra: ok, but the default `false` is recently handled via `CreateVmWizard` default prop.

suomiy: ok, just to be sure to have a better contract

PR #521, opened 2019-07-19
CloneDialog: support non-admin users

suomiy:
bz: https://bugzilla.redhat.com/show_bug.cgi?id=1726193

sibling PR: https://github.com/kubevirt/web-ui/pull/416

### DIFF
--- a/src/components/Dialog/CloneDialog/CloneDialog.js
+++ b/src/components/Dialog/CloneDialog/CloneDialog.js
@@ -11,11 +11,13 @@ import {
   START_VM_KEY,
   VIRTUAL_MACHINES_KEY,
 } from '../../Wizard/CreateVmWizard/constants';
-import { getDescription, getNamespace, getName, isVmRunning } from '../../../selectors';
+import { getDescription, getNamespace, getName, isVmRunning, getVolumes } from '../../../selectors';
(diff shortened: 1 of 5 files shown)


suomiy:
I tried not to interfiere with CloneDialog component much, because it it targeted against stable v2.0.0

src/components/Dialog/CloneDialog/CloneDialog.js
```diff
@@ -129,30 +138,50 @@ export class CloneDialog extends React.Component {
...
       <React.Fragment>
-        <Button bsStyle="default" className="btn-cancel" onClick={this.props.onClose}>
+        <Button bsStyle="default" className="btn-cancel" onClick={this.props.cancel}>
           Cancel
         </Button>
-        <Button bsStyle="primary" onClick={this.cloneVm} disabled={!this.state.valid}>
+        <Button
+          bsStyle="primary"
+          onClick={this.cloneVm}
+          disabled={!(this.state.valid && dataVolumesValid && pvcsValid)}
```
mareklibra: We should handle error state more explicitly - the user should be informed about transition from `loading` to `error` state.
I believe rendering info about just the most serious issue (as filtered by `Firehose`) is enough.

suomiy:
added error handling

PR #525, opened 2019-07-24
deprecate baseTemplates to simplify loading and OS list flicker

suomiy:
- disable fields when loading

- bug: https://bugzilla.redhat.com/show_bug.cgi?id=1730390

@mareklibra @rawagner 

IMO `web-ui-components` `base mock templates` shouldn't be needed anymore. Current usage causes inconsistent behaviour on user's side.

Thoughts?


rawagner:
+1 for removing mocked templates

mareklibra:
I agree that mock templates can be removed now, as the kubevirt's deployment has already settled down and templates are mandatory part of it. In addition, the dialog won't be rendered without kubevirt installed.

src/components/Wizard/CreateVmWizard/CreateVmWizard.js
```diff
@@ -329,6 +330,7 @@ CreateVmWizard.propTypes = {
   k8sPatch: PropTypes.func.isRequired,
   k8sKill: PropTypes.func.isRequired,
   onHide: PropTypes.func.isRequired,
+  templatesLoaded: PropTypes.bool,
```
mareklibra: I left comment regarding this prop in https://github.com/openshift/console/pull/2157

mareklibra: Agreed in 2157

PR #526, opened 2019-07-25
validate template sources with missing DataVolume

suomiy:
bug: https://bugzilla.redhat.com/show_bug.cgi?id=1687782


suomiy:
scree

mareklibra:
Snapshots need to be updated

suomiy:
updated

PR #527, opened 2019-07-26
Add MAC address validation to Create VM wizard

pcbailey:
Bug: https://bugzilla.redhat.com/1720333

### DIFF
--- a/src/components/Wizard/CreateVmWizard/NetworksTab.js
+++ b/src/components/Wizard/CreateVmWizard/NetworksTab.js
@@ -26,10 +26,12 @@ import {
   HEADER_NETWORK,
   HEADER_BINDING_METHOD,
   SELECT_BINDING,
+  MAC_ADDRESS_INVALID_ERROR,
 } from './strings';
 import { getInterfaceBinding } from '../../../selectors';
+import { isValidMAC } from '../../../utils';
 
-const validateNetwork = network => {
+export const validateNetwork = network => {
   const errors = Array(4).fill(null);
 
   if (!network || network.id == null) {
@@ -38,6 +40,10 @@ const validateNetwork = network => {
 
   if (!network.name) {
     errors[1] = NETWORK_ERROR_EMPTY_NAME;
+  }
+
+  if (network.mac && !isValidMAC(network.mac)) {
+    errors[2] = MAC_ADDRESS_INVALID_ERROR;
   }
 
   if (network.id !== 1 && !network.network) {
--- a/src/utils/tests/validation.test.js
+++ b/src/utils/tests/validation.test.js
@@ -1,5 +1,6 @@
 import {
   isPositiveNumber,
+  isValidMAC,
   validateDNS1123SubdomainValue,
   validateURL,
   validateContainer,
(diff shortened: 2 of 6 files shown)


src/components/Wizard/CreateVmWizard/NetworksTab.js
```diff
@@ -40,6 +41,11 @@ const validateNetwork = network => {
     errors[1] = NETWORK_ERROR_EMPTY_NAME;
   }
 
+  const macRegExp = /^(([A-Fa-f0-9]{2}[:\\-]){5}[A-Fa-f0-9]{2}[,]?)+$/;
```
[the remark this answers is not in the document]
suomiy: the tests should also include all the examples above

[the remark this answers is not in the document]
pcbailey: Done!

pcbailey:
I still need to test the functionality. I'm having trouble connecting to the test environment at the moment. I'll open a separate PR for the validation in console and link to this one.

src/utils/validations.js
```diff
@@ -194,3 +194,38 @@ export const validateBmcURL = value => {
...
+// IEEE 802 MAC-48, EUI-48, EUI-64, or a 20-octet IP over InfiniBand link-layer address
+// https://golang.org/pkg/net/#ParseMAC
+export const isValidMAC = mac => {
+  if (mac.length < 14) {
+    return false;
+  }
+
+  let isValid = false;
+  if (mac[2] === ':') {
+    const regex = new RegExp(
```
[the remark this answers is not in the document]
suomiy: Can we extract a `const HEXCH_REGEX = '[0-9A-Fa-f]'` variable out of these? 

Would be even better to have a function which creates a regex e.g. `f(groupLength, numOfGroups, delimiter)`

src/utils/validations.js
```diff
@@ -194,3 +194,38 @@ export const validateBmcURL = value => {
 
   return validateURL(value);
 };
+
+// Validates that the provided MAC address meets one of following formats supported by the golang ParseMAC function:
+// IEEE 802 MAC-48, EUI-48, EUI-64, or a 20-octet IP over InfiniBand link-layer address
+// https://golang.org/pkg/net/#ParseMAC
+export const isValidMAC = mac => {
```
suomiy: we could also move this function to a separate file

src/utils/validations.js
```diff
@@ -194,3 +194,38 @@ export const validateBmcURL = value => {
...
+        '^(([0-9A-Fa-f]{2}[-]){5}[0-9A-Fa-f]{2})$' // 01-23-45-67-89-ab
+    );
+    isValid = regex.test(mac);
+  } else if (mac[4] === '.') {
+    const regex = new RegExp(
+      '^(([0-9A-Fa-f]{4}.){9}[0-9A-Fa-f]{4})$|' + // 0123.4567.89ab.cdef.0000.0123.4567.89ab.cdef.0000
+      '^(([0-9A-Fa-f]{4}.){3}[0-9A-Fa-f]{4})$|' + // 0123.4567.89ab.cdef
+        '^(([0-9A-Fa-f]{4}.){2}[0-9A-Fa-f]{4})$' // 0123.4567.89ab
+    );
+    isValid = regex.test(mac);
```
suomiy: I would rather do either

```ts
return regex.test(mac)
```

or 
```ts
let regex;
...
regex = new RegExp(
```

and do test at the end

src/utils/tests/validation.test.js
```diff
@@ -172,3 +173,110 @@ describe('validation.js - validateVmName', () => {
     expect(validateVmLikeEntityName('vm1', vmSettings, props)).toEqual(getValidationObject(VIRTUAL_MACHINE_EXISTS));
   });
 });
+
+describe('validation.js - isValidMAC', () => {
+  // octets with colon separator
+  it('returns true for a valid 6 octet MAC address with colon separator', () => {
```
suomiy: This would look more readable IMO:


```js
const testConfig = [{
  title: 'returns true for a valid 6 octet MAC address with colon separator',
  mac: '01:23:45:67:89:ab',
}]
testConfig.forEach(c => {
  it(c.title,  () => {
    expect(isValidMAC(c.mac)).toBeTruthy()
  }
})
```

you can make two configs for truthy/falsy results or indicate the result in the config itself

src/utils/tests/validation.test.js
```diff
@@ -172,3 +173,110 @@ describe('validation.js - validateVmName', () => {
     expect(validateVmLikeEntityName('vm1', vmSettings, props)).toEqual(getValidationObject(VIRTUAL_MACHINE_EXISTS));
   });
 });
+
+describe('validation.js - isValidMAC', () => {
```
suomiy: we could also consider moving this test to a separate file

src/utils/validations.js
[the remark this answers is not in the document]
pcbailey: I created regexes for each of the delimiter types for which we're looking. That made things look much cleaner.

If we created a function like the one you suggested, we'd have to run the function three times for each of the delimiter regexes and then concatenate those results to capture all three possible lengths. I think that would be more confusing than just having the regexes as they are with the comments showing the format that each part of the regex is trying to match.

I also thought about the hex character regex, but I think it would decrease readability to mix the regex with string interpolation like that.

src/utils/validations.js
pcbailey: At the moment I don't think it's complex enough to warrant its own file, but if it grows in the future I could see doing that.

src/utils/validations.js
pcbailey: Done! I went with the second option.

src/utils/tests/validation.test.js
pcbailey: I did this, but mainly to be consistent with how other tests are written. I don't find it more readable than having an explicit test for each scenario, as the tests are written to be read like a sentence. With the config setup the reader has to mentally map the contents of each object to the function. So, I think the config setup is better in terms of conciseness, but inferior in terms of readability.

As an aside, in the original version, I think it would have been better for me to call the function directly in the `expect()` call instead of storing the return value to a variable. That would have saved several lines.

src/utils/tests/validation.test.js
pcbailey: As with moving the function, I think we should do this only if the function grows in complexity in the future.

src/utils/validations.js
suomiy: ok we can do it in the future

src/utils/tests/validation.test.js
suomiy: I guess it is subjective, but it is more readable to me. I can see all the testcases in one places and don't have to trawl through other unrelated lines.

src/utils/validations.js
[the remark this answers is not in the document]
suomiy: ok, but IMO for example just `HEXCH` is more readable than `[0-9A-Fa-f]` especially when it is used multiple times in the same regex. And is less prone to mistakes.

String interpolation is quite readable by itself so I don't see any problem mixing it with regex

src/utils/validations.js
```diff
@@ -231,3 +231,42 @@ export const validateBmcURL = value => {
...
+// IEEE 802 MAC-48, EUI-48, EUI-64, or a 20-octet IP over InfiniBand link-layer address
+// https://golang.org/pkg/net/#ParseMAC
+export const isValidMAC = mac => {
+  if (mac.length < 14) {
+    return false;
+  }
+
+  let regex;
+  if (mac[2] === COLON_DELIMITER) {
+    regex = new RegExp(MAC_REGEX_COLON_DELIMITER);
```
suomiy: it is not necessary to call `RegExp` each time. I would rather create it when creating the const variable.

src/utils/validations.js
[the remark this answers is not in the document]
pcbailey: I disagree, but you seem to have a stronger opinion about it than I do, so I made the change. =)

src/utils/tests/validation.test.js
pcbailey: Agreed. It's totally subjective. I'm marking this as resolved since the change has been made.

src/utils/validations.js
pcbailey: Done.

pcbailey: Done.

src/utils/validations.js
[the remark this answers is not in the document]
suomiy: I guess the readability is again subjective, but I hope you agree with the reusability/less mistakes (smaller area) part.

suomiy:
@pcbailey could you port all this functionality to https://github.com/openshift/console and fix the CreateNicRow ?

PR #528, opened 2019-07-29
Use SC defaults for DV creation

mareklibra:
The `kubevirt-storage-class-defaults` ConfigMap is read for `accessMode` and `volumeMode` defaults when creating a DataVolume.

Sibling PR: https://github.com/openshift/console/pull/2200

ConfigMap example:
```
kind: ConfigMap
apiVersion: v1
metadata:
  # well-known name
  name: kubevirt-storage-class-defaults
  namespace: openshift
data:
    # The default if a StorageClass is not listed.
    # If these ultimate defaults are missing or the ConfigMap is inaccessible, these
    # defaults are hard-coded in the UI as well.
  accessMode: ReadWriteOnce
  volumeMode: Filesystem
  local-sc.accessMode: ReadWriteOnce
  local-sc.volumeMode: Filesystem
  nfs-sc.accessMode: ReadWriteMany
  nfs-sc.volumeMode: Filesystem
```

### DIFF
--- a/src/constants/index.js
+++ b/src/constants/index.js
@@ -37,6 +37,10 @@ export const PROVISION_SOURCE_CLONED_DISK = 'Cloned Disk'; // PVC or upload imag
 
 export const PVC_ACCESSMODE_RWO = 'ReadWriteOnce';
 export const PVC_ACCESSMODE_RWM = 'ReadWriteMany';
+export const PVC_ACCESSMODE_DEFAULT = PVC_ACCESSMODE_RWO;
+export const PVC_VOLUMEMODE_FS = 'Filesystem';
+export const PVC_VOLUMEMODE_BLOCK = 'Block';
+export const PVC_VOLUMEMODE_DEFAULT = PVC_VOLUMEMODE_FS;
 
 export const TEMPLATE_API_VERSION = 'template.openshift.io/v1';
 export const TEMPLATE_FLAVOR_LABEL = 'flavor.template.kubevirt.io';
--- a/src/k8s/clone.js
+++ b/src/k8s/clone.js
@@ -14,6 +14,8 @@ import {
   getDataVolumeStorageClassName,
   getOperatingSystemName,
   getOperatingSystem,
+  getPvcVolumeMode,
+  getDataVolumeMode,
 } from '../selectors';
 import { getStartStopPatch, generateDiskName } from '../utils';
 import { addDataVolumeTemplate, addTemplateLabel } from './vmBuilder';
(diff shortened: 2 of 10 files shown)


mareklibra:
Thinking about the ConfigMap format again, I believe the ultimate default (if matching SC can not be found) does not need to be prefixed, as the key-format is recently `[sc-name].[option]`.

It means, the keys can be like:
```
  accessMode: ... # for ultimate default
  my-sc-name.accessMode: ...
  another-sc-name.accessMode: ...
```

I will adapt this PR accordingly.

mareklibra:
The ultimate defaults are simplified now.

src/constants/index.js
```diff
@@ -37,6 +37,10 @@ export const PROVISION_SOURCE_CLONED_DISK = 'Cloned Disk'; // PVC or upload imag
 
 export const PVC_ACCESSMODE_RWO = 'ReadWriteOnce';
 export const PVC_ACCESSMODE_RWM = 'ReadWriteMany';
+export const PVC_ACCESSMODE_DEFAULT = PVC_ACCESSMODE_RWO;
+export const PVC_VOLUMEMODE_FS = 'Filesystem';
+export const PVC_VOLUMEMODE_BLOCK = 'Block';
+export const PVC_VOLUMEMODE_DEFAULT = PVC_VOLUMEMODE_BLOCK;
```
aglitke: Hardcoded defaults should probably be RWO/File

mareklibra: done

src/selectors/configmap/sc-defaults.js
```diff
@@ -0,0 +1,13 @@
+import { get } from 'lodash';
+
+import { PVC_ACCESSMODE_DEFAULT, PVC_VOLUMEMODE_DEFAULT } from '../../constants';
+
+export const getDefaultSCAccessMode = (storageClassConfigMap, scName) => {
+  const defaultAccessMode = get(storageClassConfigMap, ['data', 'accessMode'], PVC_ACCESSMODE_DEFAULT);
+  return get(storageClassConfigMap, ['data', `${scName}.accessMode`], defaultAccessMode);
```
suomiy: is the property really named `${scName}.accessMode` with a dot inside?

src/k8s/clone.js
```diff
@@ -136,12 +140,21 @@ const addPvcClones = (vm, newName, persistentVolumeClaims) => {
     });
 };
 
-export const addTemplateClone = (vm, pvcName, pvcNamespace, accessModes, size, storageClassName, vmName) => {
-  const template = getPvcCloneTemplate(pvcName, pvcNamespace, accessModes, size, storageClassName, vmName);
+export const addTemplateClone = (
```
suomiy: ugh, how did we get to this? These parameters are quite unmaintainable..

src/selectors/configmap/sc-defaults.js
mareklibra: yes

src/k8s/clone.js
mareklibra: right, we need to refactor. In this particular case, I'm just adding `volumeMode`

PR #531, opened 2019-07-31
v2v: fix VM loading when race condition occurs

suomiy:
@mareklibra


mareklibra:
@suomiy , thanks for the fix :-)

PR #532, opened 2019-08-01
v2v: move v2v-vmware ConfigMap to kubevirt-hyperconverged namespace

mareklibra:
More safe to avoid unrelated modifications.

Follow-up for #507

### DIFF
--- a/src/k8s/requests/v2v/startV2VvmwareController.js
+++ b/src/k8s/requests/v2v/startV2VvmwareController.js
@@ -35,15 +35,15 @@ export const startV2VVMWareController = async ({ namespace }, { k8sGet, k8sCreat
     throw new Error('V2V VMWare: namespace must be selected');
   }
 
+  const name = V2VVMWARE_DEPLOYMENT_NAME;
+  let activeDeployment;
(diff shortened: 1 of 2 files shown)


mareklibra:
Cc @jwmatthews

src/k8s/requests/v2v/startV2VvmwareController.js
```diff
@@ -35,22 +35,36 @@ export const startV2VVMWareController = async ({ namespace }, { k8sGet, k8sCreat
     throw new Error('V2V VMWare: namespace must be selected');
   }
 
+  const name = V2VVMWARE_DEPLOYMENT_NAME;
```
rawagner: isnt this unnecessary ? why not use `V2VVMWARE_DEPLOYMENT_NAME` directly, instead of `name` ?

mareklibra: good point, I can change

suomiy: The name is used in multiple places - I would rather leave it like it is

jwmatthews:
cc @jmontleon

PR #533, opened 2019-08-01
v2v: Wait for ServiceAccount before creating conversion pod

mareklibra:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1721018


src/k8s/requests/v2v/constants.js
```diff
@@ -14,3 +14,5 @@ export const VMWARE_KUBEVIRT_VMWARE_CONFIG_MAP_NAME = 'v2v-vmware';
 export const VMWARE_TO_KUBEVIRT_OS_CONFIG_MAP_NAMESPACE = 'kube-public'; // note: common-templates are in the "openshift" namespace
 // TODO: make it configurable via VMWARE_KUBEVIRT_VMWARE_CONFIG_MAP
 export const VMWARE_TO_KUBEVIRT_OS_CONFIG_MAP_NAME = 'vmware-to-kubevirt-os'; // single configMap per cluster, contains mapping of vmware guestId to common-templates OS ID
+
+export const CONVERSION_SERVICEACCOUNT_DELAY = 5 * 1000; // in ms
```
suomiy: IMHO, I would rather do more polls more frequently (the operation is not so expensive). The UI would stall for 5,10 sec before it gets the secrets.

mareklibra: Changed to up to 15 attempts with 2 secs delay in between.

suomiy: +1

PR #537, opened 2019-08-07
Default to ReadWriteMany

yaacov:
In https://github.com/kubevirt/web-ui-components/pull/528 we changed the default PVC access mode from `PVC_ACCESSMODE_RWM` to `PVC_ACCESSMODE_DEFAULT` [1] that is `PVC_ACCESSMODE_RWO` [2]

If we run a VM create wizard that leave the PVC access mode `undefined` the VM will fail because the correct default value is `PVC_ACCESSMODE_RWM` and not `PVC_ACCESSMODE_RWO`.

This will fail the e2e tests.

Refs:
[1] https://github.com/kubevirt/web-ui-components/pull/528/files#diff-4f2c25e67b23719f1195914b33c45601R324
[2] https://github.com/kubevirt/web-ui-components/pull/528/files#diff-bc152a022be0db29b061aedc08d80a45R40

### DIFF
--- a/src/constants/index.js
+++ b/src/constants/index.js
@@ -39,7 +39,7 @@ export const PROVISION_SOURCE_UNKNOWN_DATAVOLUME_SOURCE = 'Unknown DataVolume So
 
 export const PVC_ACCESSMODE_RWO = 'ReadWriteOnce';
 export const PVC_ACCESSMODE_RWM = 'ReadWriteMany';
-export const PVC_ACCESSMODE_DEFAULT = PVC_ACCESSMODE_RWO;
+export const PVC_ACCESSMODE_DEFAULT = PVC_ACCESSMODE_RWM;
 export const PVC_VOLUMEMODE_FS = 'Filesystem';
 export const PVC_VOLUMEMODE_BLOCK = 'Block';
 export const PVC_VOLUMEMODE_DEFAULT = PVC_VOLUMEMODE_FS;


yaacov:
@suomiy @rawagner @mareklibra please review.

yaacov:
@pcbailey please review

yaacov:
Background info [ @awels ]:

> So it depends on what you want, you need RWX if you want to live migrate, but that requires a storage class that supports it like ceph or nfs. If you are using local volume storage for instance you have to use RWO since that is the only thing it supports.
> So there is no definite answer on this is what you need. It depends on what features your storage class supports, and if you want live migration or not

mareklibra:
I think we do not have right solution here, we need to look for a just good-enough.

The `RWO` was chosen as a safe default, as it is supported on (probably) all storage classes.
Considering context, these hardcoded defaults are the ultimate fall-back in a misconfigured environment. So I would prefer to keep them as safe as possible.

If this is an issue for e2e tests, we can update the test environment by a ConfigMap (see [1])

No matter what is stated above, we need to ensure that the default VM configuration created by the VM Wizard of YAML editor just by "dummy" passing-through is runable.

[1] https://github.com/kubevirt/web-ui-components/pull/528

yaacov:
@mareklibra Thanks for the info :+1: 

closing in favour of using the config map solution .

PR #536, opened 2019-08-06
add babel esm and umd exports

redallen:
Fix https://github.com/openshift/console/pull/2232 upstream.

PR #539, opened 2019-08-15
Add module to package.json

redallen:
Follow up to #536. Forgot the reference for Webpack.

Edit: Also, please publish this to NPM!


redallen:
@mareklibra @vojtechszocs @suomiy After this is merged+published I can work on slashing the [OpenShift bundle size](https://github.com/openshift/console/pull/2232) in half :)

PR #543, opened 2019-08-19
Validate hostname in cloud-init settings

pcbailey:
This PR adds hostname validation to the cloud-init settings section of the create VM/template wizard.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1648321

### DIFF
--- a/src/components/Wizard/CreateVmWizard/validations/vmSettingsTabValidation.js
+++ b/src/components/Wizard/CreateVmWizard/validations/vmSettingsTabValidation.js
@@ -8,6 +8,7 @@ import {
   validateURL,
   validateMemory,
   validateUserTemplate,
+  validateCloudInitHostName,
 } from '../../../../utils/validations';
 import { objectMerge } from '../../../../utils/utils';
 import { settingsValue } from '../../../../k8s/selectors';
@@ -22,6 +23,7 @@ import {
   PROVISION_SOURCE_TYPE_KEY,
   MEMORY_KEY,
   USER_TEMPLATE_KEY,
+  HOST_NAME_KEY,
 } from '../constants';
 import { NAMESPACE_MUST_BE_SELECTED } from '../../../../utils/strings';
 import { isProviderValid, validateProvider } from '../providers';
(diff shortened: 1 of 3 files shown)


src/components/Wizard/CreateVmWizard/validations/vmSettingsTabValidation.js
```diff
@@ -51,6 +52,7 @@ const validateResolver = {
   [IMAGE_URL_KEY]: asVmSettingsValidator(validateURL),
   [PROVIDER_KEY]: validateProviderDropdown,
   [MEMORY_KEY]: asVmSettingsValidator(validateMemory),
+  [HOST_NAME_KEY]: asVmSettingsValidator(validateVmLikeEntityName),
```
suomiy: I don't think host name should have the same validation as VM name

pcbailey: You're right. I'll create a separate function for it. I think for the moment all that's needed is validateDNS1123SubdomainValue, correct?

suomiy: y I guess, that should be enough

pcbailey: Done!

src/utils/validations.js
```diff
@@ -123,6 +123,11 @@ export const validateVmLikeEntityName = (value, vmSettings, props) => {
       );
 };
 
+export const validateCloudInitHostName = value => {
+  const dnsValidation = validateDNS1123SubdomainValue(value);
+  return dnsValidation && dnsValidation.type === VALIDATION_ERROR_TYPE ? dnsValidation : null;
```
suomiy: I don't think we have to check for the error_type here, because all the errors will be error type or none in this case. Nevertheless I do think we should check for empty error here, because the host name can be empty if I am not mistaken.

pcbailey: You're correct. Done!

src/utils/validations.js
```diff
@@ -123,6 +123,11 @@ export const validateVmLikeEntityName = (value, vmSettings, props) => {
       );
 };
 
+export const validateCloudInitHostName = value => {
+  const dnsValidation = validateDNS1123SubdomainValue(value);
+  return dnsValidation && dnsValidation.type !== EMPTY_ERROR ? dnsValidation : null;
```
suomiy: a bit differently

```suggestion
  return dnsValidation && dnsValidation.isEmptyError ? null : dnsValidation;
```

pcbailey: That looks much cleaner, thanks. =) I've updated the commit.

PR #545, opened 2019-08-23
Namespaces for kubevirt-storage-class-defaults and v2v-vmware

mareklibra:
Fixes issue with changing location of ConfigMaps in different releases.

The `kubevirt-storage-class-defaults` ConfigMap is attempted to be retrieved from the `openshift-cnv` first, then from `openshift` in case of error.

The `v2v-vmware` ConfigMap is read from `openshift-cnv` namespace first, then from `kubevirt-hyperconverged`.

### DIFF
--- a/src/k8s/requests/v2v/vmwareConfigMap.js
+++ b/src/k8s/requests/v2v/vmwareConfigMap.js
@@ -1,0 +1,40 @@
+import { VMWARE_KUBEVIRT_VMWARE_CONFIG_MAP_NAMESPACES, VMWARE_KUBEVIRT_VMWARE_CONFIG_MAP_NAME } from './constants';
+import { ConfigMapModel } from '../../../models';
+
+const { info, warn } = console;
+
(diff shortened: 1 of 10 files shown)


src/k8s/requests/v2v/vmwareConfigMap.js
```diff
@@ -0,0 +1,40 @@
...
+      `The ${VMWARE_KUBEVIRT_VMWARE_CONFIG_MAP_NAME} can not be found in the ${namespace} namespace. Trying another in one ... Error: `,
+      e
+    );
+  }
+  return null;
+};
+
+export const getVmwareConfigMap = async props => {
+  // query namespaces sequentially to respect order
+  for (let index = 0; index < VMWARE_KUBEVIRT_VMWARE_CONFIG_MAP_NAMESPACES.length; index++) {
```
suomiy: I get that the warns are different, but can we make at least this part generic? and add the comment from `getVmwareConfigMapInNamespace` to this function?

fabiand:
The best way to find the config map is to look for the namespace of the HCO CR.

PR #549, opened 2019-09-10
Align vmware secret with DNS 1123

mareklibra:


src/k8s/requests/v2v/utils.js
```diff
@@ -1,9 +1,30 @@
...
 
+// TODO: this needs to be improved to conform validations.js::validateDNS1123SubdomainValue()
+const alignWithDNS1123 = str => {
+  if (!str) {
+    return '';
+  }
+
+  // starts with alphaNum
+  if (!str.charAt(0).match(alphanumericRegex)) {
+    return alignWithDNS1123(str.slice(1));
```
suomiy: this can be done much easier without a recursion

```ts
str.replace(/^[^a-zA-Z0-9]*/, '').replace(/[^a-zA-Z0-9]*$/, '')
```

src/k8s/requests/v2v/utils.js
```diff
@@ -1,9 +1,30 @@
+import { alphanumericRegex } from '../../../utils';
+
 /**
  * Based on V2V Provider Pod manifest.yaml
  */
 
 const MAX_LEN = 30;
 
+// TODO: this needs to be improved to conform validations.js::validateDNS1123SubdomainValue()
+const alignWithDNS1123 = str => {
```
suomiy: can we rename it to `trimAccordingToDNS1123` ?

mareklibra: My intention is to further improve this method to fully conform DNS 1123, not just start and end. This first version was meant as a quick fix to unblock testing immediately. So it makes sense to keep the name as it is.

src/k8s/requests/v2v/utils.js
mareklibra: sure, but I like recursion :-)

I can change it.

PR #551, opened 2019-09-12
Use Dropdown from pf-react 3

mareklibra:
PF 3 Dropdown is replaced by PF 4 equivalent.
Component's API has changed, so our code is adapted accordingly.


mareklibra:
The later issue with the `Consoles` tab is fixed separately by https://github.com/openshift/console/pull/2727 .

This PR is now just "nice-to-have" as we can still use the older PF3 Dropdowns.

priley86:
>>> @priley86 , any idea how to fix the reference to pf-react/consoles in openshift/consoles via web-ui-components ? If we can fix that on this level, we will avoid patching pf-react/consoles as the consoles are PF3 and the working Dropdown is PF4.

hey @mareklibra - It looks like you are on the right track here w.r.t to conversion to PF4 Dropdown. My best guess is that you'll have fewer style conflicts and compatibility issues if targeting PF4 where possible. It is certainly possibly to style PF3 Dropdown to look more like PF4 Dropdown, but we noticed a lot of problems w/ consistently styling surrounding items (i.e. PF3 Buttons will no longer match PF4 Dropdown Button height, menu container is styled differently in PF4, etc.)

The upstream was unable to adopt PF4 React Dropdown components [1] due to JS API conflict, so they decided to adopt the CSS classes as an initial step [2]. So this meant replacing things like `.dropdown-menu` with `'.pf-c-dropdown__menu`, `.form-control` with `.pf-c-form-control`, `.dropdown-toggle` with `.pf-c-dropdown__toggle` , etc... I like the long term plan though is to consume `Dropdown` directly from PF-React. Since you are heavily integrated there, I would probably try to match these classes and styles best you can...

In regards to `@patternfly/react-console`, it does not seem a good option to try to mix PF3 components/markup w/ PF4 classes. In theory it should work, but I always had trouble with this and often found that one style change would lead to several others. For that reason, I would consider targeting PF4 Dropdown components inside `@patternfly/react-console`, potentially as additional JS components inside that package. I thought of maybe passing some prop such as `pf4=true` to AccessConsoles component for example, which could render alternative JSX, but the problem here of mixing APIs will be Webpack resolution and tree-shaking. I believe the long term goal is to remove all PF3 React dependencies, and this will add additional unnecessary weight... so that being said, I would split that code towards a PF4 edition/exposed module.

[1]: https://github.com/patternfly/patternfly-react/tree/master/packages/patternfly-4/react-core/src/components/Dropdown
[2]: https://github.com/openshift/console/pull/1751/files

PR #555, opened 2019-09-19
take storageClassConfigMap into account when creating disk patches

suomiy:
bug: https://bugzilla.redhat.com/show_bug.cgi?id=1753688


suomiy:
@mareklibra

mareklibra:
lgtm if my comments in https://github.com/openshift/console/pull/2782 get addressed.

PR #576, opened 2019-11-12
add default value to getsettings

yaacov:
Add a default value to the `getSetting` method used in `createVmTemplate` to make it similar to the `getSetting` method used in `createVM`

See discussion here:
https://github.com/openshift/console/pull/3355#pullrequestreview-315615211


yaacov:
@mareklibra @suomiy @rawagner @glekner @irosenzw please review

